PR #97, opened 2023-06-14
Add separate DB folders for each chain on initialization.

### LINKED ISSUE #90: Use different DB paths based on the selected chain
### Issue: 

Right now, if you use all chains use the same database directory. This is an issue because if we switch for example from westend to polkadot, the node is going to use the state of westend and potentially corrupt it.

### Solution: 
**We should have separate directories for each chain.**
 
### Bonus requirement:
We can add an environment variable that indicates to the node to use a clean directory on start up.

This can be achieved in two ways:
- Clean up the directory when the node shuts down
- Use a temporary directory that get's deleted when the node shuts down.

stoyan-lime:
Closes #90

### DIFF
--- a/src/main/java/com/limechain/storage/DBRepository.java
+++ b/src/main/java/com/limechain/storage/DBRepository.java
@@ -9,6 +9,7 @@ import org.springframework.util.SerializationUtils;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Optional;
 import java.util.logging.Level;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/limechain/storage/DBRepository.java
```diff
@@ -20,7 +21,7 @@ public class DBRepository implements KVRepository<String, Object> {
     /**
      * Main DB file
      */
-    private final static String FILE_NAME = "db";
+    private final static String FOLDER_NAME = "db";
```
georg-getz: You can change the inline doc to say folder as well.

georg-getz (approved):

bokoto000 (approved):

PR #98, opened 2023-06-14
Add db-recreate flag

stoyan-lime:
Adding flag for recreating the DB, when developing locally.

Will delete the folder for the current chain only after #90 is merged.

### DIFF
--- a/src/main/java/com/limechain/cli/Cli.java
+++ b/src/main/java/com/limechain/cli/Cli.java
@@ -47,9 +47,9 @@ public class Cli {
         try {
             CommandLineParser parser = new DefaultParser();
             CommandLine cmd = parser.parse(options, args);
-            String network = cmd.getOptionValue("network", "").toLowerCase();
+            String network = cmd.getOptionValue(NETWORK, "").toLowerCase();
             if (!validChains.contains(network) && !network.isEmpty()) throw new ParseException("Invalid network");
-            String dbPath = cmd.getOptionValue("db-path", DBInitializer.DEFAULT_DIRECTORY);
+            String dbPath = cmd.getOptionValue(DB_PATH, DBInitializer.DEFAULT_DIRECTORY);
             boolean dbRecreate = cmd.hasOption(DB_RECREATE);
 
             return new CliArguments(network, dbPath, dbRecreate);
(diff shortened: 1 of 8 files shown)


src/main/java/com/limechain/cli/Cli.java
```diff
@@ -49,8 +48,9 @@ public CliArguments parseArgs(String[] args) {
             String network = cmd.getOptionValue("network", "").toLowerCase();
             if (!validChains.contains(network) && !network.isEmpty()) throw new ParseException("Invalid network");
             String dbPath = cmd.getOptionValue("db-path", DBInitializer.DEFAULT_DIRECTORY);
+            boolean dbRecreate = cmd.hasOption(DB_RECREATE);
```
vikinatora: Not related to the implementation but I think it might be a good idea to extract the other cli options - `network` and `db-path` into static final variables just to keep the code tidy and organized.

src/main/java/com/limechain/storage/DBRepository.java
```diff
@@ -47,6 +51,19 @@ public DBRepository(String path, String chain) {
...
+            FileUtils.cleanDirectory(file.getAbsoluteFile());
+            log.log(Level.INFO, "\uD83D\uDDD1️DB cleaned");
+
+        } catch (IOException e) {
+            log.log(Level.SEVERE, String.format("Error deleting db folder. Exception: '%s', message: '%s'",
+                            e.getCause(),
+                            e.getMessage()),
+                    e);
+        }
+    }
```
vikinatora: Not sure if we should catch the error and allow the host to continue. This puts us in risk of using old DB data that can cause undefined behaviour.

vikinatora (changes requested):
Maybe we should add info that this parameter is available in the readme. Also tests for this functionality are missing from `CliTest`

src/main/java/com/limechain/cli/Cli.java
```diff
@@ -49,8 +50,9 @@ public CliArguments parseArgs(String[] args) {
             String network = cmd.getOptionValue("network", "").toLowerCase();
             if (!validChains.contains(network) && !network.isEmpty()) throw new ParseException("Invalid network");
             String dbPath = cmd.getOptionValue("db-path", DBInitializer.DEFAULT_DIRECTORY);
```
vikinatora: `network` and `db-path` strings should be replaced with the newly extracted variables

georg-getz (approved):

vikinatora (approved):

PR #104, opened 2023-06-16
Use Connection Manager instead of connections map

### LINKED ISSUE #94: Change AddressBook implementation
Change the implementation of AddressBook from Nabu/jvm-libp2p so that you're able to access the addresses
See Peergos/nabu/src/main/java/org/peergos/RamAddressBook.java.

This allows Fruzhin to have access to the addresses directly. Use them as counter, set them, remove them, etc.

stoyan-lime:
Closes #94


src/main/java/com/limechain/network/Network.java
```diff
@@ -141,14 +139,31 @@ public String[] getListenAddresses() {
...
-            PeerAddresses peer = new PeerAddresses(entry.getKey(), entry.getValue());
-            peers.add(peer);
-        }
-        return peers;
+        return kademliaService.getAddresses()
+                .entrySet()
+                .stream()
+                .map(this::mapAddressBookEntryToPeerAddresses)
+                .toList();
+    }
```
vikinatora: Since `getPeers` is not used anywhere currently, I'd suggest to remove this functionality. (it's going to be used for the `system_peers` RPC endpoint in the near future)

src/main/java/com/limechain/network/ConnectionManager.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public static ConnectionManager getInstance() {
+        if (INSTANCE == null) {
+            INSTANCE = new ConnectionManager();
+        }
+        return INSTANCE;
+    }
+
+    public void addPeer(PeerId peerId) {
+        peers.add(peerId);
```
bokoto000: Maybe we should log that we have added a peer to our connections.

src/main/java/com/limechain/network/ConnectionManager.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public void removePeer(PeerId peerId) {
+        peers.remove(peerId);
```
bokoto000, line 39: Also log the removal of a peer from our connections

src/main/java/com/limechain/network/protocol/blockannounce/BlockAnnounceEngine.java
```diff
@@ -35,42 +34,44 @@ public class BlockAnnounceEngine {
...
-        //Decode handshake from Polkadot
-        if (!hasKey && msg.length == HANDSHAKE_LENGTH) {
-            ScaleCodecReader reader = new ScaleCodecReader(msg);
-            BlockAnnounceHandshake handshake = reader.read(new BlockAnnounceHandshakeScaleReader());
-            peerHandshakes.put(peerId, handshake);
-            log.log(Level.INFO, "Received handshake from " + peerId + "\n" +
-                    handshake);
-
-            writeHandshakeToStream(stream, peerId);
+        if(!connectedToPeer && !isHandshake) {
```
bokoto000, line 40: Should we move the connectedToPeer check above so we don't log it as no handshake?

bokoto000:
Perhaps adding a connection check to the other protocols so that we can ensure communication only with our set of peers?  @vikinatora 

Created a new issue for this: #106

src/main/java/com/limechain/network/ConnectionManager.java
```diff
@@ -18,19 +22,29 @@ public static ConnectionManager getInstance() {
         return INSTANCE;
     }
 
-    public void addPeer(PeerId peerId) {
-        peers.add(peerId);
+    public void addPeer(PeerId peerId, BlockAnnounceHandshake blockAnnounceHandshake) {
+        peers.put(peerId, blockAnnounceHandshake);
+    }
+
+    public void updatePeer(PeerId peerId, BlockAnnounceMessage blockAnnounceMessage) {
```
bokoto000, line 28: Should we rename this function to updatePeerData

src/main/java/com/limechain/network/ConnectionManager.java
bokoto000: We currently log the information when receiving handshake.

src/main/java/com/limechain/network/ConnectionManager.java
bokoto000: We currently log it onClose in the block announce protocol.

bokoto000 (approved):

src/main/java/com/limechain/network/protocol/warp/dto/BlockHeader.java
```diff
@@ -26,4 +31,15 @@ public String toString() {
                 ", digest=" + Arrays.toString(digest) +
                 '}';
     }
+
+    public byte[] hash() {
```
vikinatora: I think that `getHash` would be a better name

src/main/java/com/limechain/network/ConnectionManager.java
vikinatora: I think it's fine this way

src/main/java/com/limechain/network/ConnectionManager.java
```diff
@@ -0,0 +1,50 @@
...
+import io.emeraldpay.polkaj.types.Hash256;
+import io.libp2p.core.PeerId;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class ConnectionManager {
+    private static ConnectionManager INSTANCE;
+    protected final Map<PeerId, BlockAnnounceHandshake> peers = new HashMap<>();
```
vikinatora: Handshakes are one time messages and shouldn't be updated.  I'd rather create a new class like `PeerInfo` instead of using `BlockAnnounceHandshake` even if it contains the same fields in the begginning. In the future it will probably contain other fields like `reputation` and others

vikinatora (changes requested):

src/main/java/com/limechain/network/protocol/blockannounce/scale/BlockAnnounceMessage.java
```diff
@@ -1,16 +1,20 @@
...
-    public boolean isBestBlock;
+    private BlockHeader header;
+    private boolean bestBlock;
 
     @Override
     public String toString() {
         return "BlockAnnounceMessage{" +
                 "header=" + header +
-                ", isBestBlock=" + isBestBlock +
+                ", isBestBlock=" + bestBlock +
```
georg-getz: You forgot to change the string as well.

vikinatora (approved):

PR #109, opened 2023-07-03
Add runtime and chain information download state

bokoto000:
Added:
- More boot nodes
- Shorten light messages logs
- Runtime build state
- Chain information download state

TODO:
- Fix `wasmer_jni` not found in java path bug - Fix for development you can put the wasmer_jni.dlyb in a java library folder


build.gradle.kts
```diff
@@ -27,6 +27,7 @@ repositories {
 
 dependencies {
     implementation("org.rocksdb:rocksdbjni:7.8.3")
+    implementation("com.google.guava:guava:31.1-jre")
```
vikinatora: Guava isn't used anywhere.

build.gradle.kts
```diff
@@ -57,6 +58,7 @@ dependencies {
     implementation("org.java-websocket:Java-WebSocket:1.5.3")
     implementation("org.javatuples:javatuples:1.2")
 
+    implementation("com.github.luben:zstd-jni:1.5.2-5")
```
vikinatora, line 60: This is not used anywhere as well

src/main/java/com/limechain/network/Network.java
```diff
@@ -197,7 +197,7 @@ private void updateCurrentSelectedPeer() {
     }
 
     public WarpSyncResponse makeWarpSyncRequest(String blockHash) {
-        if (isPeerInvalid()) return null;
+        if (validatePeer()) return null;
```
vikinatora: I don't like this change. I purposefully renamed it to `isPeerInvalid` because by reading `validatePeer` you don't understand whether it will return true or false for a valid peer. In fact, it returns `false` for a valid peer.

src/main/java/com/limechain/sync/warpsync/runtime/RuntimeVersion.java
```diff
@@ -0,0 +1,53 @@
...
+        byte[][] apiVersions = new byte[apiVersionsSize][];
+        long[] apiVersionsNumbers = new long[apiVersionsSize];
+        for (int i = 0; i < apiVersionsSize; i++) {
+            apiVersions[i] = reader.readByteArray(8);
+            apiVersionsNumbers[i] = reader.readUint32();
+        }
+
+        this.setTransactionVersion(BigInteger.valueOf(reader.readUint32()));
+        this.setStateVersion(BigInteger.valueOf(reader.readUByte()));
+    }
```
vikinatora: The convention we've used until now is to have the SCALE decoding functionality in a separate class

vikinatora (changes requested):
I've marked only the things that made the biggest impression on me. I haven't commented on the left over TODOs and unfinished logic

build.gradle.kts
bokoto000: It is used in the RuntimeBuilder at line 24

build.gradle.kts
bokoto000: com.google.common is part of it

src/main/java/com/limechain/sync/warpsync/ChainInformation.java
```diff
@@ -0,0 +1,37 @@
...
+    //aura_autorities_call_output
+    //aura_slot_duration_call_output
+    //babe_current_epoch_call_output
+    //babe_next_epoch_call_output
+    //babe_configuration_call_output
+    //grandpa_autorities_call_output
+    //grandpa_current_set_id_call_output
+
+    public boolean runtimeBabeApiIsV1() {
+        return runtimeBabeVersion == BigInteger.ONE;
```
georg-getz: Never equate BigIntegers with ==, use `.equals` instead

src/main/java/com/limechain/sync/warpsync/ChainInformation.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public boolean runtimeGrandpaSupportsCurrentSetId() {
+        return runtimeGrandpaVersion.equals(BigInteger.valueOf(3));
```
georg-getz: magic number 3

src/main/java/com/limechain/sync/warpsync/runtime/RuntimeApis.java
```diff
@@ -0,0 +1,35 @@
...
+        return runtimeApis;
+    }
+
+    public BigInteger getApiVersion(byte[] key) {
+        for (int i = 0; i < apiVersions.size(); i++) {
+            if (Arrays.equals(apiVersions.get(i), key)) {
+                return apiVersionsNumbers.get(i);
+            }
+        }
+        return BigInteger.valueOf(-1);
```
georg-getz: Is there a reason we prefer having a -1 BigInt instead of a `return null`?

src/main/java/com/limechain/sync/warpsync/runtime/RuntimeVersion.java
```diff
@@ -0,0 +1,53 @@
...
+@NoArgsConstructor
+public class RuntimeVersion {
+    String specName;
+    String implementation;
+    BigInteger authoringVersion;
+    BigInteger specVersion;
+    BigInteger implementationVersion;
+    BigInteger transactionVersion;
+    BigInteger stateVersion;
+    RuntimeApis runtimeApis;
```
georg-getz: Almost none of the fields in this PR are `private`. Would be good to add that everywhere where it's missing unless we have a good reason to leave the fields package private (default).

src/main/java/com/limechain/sync/warpsync/runtime/RuntimeVersion.java
```diff
@@ -0,0 +1,53 @@
...
+import io.emeraldpay.polkaj.scale.ScaleCodecReader;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+import java.math.BigInteger;
+
+@Getter
+@Setter
+@NoArgsConstructor
```
georg-getz: `@NoArgsConstructor` on a class with no constructor is a wasted line.

src/main/java/com/limechain/sync/warpsync/runtime/RuntimeVersion.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public String toString() {
+        return "RuntimeVersion:{specName=" + this.specName +
+                "; implementation=" + this.implementation +
+                "; authoringVersion=" + this.authoringVersion +
+                "; specVersion=" + this.specVersion +
+                "; implementationVersion=" + this.implementationVersion +
+                "; transactionVersion=" + this.transactionVersion +
+                "; stateVersion=" + this.stateVersion;
+    }
```
georg-getz: You can add closing curly braces at the end of the string.

src/main/java/com/limechain/sync/warpsync/runtime/WasmSections.java
```diff
@@ -0,0 +1,75 @@
...
+import io.emeraldpay.polkaj.scale.ScaleCodecReader;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.extern.java.Log;
+
+import java.util.Arrays;
+import java.util.logging.Level;
+
+@Log
+@NoArgsConstructor
```
georg-getz: remove

build.gradle.kts
vikinatora: If we're using it only for this line, maybe we can just copy the function locally, so that we don't bloat the build size?
```
        int x = Bytes.indexOf(wasmBinary, searchKey);
```
wdyt @georg-getz

build.gradle.kts
vikinatora: Okay, my bad

src/test/java/com/limechain/sync/warpsync/state/RuntimeDownloadTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+            hostBuilder1.addProtocols(List.of(new Ping(), lightMessages, kademliaService.getProtocol()));
+            senderNode = hostBuilder1.build();
+
+            senderNode.start().join();
+
+            kademliaService.setHost(senderNode);
+            var peerId = PeerId.fromBase58("12D3KooWFFqjBKoSdQniRpw1Y8W6kkV7takWv1DU2ZMkaA81PYVq");
+            var receivers = new String[]{
+//                    "/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWPyKTVdykB9iEXEggRkbMTV4qsWywhtT42qz16eBTvReA"
```
georg-getz: This can be deleted if it's not in use. Also some comments in this test would be nice as the code is not very easy to read.

src/main/java/com/limechain/sync/warpsync/runtime/RuntimeVersion.java
bokoto000: The runtime api versions are located in a different wasm section. Therefore it is not suitable to use the convention of the other readers for the runtime version.

src/main/java/com/limechain/sync/warpsync/runtime/RuntimeApis.java
bokoto000: I feel like setting the version to -1 is a stronger way of saying the hosts does not support the runtime api. Null on the other hand might suggest that the information was just not available at the time.

build.gradle.kts
georg-getz: We can copy it to decrease the amount of dependencies.

src/main/java/com/limechain/sync/warpsync/WarpSyncMachine.java
```diff
@@ -38,9 +39,10 @@ public class WarpSyncMachine {
     @Setter
     private PriorityQueue<Pair<BigInteger, Authority[]>> scheduledAuthorityChanges =
             new PriorityQueue<>(Comparator.comparing(Pair::getValue0));
+
     @Getter
     @Setter
-    private boolean isFinished;
+    private ChainInformation chainInformation = new ChainInformation();
```
georg-getz: Do we need a `@Setter`?

georg-getz (approved):

src/main/java/com/limechain/network/protocol/lightclient/LightMessages.java
```diff
@@ -43,7 +43,7 @@ public LightClientMessage.Response remoteReadRequest(Host us, AddressBook addrs,
                             StringUtils.remove0xPrefix(blockHash),
                             hexKeys)
                     .get();
-            log.log(Level.INFO, "Received response: " + resp.toString());
+            log.log(Level.INFO, "Received light client message response: " + resp.toByteArray().length);
```
vikinatora: Missing "... with length"

src/main/java/com/limechain/network/protocol/warp/WarpSync.java
```diff
@@ -23,7 +23,7 @@ public WarpSyncResponse warpSyncRequest(Host us, AddressBook addrs, PeerId peer,
         WarpSyncController controller = dialPeer(us, peer, addrs);
         try {
             WarpSyncResponse resp = controller.warpSyncRequest(blockHash).get(10, TimeUnit.SECONDS);
-            log.log(Level.INFO, "Received warp sync response: " + resp.toString());
+            log.log(Level.INFO, "Received warp sync response with " + resp.getFragments().length + " fragments");
```
vikinatora, line 26: Missing "... with length"

src/main/java/com/limechain/sync/warpsync/ChainInformation.java
```diff
@@ -0,0 +1,38 @@
...
+
+import lombok.Getter;
+import lombok.Setter;
+
+import java.math.BigInteger;
+
+@Getter
+@Setter
+public class ChainInformation {
+    public static final BigInteger GRANDPA_VERSION_SUPPORTING_CURRENT_SET_ID = BigInteger.valueOf(3);
```
vikinatora, line 11: Isn't it more correct `GRANDPA_CURRENT_SET_ID` ?

src/main/java/com/limechain/sync/warpsync/runtime/Runtime.java
```diff
@@ -0,0 +1,36 @@
+package com.limechain.sync.warpsync.runtime;
```
vikinatora: I think the Runtime logic shouldn't be inside `warpsync` folder. It's a separate module itself.

vikinatora (approved):
Apart from the missing javadoc for some classes and a few wording improvements, I think it's ok

src/main/java/com/limechain/sync/warpsync/ChainInformation.java
bokoto000: I think GRANDPA_CURRENT_SET_ID would not make it clear that it is a grandpa version not a set id

PR #112, opened 2023-07-05
`BlockAnnounceHandshake` and `currentSelectedPeerId` Improvements

vikinatora:
## Changes:
- Fruzhin now sends a real `BlockAnnounceHandshake` based on the sync data up until that point.
- `currentSelectedPeerId` gets selected after we establish a connection with at least one node.
- Warp sync waits for at least one connection and `currentSelectedPeerId` to exist before starting.
- Added `GenesisHash` class to hold all genesis hashes since they're static.
- Fixed an issue where when `db-recreate` param is passed, the node crashes because it tries to delete an non-existing folder.
- Improved README to feature local development instructions.
- Renamed a few variables and fields in order to improve readability.

## Suggestions wanted:
- How to improve light client's start function in order to avoid the infinite loop. Had to remove an unit test because it hangs due to the infinite nature of the function

### DIFF
--- a/src/main/java/com/limechain/storage/DBRepository.java
+++ b/src/main/java/com/limechain/storage/DBRepository.java
@@ -56,8 +56,8 @@ public class DBRepository implements KVRepository<String, Object> {
         try {
             if (file.exists()) {
                 FileUtils.cleanDirectory(file.getAbsoluteFile());
+                log.log(Level.INFO, "\uD83D\uDDD1️DB cleaned");
             }
-            log.log(Level.INFO, "\uD83D\uDDD1️DB cleaned");
 
         } catch (IOException e) {
             log.log(Level.SEVERE, String.format("Error deleting db folder. Exception: '%s', message: '%s'",
--- a/src/main/java/com/limechain/storage/DBRepository.java
+++ b/src/main/java/com/limechain/storage/DBRepository.java
@@ -54,7 +54,9 @@ public class DBRepository implements KVRepository<String, Object> {
 
     private void cleanDatabaseFolder(File file) {
         try {
-            FileUtils.cleanDirectory(file.getAbsoluteFile());
+            if (file.exists()) {
+                FileUtils.cleanDirectory(file.getAbsoluteFile());
+            }
             log.log(Level.INFO, "\uD83D\uDDD1️DB cleaned");
 
         } catch (IOException e) {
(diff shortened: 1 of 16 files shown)


src/main/java/com/limechain/storage/DBRepository.java
```diff
@@ -54,7 +54,9 @@ public DBRepository(String path, String chain, boolean dbRecreate) {
 
     private void cleanDatabaseFolder(File file) {
         try {
-            FileUtils.cleanDirectory(file.getAbsoluteFile());
+            if (file.exists()) {
+                FileUtils.cleanDirectory(file.getAbsoluteFile());
+            }
             log.log(Level.INFO, "\uD83D\uDDD1️DB cleaned");
```
georg-getz: I think the log should be in the `if` body as well.

src/main/java/com/limechain/sync/warpsync/SyncedState.java
```diff
@@ -0,0 +1,64 @@
...
+
+        Hash256 lastFinalizedBlockHash = this.getLastFinalizedBlockHash() == null
+                ? genesisBlockHash
+                : this.getLastFinalizedBlockHash();
+        return new BlockAnnounceHandshake(
+                4,
+                this.getLastFinalizedBlockNumber().toString(),
+                lastFinalizedBlockHash,
+                genesisBlockHash
+        );
```
georg-getz: 4 is a magic number.

src/main/java/com/limechain/sync/warpsync/SyncedState.java
```diff
@@ -0,0 +1,64 @@
...
+    @Getter
+    @Setter
+    public byte[] runtime;
+    @Getter
+    @Setter
+    public byte[] heapPages;
+
+    public static SyncedState getINSTANCE() {
+        return INSTANCE;
+    }
```
georg-getz: I think the function should be camel case regardless of the field being fully capitalized.

src/main/java/com/limechain/sync/warpsync/SyncedState.java
```diff
@@ -0,0 +1,64 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+import java.math.BigInteger;
+
+public class SyncedState {
+    private static final SyncedState INSTANCE = new SyncedState();
+    @Getter
+    @Setter
+    public Hash256 lastFinalizedBlockHash;
```
stoyan-lime: I think it is a better practice to declare variables as `private` and use the getters/setters to access them.

src/main/java/com/limechain/sync/warpsync/SyncedState.java
vikinatora: Yes, not sure how it turned out to be this way

georg-getz (approved):

stoyan-lime (approved):

PR #114, opened 2023-07-17
Add full sync after warp sync is finished. Add Grandpa protocol.

stoyan-lime:
**Full Sync**
Added logic for verifying and updating latest block in SyncedState.
The expected flow is:
1. Grandpa handshake received.
2. Grandpa handshake sent back
3. Neighbour message received
4. Neighbour message sent back (needs to be sent when set id is updated as well)
5. Commit message received
6. Commit message justifications verified (currently always true)
7. State updated

Problems:
- Commits are only received from local debug polkadot node
- Set id does not match the set id of nodes we connect to

Problems for syncing through block announce:
- Sometimes we get block announce from peers not in our address book, so we get an error when trying to send a block request back to them (No addresses for peer)
- When debugging I did not seem to get a response to block requests (no timeout either)


src/main/java/com/limechain/network/ProtocolUtils.java
```diff
@@ -24,4 +24,11 @@ public static String getLegacyBlockAnnounceProtocol(String chainId) {
     public static String getLegacyKadProtocol(String chainId) {
         return String.format("/%s/kad", chainId);
     }
+
+    public static String getGrandpaLegacyProtocol(String chainId) {
+        if (chainId.equals("dot")) {
+            chainId = "paritytech";
+        }
+        return String.format("/%s/grandpa/1", chainId);
```
vikinatora: I think it should always be `/paritytech/grandpa/1`

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
```diff
@@ -0,0 +1,104 @@
...
+            return;
+        }
+
+        switch (messageType) {
+            case HANDSHAKE -> handleHandshake(msg, peerId, stream);
+            case NEIGHBOUR -> handleNeighbourMessage(msg, peerId, stream);
+            case COMMIT -> handleCommitMessage(msg, peerId);
+            default -> log.log(Level.WARNING, "Unknown grandpa message type from Peer " + peerId);
+        }
+    }
```
vikinatora: I think we should add the other message types as well and only log that they've been received

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
```diff
@@ -0,0 +1,104 @@
...
+        return GrandpaMessageType.getByType(msg[0]);
+    }
+
+    private void handleHandshake(byte[] msg, PeerId peerId, Stream stream) {
+        PeerInfo peerInfo = connectionManager.getPeerInfo(peerId);
+
+        if (peerInfo != null && peerInfo.isGrandpaConnected()) {
+            log.log(Level.INFO, "Received existing grandpa handshake from " + peerId);
+        } else {
+            if(peerInfo == null) {
```
vikinatora: Inconsistet if brackets formating - sometimes `if(` and other times `if (`. Other parts of the code should be fixed as well

src/main/java/com/limechain/network/protocol/grandpa/GrandpaProtocol.java
```diff
@@ -0,0 +1,76 @@
...
+import io.netty.buffer.ByteBuf;
+import io.netty.handler.codec.bytes.ByteArrayEncoder;
+import lombok.extern.java.Log;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.concurrent.CompletableFuture;
+
+@Log
+public class GrandpaProtocol extends ProtocolHandler<GrandpaController> {
+    public static final int MAX_HANDSHAKE_SIZE = 1024 * 1024 * 1024;
```
vikinatora: Should be 4
Source: https://github.com/smol-dot/smoldot/blob/c04fe69400c530b51943a180043b473849a7da79/lib/src/network/service/notifications.rs#L104-L105

src/main/java/com/limechain/network/protocol/grandpa/messages/commit/CommitMessageScaleReader.java
```diff
@@ -0,0 +1,32 @@
...
+        commitMessage.setVote(new VoteScaleReader().read(reader));
+
+        //TODO: check parsing after we start to receive more than 1 authority
+        int precommitsCount = reader.readCompactInt();
+        Precommit[] precommits = new Precommit[precommitsCount];
+        PrecommitReader precommitReader = new PrecommitReader();
+        for (int i = 0; i < precommitsCount; i++) {
+            precommits[i] = precommitReader.read(reader);
+        }
+        commitMessage.setPrecommits(precommits);
```
vikinatora: Can be replaced with `ListReader`. See other scale readers for example

src/main/java/com/limechain/network/protocol/grandpa/messages/neighbour/NeighbourMessageScaleReader.java
```diff
@@ -0,0 +1,22 @@
...
+import io.emeraldpay.polkaj.scale.ScaleCodecReader;
+import io.emeraldpay.polkaj.scale.ScaleReader;
+import io.emeraldpay.polkaj.scale.reader.UInt64Reader;
+
+public class NeighbourMessageScaleReader implements ScaleReader<NeighbourMessage> {
+    @Override
+    public NeighbourMessage read(ScaleCodecReader reader) {
+        int messageType = reader.readByte();
+        if (messageType != GrandpaMessageType.NEIGHBOUR.getType()) {
+            return null;
```
vikinatora: Maybe throw an error?

src/main/java/com/limechain/sync/warpsync/SyncedState.java
```diff
@@ -69,4 +97,81 @@ public BlockAnnounceHandshake getHandshake() {
...
     }
 
+    public NeighbourMessage getNeighbourMessage() {
+        NeighbourMessage message = new NeighbourMessage();
+        message.setVersion(NEIGHBOUR_MESSAGE_VERSION);
+        message.setSetId(this.setId);
+        message.setRound(NEIGHBOUR_MESSAGE_LIGHT_CLIENT_ROUND);
+        message.setLastFinalizedBlock(this.lastFinalizedBlockNumber.longValue());
+        return message;
+    }
```
vikinatora: Maybe `NeighbourMessage` should have `AllArgsConstructor`?

vikinatora (changes requested):
Left some comments

src/main/java/com/limechain/network/ConnectionManager.java
```diff
@@ -25,22 +27,42 @@ public void addPeer(PeerId peerId, PeerInfo peerInfo) {
...
+       return peers.get(peerId);
+    }
+
     public void updatePeer(PeerId peerId, BlockAnnounceMessage blockAnnounceMessage) {
-        if (!blockAnnounceMessage.isBestBlock()) {
-            return;
+        final var peer = peers.get(peerId);
+        updateLatestBlock(peer, blockAnnounceMessage.getHeader().getBlockNumber());
+
+        if(blockAnnounceMessage.isBestBlock()) {
```
georg-getz: Put a space between 'if' and '(', this goes for all such occurrences. Might be best to see if this can be added to the checkstyle as a rule so that it's more easily checked.

src/main/java/com/limechain/network/Network.java
```diff
@@ -45,6 +51,8 @@ public class Network {
     public LightMessagesService lightMessagesService;
     public KademliaService kademliaService;
     public BlockAnnounceService blockAnnounceService;
+
+    public GrandpaService grandpaService;
     public Ping ping;
```
georg-getz: Why separate this service from the others with a new line?

src/main/java/com/limechain/network/StrictProtocolBinding.java
```diff
@@ -11,7 +11,7 @@ public StrictProtocolBinding(String protocolId, P2PChannelHandler<T> protocol) {
         super(protocolId, protocol);
     }
 
-    public T dialPeer(Host us, PeerId peer, AddressBook addrs) {
+    public synchronized T dialPeer(Host us, PeerId peer, AddressBook addrs) {
```
georg-getz: Are we sure we want this to be synchronized? How long is the timeout of `dial`? If we need the synchronization what's the drawback of having a synchronized block just around `dial` instead of the whole function being synchonized?
Also always add javadoc to synchronized functions as they are very prone to breaking/causing deadlocks.

src/main/java/com/limechain/network/protobuf/SyncMessage.proto
```diff
@@ -58,4 +57,47 @@ message BlockData {
   // doesn't make in possible to differentiate between a lack of justification and an empty
   // justification.
   bool is_empty_justification = 7; // optional, false if absent
+  // Justifications if requested.
+  // Unlike the field for a single justification, this field does not required an associated
```
georg-getz: require*

src/main/java/com/limechain/network/protocol/grandpa/GrandpaProtocol.java
```diff
@@ -0,0 +1,75 @@
...
+        public NotificationHandler(GrandpaEngine engine, Stream stream) {
+            this.engine = engine;
+            this.stream = stream;
+        }
+        @Override
+        public void onMessage(@NotNull Stream stream, ByteBuf msg) {
+            byte[] messageBytes = new byte[msg.readableBytes()];
+            msg.readBytes(messageBytes);
+            engine.receiveRequest(messageBytes, stream.remotePeerId(), stream);
+        }
```
georg-getz, line 62: How do we know that the stream cannot be null?

src/main/java/com/limechain/sync/warpsync/SyncedState.java
```diff
@@ -69,4 +98,81 @@ public BlockAnnounceHandshake getHandshake() {
...
+        if (!warpSyncFinished || blockNumber.compareTo(lastFinalizedBlockNumber) <= 0) {
+            return;
+        }
+
+        final var blockSyncResponse = network.syncBlock(peerId, blockNumber, lastFinalizedBlockNumber);
+        verifyBlockResponse(blockSyncResponse).ifPresentOrElse(
+                this::updateLatestBlockData,
+                () -> log.log(Level.WARNING, "Failed verifying block " + blockNumber + " from peer " + peerId)
+        );
+    }
```
georg-getz: Add javadoc

src/main/java/com/limechain/network/protocol/grandpa/GrandpaProtocol.java
stoyan-lime: This was misnamed and is actually the traffic limit of streams we initiate.

src/main/java/com/limechain/network/protocol/grandpa/messages/commit/CommitMessageScaleReader.java
stoyan-lime: Updated with compact justification reader.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaProtocol.java
stoyan-lime: The method being overridden is a Kotlin method with a non-null type.

src/main/java/com/limechain/sync/warpsync/SyncedState.java
stoyan-lime: Unused method removed.

src/main/java/com/limechain/network/StrictProtocolBinding.java
stoyan-lime: Not needed. Removed.

src/main/java/com/limechain/network/protocol/grandpa/messages/neighbour/NeighbourMessageScaleReader.java
stoyan-lime: Done.

src/main/java/com/limechain/network/ConnectionManager.java
```diff
@@ -25,22 +27,42 @@ public void addPeer(PeerId peerId, PeerInfo peerInfo) {
...
+
+        if (blockAnnounceMessage.isBestBlock()) {
+            updateBestBlock(peer, blockAnnounceMessage.getHeader());
         }
+    }
 
-        final var peer = peers.get(peerId);
-        peer.setBestBlock(blockAnnounceMessage.getHeader().getBlockNumber().toString());
-        peer.setBestBlockHash(new Hash256(blockAnnounceMessage.getHeader().getHash()));
+    // TODO: decide if needed
```
vikinatora: What should we do with this TODO?

src/main/java/com/limechain/network/protocol/grandpa/Grandpa.java
```diff
@@ -0,0 +1,24 @@
+package com.limechain.network.protocol.grandpa;
+
+import com.limechain.network.StrictProtocolBinding;
+import io.libp2p.core.AddressBook;
+import io.libp2p.core.Host;
+import io.libp2p.core.PeerId;
+import lombok.extern.java.Log;
+
+@Log
```
vikinatora: Log isn't used

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        switch (messageType) {
+            case HANDSHAKE -> handleHandshake(msg, peerId, stream);
+            case VOTE -> log.log(Level.INFO, "Vote message received from Peer " + peerId);
+            case COMMIT -> handleCommitMessage(msg, peerId);
+            case NEIGHBOUR -> handleNeighbourMessage(msg, peerId);
+            case CATCH_UP_REQUEST -> log.log(Level.INFO, "Catch up request received from Peer " + peerId);
+            case CATCH_UP_RESPONSE -> log.log(Level.INFO, "Catch up response received from Peer " + peerId);
+            default -> log.log(Level.WARNING, "Unknown grandpa message type from Peer " + peerId);
```
vikinatora: Maybe add message type to the log

src/main/java/com/limechain/network/protocol/grandpa/GrandpaProtocol.java
```diff
@@ -0,0 +1,74 @@
...
+import io.netty.buffer.ByteBuf;
+import io.netty.handler.codec.bytes.ByteArrayEncoder;
+import lombok.extern.java.Log;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.concurrent.CompletableFuture;
+
+@Log
+public class GrandpaProtocol extends ProtocolHandler<GrandpaController> {
+    public static final int TRAFFIC_LIMIT = 1024 * 1024;
```
vikinatora, line 17: I'd prefer having two variables in order to follow the same coding style as for the other protocols

src/main/java/com/limechain/network/protocol/grandpa/GrandpaProtocol.java
vikinatora: It's fine IMO

src/main/java/com/limechain/network/protocol/grandpa/messages/commit/CompactJustificationScaleReader.java
```diff
@@ -0,0 +1,36 @@
+package com.limechain.network.protocol.grandpa.messages.commit;
+
+import com.limechain.network.protocol.warp.dto.Precommit;
+import io.emeraldpay.polkaj.scale.ScaleCodecReader;
+import io.emeraldpay.polkaj.scale.ScaleReader;
+import io.emeraldpay.polkaj.types.Hash256;
+import io.emeraldpay.polkaj.types.Hash512;
+
+public class CompactJustificationScaleReader implements ScaleReader<Precommit[]> {
```
vikinatora, line 9: Maybe it's justified but reads strange because the name is `CompactJustificationScaleReader` but implements `ScaleReader<Precommit[]>`

src/main/java/com/limechain/sync/warpsync/SyncedState.java
```diff
@@ -69,4 +91,34 @@ public BlockAnnounceHandshake getHandshake() {
...
+    public void syncCommit(CommitMessage commitMessage, PeerId peerId) {
+        if (!verifyCommitJustification(commitMessage)) {
+            log.log(Level.WARNING, "Could not verify commit from peer: " + peerId);
+            return;
+        }
+
+        if (warpSyncFinished) {
+            updateState(commitMessage);
+        }
+    }
```
vikinatora: Not sure it's good practice to have a function `verifyCommitJustification` inside `SyncedState`. Since it's just a wrapper around `JustificationVerifier.verify` why not just invoke the function directly?

vikinatora (commented):
Requesting some minor comsetic changes. Some are preferrences to you can decide if it's ok or not to implement them

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
georg-getz: It's already there, no? `CATCH_UP_REQUEST` logs "request" and the other - "response"

src/main/java/com/limechain/network/protocol/grandpa/GrandpaProtocol.java
georg-getz: I too prefer using 1 variable instead of 2 wherever possible, but it would be nice to update the other protocols as well. In `BlockAnounceProtocol` for example we have 2 separate fields holding the same value.

src/main/java/com/limechain/network/protocol/grandpa/messages/commit/CompactJustificationScaleReader.java
georg-getz: I don't see this as a strange read. What would you change it to?

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
vikinatora: I meant for the default case. Probably won't be applicable as these are all of the message types

stoyan-lime: Added the first byte of the message, that should specify the type, to the log

src/main/java/com/limechain/network/protocol/grandpa/messages/commit/CompactJustificationScaleReader.java
stoyan-lime: My thinking was, that this is consistent with the polkadot documentation where the Compact Justification is an optimized way to store precommits. So this is a reader for compact justifications that produces an array of precommits.

src/main/java/com/limechain/sync/warpsync/SyncedState.java
stoyan-lime: Removed

vikinatora (approved):

PR #121, opened 2023-08-15
Unify all unsigned integers to be of type BigInteger

### LINKED ISSUE #85: Unify all unsigned integers to be of type BigInteger
Right now, there's a mix of `long` and `BigInteger` when reading `u32`, `u64`, `u128` using `ScaleReader`. 

We aim to have all unsigned integers of one type - `BigInteger`. Although using `BigInteger` provides some overhead, the code looks cleaner, makes it easier(and safer) to execute arithmetic operations on them.

The assignee of the task should go though all `ScaleReader` implementations and work their way through converting all `long` fields to `BigInteger` and adjusting the reading(setter) logic.

ablax:
* change long values to BigInteger for uint32 readers

* change BlockAnnounceHandshake fields to private and user proper getters and setters

Closes #85

### DIFF
--- a/src/main/java/com/limechain/network/protocol/blockannounce/scale/BlockAnnounceHandshake.java
+++ b/src/main/java/com/limechain/network/protocol/blockannounce/scale/BlockAnnounceHandshake.java
@@ -15,4 +15,14 @@ public class BlockAnnounceHandshake {
     private BigInteger bestBlock;
     private Hash256 bestBlockHash;
     private Hash256 genesisBlockHash;
+
+    @Override
+    public String toString() {
+        return "BlockAnnounceHandShake{" +
+                "nodeRole=" + nodeRole +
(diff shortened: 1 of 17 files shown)


src/main/java/com/limechain/network/protocol/blockannounce/scale/BlockAnnounceHandshake.java
```diff
@@ -4,15 +4,19 @@
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+import java.math.BigInteger;
 
 @Getter
+@Setter
```
stoyan-lime: We can use @Data to replace @Getter @Setter and the toString method

vikinatora (approved):

stoyan-lime (approved):

PR #129, opened 2023-08-16
Increase GRANDPA traffic limit and synchronize update state

stoyan-lime:
Increase GRANDPA traffic limit to Long.MAX_VALUE, make updateState, used when receiving commit messages, synchronized and handle receiveMessage NPE.

### DIFF
--- a/src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
+++ b/src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
@@ -30,6 +30,11 @@ public class GrandpaEngine {
         boolean connectedToPeer = connectionManager.isGrandpaConnected(peerId);
         GrandpaMessageType messageType = getGrandpaMessageType(msg);
 
+        if (messageType == null) {
+            log.log(Level.WARNING, String.format("Unknown grandpa message type \"%d\" from Peer %s", msg[0], peerId));
+            return;
+        }
+
         if (!connectedToPeer && messageType != GrandpaMessageType.HANDSHAKE) {
             log.log(Level.WARNING, "No handshake for grandpa message from Peer " + peerId);
             return;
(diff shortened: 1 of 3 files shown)


src/main/java/com/limechain/sync/warpsync/SyncedState.java
```diff
@@ -112,7 +112,10 @@ public void syncCommit(CommitMessage commitMessage, PeerId peerId) {
         }
     }
 
-    private void updateState(CommitMessage commitMessage) {
+    private synchronized void updateState(CommitMessage commitMessage) {
```
vikinatora, line 115: Why should this method be `synchronized`

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
```diff
@@ -87,7 +90,7 @@ private void handleCommitMessage(byte[] msg, PeerId peerId) {
         ScaleCodecReader reader = new ScaleCodecReader(msg);
         CommitMessage commitMessage = reader.read(new CommitMessageScaleReader());
         if (isBlockAlreadyReached(commitMessage.getVote().getBlockNumber())) {
-            log.log(Level.INFO, String.format("Received commit message for already reached block %d from peer %s",
+            log.log(Level.INFO, String.format("Received commit message for finalized block %d from peer %s",
```
vikinatora: No need to spam the logs with this message IMO. Level.FINE would be more suitable.

vikinatora (approved):

src/main/java/com/limechain/sync/warpsync/SyncedState.java
stoyan-lime: We are receiving commit messages from multiple peers at the same time and they start updating the state at the same time. We were updating the finalized block multiple times for the same block.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
stoyan-lime: Updated.

PR #135, opened 2023-08-18
Network module refactor

ablax:


src/main/java/com/limechain/network/Network.java
```diff
@@ -80,29 +80,12 @@ public class Network {
      * @param repository
      * @param cliArgs
      */
-    private Network(ChainService chainService, HostConfig hostConfig, KVRepository<String, Object> repository,
+    public Network(ChainService chainService, HostConfig hostConfig, KVRepository<String, Object> repository,
```
vikinatora, line 83: Remove "Singleton" from "A Singleton Network class that handles all peer connections and Kademlia" at the start of the file

vikinatora (approved):

PR #142, opened 2023-08-23
Persist finalized state and be able to load it from DB

### LINKED ISSUE #136: Persist finalized state and be able to load it from DB
Fruzhin should persist the state trie and new finalized state to the DB everytime it verifies a new block.(Think about whether it should overwrite the old state or create a new one)

Also think about whether we should save peer info and be able to load it from the DB, as well as other Host information

ablax:
* Save and load synced state on commit
* Save lightSyncState and load from db
* Save the latest trie in db

Closes #136


src/main/java/com/limechain/network/Network.java
```diff
@@ -85,7 +85,6 @@ public Network(ChainService chainService, HostConfig hostConfig, KVRepository<St
         this.initializeProtocols(chainService, hostConfig, repository, cliArgs);
         this.bootNodes = chainService.getGenesis().getBootNodes();
         this.chain = hostConfig.getChain();
-        SyncedState.getInstance().setNetwork(network);
```
vikinatora, line 88: Why is this removed?

src/main/java/com/limechain/network/protocol/blockannounce/BlockAnnounceEngine.java
```diff
@@ -68,6 +68,7 @@ private void handleBlockAnnounce(byte[] msg, PeerId peerId) {
                 " with hash:0x" + HexUtils.toHexString(announce.getHeader().getHash()) +
                 " parentHash:" + announce.getHeader().getParentHash() +
                 " stateRoot:" + announce.getHeader().getStateRoot());
+        //TODO: Should update Trie (and merkle proofs?) and save them in db
```
vikinatora, line 71: _**I think that**_ if the state root of the header has changed the host should send out a state request for the block hash(since light client can't process extrinsics) and update it's trie.

IMO this should be in a separate task in order to not bloat this PR

src/main/java/com/limechain/sync/warpsync/SyncedState.java
```diff
@@ -120,5 +106,60 @@ private synchronized void updateState(CommitMessage commitMessage) {
         lastFinalizedBlockHash = commitMessage.getVote().getBlockHash();
         lastFinalizedBlockNumber = commitMessage.getVote().getBlockNumber();
         log.log(Level.INFO, "Reached block #" + lastFinalizedBlockNumber);
+        saveState();
+    }
+
+    public void saveState() {
```
vikinatora: let's change the function name to `persistState`

src/main/java/com/limechain/storage/DBConstants.java
```diff
@@ -9,4 +9,19 @@ public class DBConstants {
...
+    /**
+     * Key under which the --latest-- state trie proof is stored
+     * TODO: Currently only the latest loaded in sync is stored
+     */
+    public static final String STATE_TRIE_PROOF = "stateTrieProof";
+    /**
+     * Key under which the --latest-- state trie root state is stored
+     * TODO: Currently only the latest loaded in sync is stored
+     */
+    public static final String STATE_TRIE_ROOT_STATE = "stateTrieRootState";
```
vikinatora: let's rename it to `STATE_TRIE_ROOT_HASH`

src/main/java/com/limechain/storage/DBConstants.java
```diff
@@ -9,4 +9,19 @@ public class DBConstants {
...
     public static final String GENESIS_KEY = "genesis";
+    /**
+     * Key under which the latest sync state is stored
+     */
+    public static final String SYNC_STATE_KEY = "syncState";
+    /**
+     * Key under which the --latest-- state trie proof is stored
+     * TODO: Currently only the latest loaded in sync is stored
+     */
+    public static final String STATE_TRIE_PROOF = "stateTrieProof";
```
vikinatora: let's change it to `STATE_TRIE_MERKLE_PROOF`

vikinatora (changes requested):
Some minor changes are requested

src/main/java/com/limechain/network/Network.java
ablax: It's not used anywhere

src/main/java/com/limechain/network/protocol/blockannounce/BlockAnnounceEngine.java
ablax: I agree that it should be in separate task, that's why I left it with todo

src/main/java/com/limechain/sync/warpsync/WarpSyncMachine.java
```diff
@@ -64,7 +64,7 @@ public void start() {
         final Hash256 initStateHash;
         if (this.chainService.getGenesis().getLightSyncState() != null) {
             this.syncedState.setRepository(repository);
-            boolean stateLoaded = this.syncedState.loadState();
+            boolean stateLoaded = !chainService.isLocalChain() && this.syncedState.loadState();
```
vikinatora: Not sure if this is expected behaviour. In most cases this would be the intuitive solution but it would confuse some users that want to restart their Host and continue with the sync.

I'd remove `!chainService.isLocalChain()`. If users want to reset their DB, they can use the `-dbc` flag

src/main/java/com/limechain/network/Network.java
vikinatora: I answered my own question - it's not used. You can resolve this comment

vikinatora (approved):

PR #181, opened 2023-09-08
Update runtime during sync

stoyan-lime:
When receiving block announces, check header digest for a runtime update message and if present schedule the update.

When block with scheduled runtime update is received in a commit message and verified, update the runtime as well. LightMessages is used to request the update.

To be added:
- [ ] Tests


src/main/java/com/limechain/sync/warpsync/SyncedState.java
```diff
@@ -162,9 +193,101 @@ private void updateState(CommitMessage commitMessage) {
...
+            }
+            //TODO Heap pages should be fetched from out storage
+            setRuntimeCode(code);
+            log.log(Level.INFO, "Runtime and heap pages downloaded");
+
+        } catch (TrieDecoderException e) {
+            throw new RuntimeCodeException("Couldn't build trie from proofs list: " + e.getMessage());
+        }
+    }
+
```
ablax: Maybe move these methods to a separate RuntimeBuilder class

vikinatora (approved):

src/main/java/com/limechain/sync/warpsync/dto/RuntimeCodeException.java
```diff
@@ -0,0 +1,7 @@
+package com.limechain.sync.warpsync.dto;
+
+public class RuntimeCodeException extends Exception {
+    public RuntimeCodeException(String errorMessage) {
+        super(errorMessage);
+    }
+}
```
georg-getz: `Exception` is a checked exception - never extend it. Instead extend `RuntimeException`

src/main/java/com/limechain/sync/warpsync/SyncedState.java
```diff
@@ -162,9 +194,102 @@ private void updateState(CommitMessage commitMessage) {
...
+                .orElseThrow(() -> new RuntimeCodeException("No available merkle proof"));
+        Hash256 stateRoot = repository.find(DBConstants.STATE_TRIE_ROOT_HASH)
+                .map(storedRootState -> Hash256.from(storedRootState.toString()))
+                .orElseThrow(() -> new RuntimeCodeException("No available state root"));
+
+        setCodeAndHeapPages(merkleProof, stateRoot);
+    }
+
+    private void setCodeAndHeapPages(byte[][] decodedProofs, Hash256 stateRoot) throws RuntimeCodeException {
+        Trie trie;
```
georg-getz: No need to define this outside of the `try` block since it's not used outside of it.

src/main/java/com/limechain/sync/warpsync/state/RuntimeBuildState.java
```diff
@@ -13,7 +11,7 @@
  */
 @Log
 public class RuntimeBuildState implements WarpSyncState {
-    private final SyncedState syncedState = SyncedState.getInstance();
+    protected SyncedState syncedState = SyncedState.getInstance();
```
georg-getz: What's the point of this change? `RuntimeBuildState` has no children right?

src/main/java/com/limechain/sync/warpsync/SyncedState.java
georg-getz: Don't forget to remove `RuntimeCodeException` from any throws clauses when you change it to extend `RuntimeException`.

src/main/java/com/limechain/sync/warpsync/state/RuntimeBuildState.java
stoyan-lime: The change is to make the field available for testing.

vikinatora (approved):

src/main/java/com/limechain/sync/warpsync/state/RuntimeBuildState.java
ablax: You can keep it private and add

@AllArgsConstructor
@NoArgsConstructor

It will work fine and you will have no warning messages that it's not final.

georg-getz: It would be better to add a `@Getter` on the field and access is through that instead of changing its access modifier.

stoyan-lime: Updated.

georg-getz (approved):

PR #192, opened 2023-09-27
Incoming requests for /sync/2 substream

### LINKED ISSUE #162: Incoming requests for /sync/2 substream
- [ ] Handle incoming block requests
- [ ] Handle incoming state requests

ablax:
* Handle incoming requests for /sync/2 substream

### DIFF
--- a/src/main/java/com/limechain/network/protocol/sync/SyncMessages.java
+++ b/src/main/java/com/limechain/network/protocol/sync/SyncMessages.java
@@ -2,12 +2,15 @@ package com.limechain.network.protocol.sync;
 
 import com.limechain.network.StrictProtocolBinding;
 import com.limechain.network.protocol.sync.pb.SyncMessage;
+import com.limechain.utils.StringUtils;
 import io.libp2p.core.AddressBook;
 import io.libp2p.core.Host;
 import io.libp2p.core.PeerId;
 import lombok.extern.java.Log;
 
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
 
 @Log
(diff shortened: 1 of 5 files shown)


src/main/java/com/limechain/network/protocol/sync/SyncMessages.java
```diff
@@ -34,4 +37,17 @@ public SyncMessage.BlockResponse remoteBlockRequest(Host us, AddressBook addrs,
             throw new RuntimeException(e);
         }
     }
+
+    public SyncMessage.StateResponse remoteStateRequest(Host us, AddressBook addrs, PeerId peer, String blockHash) {
+        SyncController controller = dialPeer(us, peer, addrs);
+        try {
+            SyncMessage.StateResponse resp = controller.sendStateRequest(StringUtils.remove0xPrefix(blockHash))
+                    .get(30, TimeUnit.SECONDS);
```
vikinatora: 30 seconds is too much IMO, 10 would be more suitable

vikinatora (approved):

src/main/java/com/limechain/network/protocol/sync/SyncProtocol.java
```diff
@@ -21,25 +26,35 @@ public SyncProtocol() {
         super(MAX_REQUEST_SIZE, MAX_RESPONSE_SIZE);
     }
 
-    @Override
-    protected CompletableFuture<SyncController> onStartInitiator(Stream stream) {
+    @NotNull
```
georg-getz: Do we need the `@NotNull`? It's only a cosmetic annotation if we aren't using some kind of Intellij static analysis.

georg-getz: Goes for the `@NotNull`s in method args as well.

src/main/java/com/limechain/network/protocol/sync/SyncProtocol.java
```diff
@@ -68,4 +106,59 @@ public void onException(Throwable cause) {
...
+//            stream.writeAndFlush(stateResponse);
+            stream.closeWrite();
+        }
+
+        private void handleBlockRequest(SyncMessage.BlockRequest blockRequest) {
+            //Todo: create block response (must get the block from the storage)
+//            SyncMessage.BlockResponse blockResponse = SyncMessage.BlockResponse.newBuilder().build();
+//            stream.writeAndFlush(blockResponse);
+            stream.closeWrite();
+        }
```
georg-getz: The comments here should either be uncommented or deleted

src/main/java/com/limechain/network/protocol/sync/SyncProtocol.java
ablax: Well, we don't really need it, but the static analyzer complains because the methods onStartInitiator and onStartResponder both override a method that has NotNull annotation. It's pretty much the same as skipping the Override annotation on one of both methods. It will work without them but you will be safer with them,

georg-getz (approved):

PR #198, opened 2023-10-11
Implement Allocator API

### LINKED ISSUE #168: Implement Allocator API

stoyan-lime:
Created a Freeing Bump Allocator to handle heap memory.
Allocator allocates only blocks of sizes equal to powers of two. If we want to allocate a size that is not a power of 2, we allocate the next power of 2.
A header is added before each allocated block, containing information about it. 

To be added:
- [x] API implementation
- [x] Tests

Closes #168


src/main/java/com/limechain/runtime/allocator/FreeingBumpHeapAllocator.java
```diff
@@ -0,0 +1,226 @@
...
+        int requiredPages = pagesFromSize(size);
+        int currentPages = pagesFromSize(memory.buffer().limit());
+
+        if (currentPages >= MAX_WASM_PAGES) {
+            throw new AllocationError();
+        }
+
+        if (requiredPages > MAX_WASM_PAGES) {
+            throw new AllocationError();
+        }
```
georg-getz: Could add a message to the errors.

src/main/java/com/limechain/runtime/allocator/Header.java
```diff
@@ -0,0 +1,81 @@
...
+    private static final long FREE_HEADER_MASK = 0x00000000_00000000L;
+
+    private final boolean occupied;
+
+    private Integer nextFreeHeaderPointer;
+    private Integer order;
+
+    private Header(boolean occupied) {
+        this.occupied = occupied;
+    }
```
georg-getz: this could be a `@RequiredArgsConstructor` with accessibility argument set to private

src/main/java/com/limechain/runtime/allocator/Order.java
```diff
@@ -0,0 +1,59 @@
...
+    public static final int MIN_POSSIBLE_ALLOCATION = 8; // 2^3 bytes, 8 bytes
+    public static final int MAX_POSSIBLE_ALLOCATION = 32 * 1024 * 1024; // 2^25 bytes, 32 MiB
+
+    private final int value;
+    private final int blockSize;
+    /**
+     * Pointer to the next free block of this order.
+     */
+    @Nullable
+    private Integer freeHeaderPointer;
```
georg-getz: No need to specify `@Nullable` on Integer fields

src/main/java/com/limechain/runtime/hostapi/HostApi.java
```diff
@@ -74,6 +76,10 @@ protected static Memory getMemory() {
         return runtime.getInstance().exports.getMemory("memory");
     }
 
+    protected static int getHeapBase() {
+       return runtime.getInstance().exports.getGlobal("heap_base").getIntValue();
```
georg-getz: The probably won't work as the heap base global is preceded by 2 underscores.
Also you can create an enum with the commonly exported WASM globals, `__heap_base` and `__data_end`. You can probably also add `memory` there as that would most likely always be the name of the exported/imported memory.

georg-getz:
Would be good to add a test showcasing that the reading of wasm globals is correct. This will require a wasm file and a wat file (since wasm is unreadable) to be added to test resources and have them being read to check wether the global values shown in the wat file are the same numbers we are getting in Java.

You can take the runtime wasm and wat from [here](https://github.com/LimeChain/wasmer-java/tree/dev/examples) as it is the smallest runtime we have, but before you do that you can see if the `hello_world.wasm` also exports `__heap_base` (I think all wasm files do) and if it is a value greater than 0 you can use it to do the above mentioned test as it would be way more lightweight and would require no function imports as opposed to the runtime one. You would need to use `wasm2wat` or a similar program to generate the wat file for `hello_world.wasm`

src/main/java/com/limechain/runtime/hostapi/HostApi.java
```diff
@@ -50,69 +53,77 @@ protected static ImportObject getImportObject(final String functionName,
...
+    @Deprecated(forRemoval = true)
     public static byte[] getDataFromMemory(long pointer) {
-        int ptr = (int) pointer;
-        int ptrLength = (int) (pointer >> 32);
-        byte[] data = new byte[ptrLength];
-
-        Memory memory = getMemory();
-        memory.buffer().get(ptr, data, 0, ptrLength);
-        return data;
+        return null;
```
ablax: Maybe until removal provide a fallback implementation to the deprecated method
```suggestion
        return getDataFromMemory(new RuntimePointerSize(pointer));
```

src/main/java/com/limechain/runtime/hostapi/HostApi.java
```diff
@@ -50,69 +53,77 @@ protected static ImportObject getImportObject(final String functionName,
...
-        return runtime.getInstance().exports.getMemory("memory");
-    }
-
-    public static void setRuntime(Runtime runtime) {
-        HostApi.runtime = runtime;
-    }
-
-    public static void setRepository(KVRepository<String, Object> repository) {
-        HostApi.repository = repository;
+        return -1;
```
ablax: Maybe until removal provide a fallback implementation to the deprecated method

```suggestion
        return writeDataToMemory(data).pointer();
```

src/main/java/com/limechain/runtime/hostapi/HostApi.java
```diff
@@ -1,34 +1,37 @@
...
  * HostApi functions implementations
  */
 @Log
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
 public class HostApi {
-
     protected static final List<Number> EMPTY_LIST_OF_NUMBER = List.of();
     protected static final List<Type> EMPTY_LIST_OF_TYPES = List.of();
-
     protected static final String KEY_TO_IGNORE = ":child_storage:default:";
```
ablax: This line is obsolete
```suggestion
```

src/main/java/com/limechain/runtime/allocator/AllocationStats.java
```diff
@@ -0,0 +1,24 @@
...
+
+import lombok.Data;
+
+import java.math.BigInteger;
+
+@Data
+public class AllocationStats {
+    private int bytesAllocated;
+    private int bytesAllocatedPeak;
+    private BigInteger bytesAllocatedSum;
```
ablax: allocated:15 throws nullpointer exception
```suggestion
    private BigInteger bytesAllocatedSum = BigInteger.ZERO;
```

ablax (approved):
lgtm

PR #203, opened 2023-10-17
Warp sync from bootnodes

bokoto000:
Requests fragments from bootnodes instead of connected peers.


ablax (approved):

src/main/java/com/limechain/sync/warpsync/state/RequestFragmentsState.java
```diff
@@ -47,15 +45,26 @@ public void next(WarpSyncMachine sync) {
 
     @Override
     public void handle(WarpSyncMachine sync) {
+        WarpSyncResponse resp = null;
+        for (int i = 0; i < sync.getNetworkService().kademliaService.getBootNodePeerIds().size(); i++) {
+            try {
+                Thread.sleep(1000);
```
vikinatora: Why are we sleeping here?

bokoto000: Removed

src/main/java/com/limechain/sync/warpsync/state/RequestFragmentsState.java
```diff
@@ -47,15 +45,25 @@ public void next(WarpSyncMachine sync) {
...
+        for (int i = 0; i < sync.getNetworkService().kademliaService.getBootNodePeerIds().size(); i++) {
+            try {
+                resp = sync.getNetworkService().makeWarpSyncRequest(blockHash.toString());
+                break;
+            } catch (Exception e) {
+                if (!sync.getNetworkService().updateCurrentSelectedPeerWithBootnode(i)) {
+                    this.error = new RuntimeException("Could not update current selected peer");
+                    return;
+                }
+            }
```
georg-getz: It's always good to print the initial exception (in this case `e`) because sometimes it's different than what we expect.

georg-getz (approved):

src/main/java/com/limechain/lightclient/LightClient.java
```diff
@@ -40,8 +40,8 @@ public void start() {
         this.network.start();
 
         while (true) {
-            if (!connectionManager.getPeerIds().isEmpty()) {
-                if (this.network.currentSelectedPeer != null) {
+            if (network.kademliaService.getBootNodePeerIds().size()>0) {
```
georg-getz: you could add spaces before and after `>`

PR #231, opened 2023-10-31
Implement some of Offchain API

stoyan-lime:


src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
```diff
@@ -13,29 +29,39 @@
...
+@AllArgsConstructor
 public class OffchainHostFunctions {
+    private final HostApi hostApi;
+    private final HostConfig config;
 
+    public OffchainHostFunctions() {
+        hostApi = HostApi.getInstance();
+        config = AppBean.getBean(HostConfig.class);
+    }
     public static List<ImportObject> getFunctions() {
```
georg-getz: nit: add an empty line before `getFunctions()`

src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
```diff
@@ -75,4 +101,109 @@ public static List<ImportObject> getFunctions() {
...
+            multiAddresses.stream().map(Multiaddr::serialize).forEach(data::put);
+
+            Result<byte[], Exception> result = new Result<>(Result.ResultMode.OK, data.array(), null);
+
+            new ResultWriter<byte[], Exception>()
+                    .writeResult(writer, ScaleCodecWriter::writeByteArray, null, result);
+            return buf.toByteArray();
+        } catch (IOException e) {
+            return scaleEncodedEmptyResult(false);
+        }
```
georg-getz: We always want to have access to the exception stack trace without having to debug the application. I will leave it up to you to determine the correct logging level, judging by the fact that we are not throwing an exception `error` would probably be too high of a level, but it's hard for me to determine whether `warning` is just right or too high.

stoyan-lime: Added a `warning` log

georg-getz (approved):

PR #234, opened 2023-11-13
Implement Offchain HostAPI http request methods

stoyan-lime:


src/test/java/com/limechain/runtime/hostapi/OffchainHostFunctionsTest.java
```diff
@@ -76,4 +106,130 @@ void extOffchainRandomSeedReturnsAPointerToA32BitNumber() {
...
+            byte[] scaleEncodedHeaders = offchainHostFunctions.scaleEncodeHeaders(responseHeaders);
+            when(hostApi.writeDataToMemory(scaleEncodedHeaders)).thenReturn(resultPointer);
+
+            RuntimePointerSize result = offchainHostFunctions.extOffchainHttpResponseHeadersVersion1(requestId);
+            assertEquals(resultPointer, result);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
```
bokoto000, line 374: Tests were commented out because by using the response wait method we can't mock Instant.now adequately.

src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
```diff
@@ -213,4 +239,245 @@ public int extOffchainRandomSeed() {
...
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        ScaleCodecWriter writer = new ScaleCodecWriter(outputStream);
+        for (String header : headers.keySet()) {
+            List<String> value = headers.get(header);
+            for (String headerValue : value) {
+                writer.writeByteArray(headerValue.getBytes());
+            }
+            pairs.add(new Pair<>(header, outputStream.toByteArray()));
+        }
+        return ScaleUtils.Encode.encode(pairs, String::getBytes, Function.identity());
```
David-Petrov: What I strongly suspect is, those multi-values of a single header key might need to be concatenated with an `" "` or `", "` as per HTTP specs, but I'm not entirely sure. Just raising this as a consideration.

In any case, a suggested nitpick to shorten the implementation (we can concat in a string, and then encode the string, I think we needn't the entire value-by-value encoding and then concatenation).
```suggestion
        List<Pair<String, String>> pairs = new ArrayList<>(headers.size());
        for (String header : headers.keySet()) {
            // NOTE: Not sure how the Spec expects us to handle repeated AND multi-valued headers, so we just concat the values (a little sus?)
            String value = String.join("", headers.get(header)); 
            pairs.add(new Pair<>(header, value));
        }
        return ScaleUtils.Encode.encode(pairs, String::getBytes, String::getBytes);
```

src/test/java/com/limechain/runtime/hostapi/OffchainHostFunctionsTest.java
David-Petrov: This particular test `extOffchainHttpResponseBodyVersion1ReturnsDeadlineReachedTest` seems to fail because we're mocking `requests.waitRequest` but it's not actually used in the logic behind (seems to only be used in the test's mock?). What fails is that (separately) `requests.getRequestsResponses` returns `null` as the mock doesn't have a provided behaviour for that invocation.

I'm leaving this at that for now, since I'm not entirely sure what the intention of these tests is and how significant they are. :D

bokoto000: Tests were changed and fixed

src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
```diff
@@ -80,24 +93,41 @@ public List<ImportObject> buildFunctions() {
...
-                }, List.of(Type.I64, Type.I64), Type.I64),
-                HostApi.getImportObject("ext_offchain_http_response_headers_version_1", argv -> {
-                    return 0;
-                }, List.of(Type.I32), Type.I64),
-                HostApi.getImportObject("ext_offchain_http_response_read_body_version_1", argv -> {
-                    return 0;
-                }, List.of(Type.I32, Type.I64, Type.I64), Type.I64),
+                HostApi.getImportObject("ext_offchain_http_request_start_version_1", argv ->
+                                extOffchainHttpRequestStart(new RuntimePointerSize(argv.get(0)),
+                                        new RuntimePointerSize(argv.get(1)), new byte[0]).pointerSize(),
```
vikinatora: Format it like the other ones.

src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
```diff
@@ -80,24 +93,41 @@ public List<ImportObject> buildFunctions() {
...
+                        List.of(Type.I64, Type.I64, Type.I64), Type.I64),
+                HostApi.getImportObject("ext_offchain_http_request_add_header_version_1", argv ->
+                                extOffchainHttpRequestAddHeader(
+                                        argv.get(0).intValue(),
+                                        new RuntimePointerSize(argv.get(1)),
+                                        new RuntimePointerSize(argv.get(2))).pointerSize(),
```
vikinatora: Format like the other functions

src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
```diff
@@ -214,4 +238,249 @@ public int extOffchainRandomSeed() {
...
+                .map(Long::intValue)
+                .orElse(0);
+    }
+
+    int[] decodeRequestIdArray(byte[] encodedRequestIds) {
+        ScaleCodecReader reader = new ScaleCodecReader(encodedRequestIds);
+        int length = reader.readCompactInt();
+        int[] requestIds = new int[length];
+        for (int i = 0; i < length; i++) {
+            requestIds[i] = (int) reader.readUint32();
```
georg-getz: Is it safe to cast u32 to int here? The upper half of u32 values overflow int

src/main/java/com/limechain/runtime/hostapi/dto/HttpResponseType.java
```diff
@@ -0,0 +1,47 @@
...
+    public int getValue() {
+        return value;
+    }
+
+    public byte[] scaleEncodedResult() {
+        try (
+                ByteArrayOutputStream buf = new ByteArrayOutputStream();
+                ScaleCodecWriter writer = new ScaleCodecWriter(buf);
+        ) {
+            if (value == 3) {
```
georg-getz: 3 is a magic number here.

src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
bokoto000: Overflowing would make the integer negative but since we store the ids in a Map it would still find the corresponding request

bokoto000: This [struct](https://github.com/paritytech/polkadot-sdk/blob/9f5221cc2f7f10adfd22b293ceed060617468936/substrate/primitives/core/src/offchain/mod.rs#L101C1-L101C1) is used for containing the Ids of the requests and is of type i16 so we won't get to a point where we overflow.

src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
```diff
@@ -214,4 +242,253 @@ public int extOffchainRandomSeed() {
...
+             ScaleCodecWriter writer = new ScaleCodecWriter(buf)) {
+            int createdId = requests.addRequest(method, uri);
+
+            Result<Integer, Exception> result = new Result<>(Result.ResultMode.OK, createdId, null);
+            new ResultWriter<Integer, Exception>()
+                    .writeResult(writer, ScaleCodecWriter::writeUint16, null, result);
+
+            return hostApi.writeDataToMemory(buf.toByteArray());
+        } catch (IOException e) {
+            log.log(Level.SEVERE, e.getMessage(), e.getStackTrace());
```
georg-getz: Logging level seems wrong. Other IOExceptions are handled as warnings.

src/main/java/com/limechain/runtime/hostapi/dto/InvalidRequestId.java
```diff
@@ -0,0 +1,7 @@
+package com.limechain.runtime.hostapi.dto;
+
+public class InvalidRequestId extends RuntimeException{
```
georg-getz: space after RuntimeException

georg-getz (approved):

src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
```diff
@@ -5,31 +5,43 @@
...
 import io.emeraldpay.polkaj.scale.ScaleCodecReader;
 import io.emeraldpay.polkaj.scale.ScaleCodecWriter;
+import io.emeraldpay.polkaj.scale.reader.ListReader;
+import io.emeraldpay.polkaj.scale.reader.UInt64Reader;
+import io.emeraldpay.polkaj.scale.writer.ListWriter;
 import io.emeraldpay.polkaj.scaletypes.Result;
 import io.emeraldpay.polkaj.scaletypes.ResultWriter;
 import io.libp2p.core.PeerId;
 import io.libp2p.core.multiformats.Multiaddr;
+import kotlin.Pair;
```
David-Petrov, line 28: Since `kotlin` is a transitive dependency, we had earlier agreed to use `org.javatuples.Pair` instead. Just for consistency.

David-Petrov (approved):

src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
stoyan-lime: The kotlin pair is used in `ScaleUtils.Encode.encode`, which we call in this class.

David-Petrov: Okay, sorry, disregard. Will be refactored later.

PR #246, opened 2023-11-21
Introduce FullClient

David-Petrov:
- refactor to separate the RpcApp from LightCLient
- extract bean stopping concerns to RpcApp::stop()
- introduce a FullClient alternative to the LightClient
- add logic in Main to switch client roles depending on the -node-mode CLI argument
- also, disable failing tests in SyncTest; there seems to be an issue with the chosen bootnode


src/main/java/com/limechain/Main.java
```diff
@@ -1,15 +1,59 @@
...
+import lombok.extern.java.Log;
 import sun.misc.Signal;
 
+import java.util.logging.Level;
+
+@Log
 public class Main {
+
     public static void main(String[] args) {
+        // Instantiate and start the spring application, so we get teh global context
```
vikinatora:
```suggestion
        // Instantiate and start the spring application, so we get the global context
```

vikinatora (approved):
LGTM, a lot of TODOs left but it's fine imo.

src/main/java/com/limechain/lightclient/FullClient.java
```diff
@@ -0,0 +1,41 @@
...
+import com.limechain.network.Network;
+import com.limechain.rpc.server.AppBean;
+import com.limechain.sync.warpsync.WarpSyncMachine;
+import lombok.SneakyThrows;
+import lombok.extern.java.Log;
+
+import java.util.logging.Level;
+
+@Log
+public class FullClient implements NodeClient {
```
vikinatora: FullNode is the term Polkadot uses.

src/main/java/com/limechain/lightclient/NodeClient.java
```diff
@@ -0,0 +1,13 @@
+package com.limechain.lightclient;
+
+// TODO: Think of a better name... that's tautological, simply `Client` is awfully generic
+public interface NodeClient {
```
vikinatora:
```suggestion
// TODO: Think of a better name... that's tautological, simply `Client` is awfully generic
public interface HostNode {
```

georg-getz (approved):

src/main/java/com/limechain/lightclient/FullClient.java
```diff
@@ -0,0 +1,41 @@
+package com.limechain.lightclient;
```
David-Petrov: What about renaming the package from `lightclient` to something more fitting, like just `client`, or `node` or something... I'm not sure? @vikinatora

David-Petrov: @vikinatora

vikinatora: Makes sense 👍

David-Petrov: We'll go with `client`.

PR #247, opened 2023-11-22
Transaction queue

### LINKED ISSUE #495: P2P communication: Set up Kademlia DHT running in the web

### LINKED ISSUE #576: EPIC: Transaction validation
In order to have a working BABE algorithm Fruzhin needs to change some things that are blocking it from validating transactions.

1. [x] #574 
2. [x] #573
3. [x] #575

ablax:
- [x] Create transaction pool
- [x] Create transaction queue
- [x] Prepare for transaction validation


src/main/java/com/limechain/network/protocol/transaction/TransactionsEngine.java
```diff
@@ -126,7 +142,9 @@ public void writeHandshakeToStream(Stream stream, PeerId peerId) {
     public void writeTransactionsMessage(Stream stream, PeerId peerId) {
         ByteArrayOutputStream buf = new ByteArrayOutputStream();
         try (ScaleCodecWriter writer = new ScaleCodecWriter(buf)) {
-            writer.write(new TransactionsWriter(), new byte[][]{new byte[]{}, new byte[]{}});
+            writer.write(new TransactionsWriter(), new Extrinsics[]{
+                    new Extrinsics(new byte[]{}), new Extrinsics(new byte[]{})
```
bokoto000: Add TODO to replace empty transaction messages in future

src/main/java/com/limechain/network/protocol/transaction/state/TransactionState.java
```diff
@@ -0,0 +1,103 @@
...
+        return null;
+    }
+
+    @NotNull
+    private Future<ValidTransaction> popFutureTransaction() {
+        return executor.submit(() -> {
+            ValidTransaction transaction = null;
+            while (transaction == null) {
+                transaction = popTransaction();
+                Thread.sleep(50);
```
bokoto000: Why do we use sleep here

src/main/java/com/limechain/network/protocol/transaction/state/TransactionState.java
```diff
@@ -0,0 +1,103 @@
...
+
+    public void removeExtrinsicFromPool(byte[] extrinsic) {
+        transactionPool.removeExtrinsic(extrinsic);
+    }
+
+    public byte[] addToPool(ValidTransaction validTransaction) {
+        return transactionPool.insert(validTransaction);
+    }
+
+    //public void notifyStatus
```
bokoto000: Add as TODO

src/main/java/com/limechain/network/protocol/transaction/state/ValidTransaction.java
```diff
@@ -0,0 +1,37 @@
...
+    public ValidTransaction(byte[] extrinsic, Validity validity){
+        this.extrinsic = extrinsic;
+        this.validity = validity;
+    }
+
+    public int compareTo(@NotNull ValidTransaction transaction) {
+        return new ValidTransactionComparator().compare(this, transaction);
+    }
+
+    static class ValidTransactionComparator implements Comparator<ValidTransaction> {
```
bokoto000: Maybe rename this to ValidTransactionPriorityComparator as I'm not sure if its clear that we are comparing by priority

src/main/java/com/limechain/network/protocol/transaction/state/TransactionState.java
ablax: To avoid high CPU usage in this almost infinite loop

src/main/java/com/limechain/network/protocol/transaction/state/Pool.java
```diff
@@ -0,0 +1,37 @@
+package com.limechain.network.protocol.transaction.state;
+
+import com.limechain.utils.HashUtils;
+import lombok.NoArgsConstructor;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@NoArgsConstructor
```
georg-getz: No arg constructor is implied if no other constructor is found.

src/main/java/com/limechain/network/protocol/transaction/state/ValidTransaction.java
```diff
@@ -0,0 +1,29 @@
...
+    private Validity validity;
+
+    public ValidTransaction(byte[] extrinsic){
+        this.extrinsic = extrinsic;
+    }
+
+    public ValidTransaction(byte[] extrinsic, Validity validity){
+        this.extrinsic = extrinsic;
+        this.validity = validity;
+    }
```
georg-getz: space before opening curly brace `{`

src/main/java/com/limechain/network/protocol/transaction/state/Validity.java
```diff
@@ -0,0 +1,24 @@
...
+    private BigInteger priority;
+    private byte[][] requires;
+    private byte[][] provides;
+    private BigInteger longevity;
+    private boolean propagate;
+
+    public Validity(BigInteger priority){
+        this.priority = priority;
+    }
+}
```
georg-getz: Add a comment explaining what this class represents and how the priority works. If I have priority 1 and someone else has 18, am I with a higher priority or not? Also how are the negative cases handled? Would be nice to have a test for that case as well.

ablax:
Currently blocked, as it depends on Trie

Zurcusa:
This PR can be merged with its current state. The logic for sending a validation request to the runtime and decoding the response is there. However, it seems we have some kind of an issue that causes the runtime to respond with an invalid transaction validation error for every transaction. 

We have:
- Logic for validating incoming transaction from the transactions protocol.
- Logic for appropriately putting transactions in queue or pool.

What is still needed:
- Fix the issue that causes validation error. (could be related to the fact that we don't execute blocks after full sync, or as Kamil from Kagome mentioned it could be because the validation runtime request is trying to alter the state)
- Send known non repetitive transactions to peers in a transaction message.

src/main/java/com/limechain/transaction/TransactionState.java
```diff
@@ -0,0 +1,119 @@
...
+        transactionPool = new TransactionPool();
+        executor = Executors.newSingleThreadExecutor();
+        transactionQueue = new PriorityQueue<>();
+    }
+
+    public void pushTransaction(ValidTransaction validTransaction) {
+        transactionQueue.add(validTransaction);
+    }
+
+    public ValidTransaction popTransaction() {
```
Grigorov-Georgi: Not a big deal, but since 'pop' semantic suggests a stack-related operation, it might be better to rename it to pollTransaction for clarity.

Grigorov-Georgi (approved):

src/main/java/com/limechain/transaction/TransactionPool.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public class TransactionPool {
+
+    private final Map<byte[], ValidTransaction> transactions;
+
+    public TransactionPool() {
+        transactions = new HashMap<>();
+    }
```
georg-getz, line 16: You can move the initialization of the map to the field declaration and delete the constructor entirely

src/main/java/com/limechain/transaction/TransactionState.java
```diff
@@ -0,0 +1,119 @@
...
+        try {
+            return futureTransaction.get(timeout, java.util.concurrent.TimeUnit.MILLISECONDS);
+        } catch (ExecutionException e) {
+            log.severe("Error while waiting for transaction: " + e.getMessage());
+        } catch (TimeoutException e) {
+            futureTransaction.cancel(true);
+        } catch (InterruptedException e) {
+            if (Thread.interrupted())
+                throw new InterruptedException();
+        }
```
georg-getz: InterruptedException is a checked excaption and we don't want to throw it, we have ThreadInterruptedException for cases like this one

src/test/java/com/limechain/network/protocol/transaction/state/TransactionStateTest.java
```diff
@@ -0,0 +1,126 @@
...
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class TransactionStateTest {
+
+    private static final byte[] TAG_1 = "tag1".getBytes();
+    private static final byte[] TAG_2 = "tag2".getBytes();
+
+    private TransactionState sut;
```
georg-getz, line 24: can you rename this field

src/main/java/com/limechain/transaction/TransactionPool.java
Zurcusa: I usually prefer keeping init code inside of the ctors, especially in spring projects. Since in spring ctor bean injection is preferred it helps keep a uniform init standard in the whole project.

src/test/java/com/limechain/network/protocol/transaction/state/TransactionStateTest.java
Zurcusa: I always name the class I test following this approach https://en.wikipedia.org/wiki/System_under_test

georg-getz (approved):

src/main/java/com/limechain/transaction/TransactionState.java
```diff
@@ -0,0 +1,120 @@
...
+
+        try {
+            return futureTransaction.get(timeout, java.util.concurrent.TimeUnit.MILLISECONDS);
+        } catch (ExecutionException e) {
+            log.severe("Error while waiting for transaction: " + e.getMessage());
+        } catch (TimeoutException e) {
+            futureTransaction.cancel(true);
+        } catch (InterruptedException e) {
+            if (Thread.interrupted())
+                throw new ThreadInterruptedException(e);
```
georg-getz: If you are catching an `InterruptedException` then you don't need to check whether the thread was interrupted but rather force it to be interrupted like this `Thread.currentThread().interrupt();`
This will clear the sonar cloud error

PR #248, opened 2023-11-23
Extract some reused SCALE en/decoding logic into `utils`

David-Petrov:
A beginning of a vision to extract away all the SCALE encoding/decoding logic into `utils`.

### DIFF
--- a/src/main/java/com/limechain/utils/scale/ScaleUtils.java
+++ b/src/main/java/com/limechain/utils/scale/ScaleUtils.java
@@ -1,0 +1,71 @@
+package com.limechain.utils.scale;
+
+import com.limechain.utils.scale.writers.PairWriter;
+import io.emeraldpay.polkaj.scale.ScaleCodecReader;
+import io.emeraldpay.polkaj.scale.ScaleCodecWriter;
+import io.emeraldpay.polkaj.scale.ScaleReader;
+import io.emeraldpay.polkaj.scale.reader.ListReader;
+import io.emeraldpay.polkaj.scale.writer.ListWriter;
+import kotlin.Pair;
+import lombok.experimental.UtilityClass;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.Function;
+
+// TODO: This class is currently just a helper utility class,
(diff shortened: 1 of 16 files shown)


bokoto000 (approved):
lgtm

src/main/java/com/limechain/utils/scale/ScaleUtils.java
```diff
@@ -0,0 +1,71 @@
...
+@UtilityClass
+public class ScaleUtils {
+
+    @UtilityClass
+    public class Decode {
+        public <T> T decode(byte[] encodedData, ScaleReader<T> reader) {
+            try {
+                return new ScaleCodecReader(encodedData).read(reader);
+            } catch (RuntimeException e) {
+                throw new RuntimeException("Error while SCALE decoding.", e); // TODO: is this a code smell?
```
georg-getz: Yes, we'll talk about this in tomorrow's daily but we won't throwing `RuntimeException` anymore. Extend it with a custom more specific one and throw that one instead.

src/main/java/com/limechain/utils/scale/ScaleUtils.java
```diff
@@ -0,0 +1,71 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.Function;
+
+// TODO: This class is currently just a helper utility class,
+//  planned to grow into a unified scale encode/decode util class with whatever methods are useful
+//  WIP
+// Currently trying out different approaches to spare some of the boilerplate around SCALE en/decoding
```
georg-getz: Indentation is a big weird, why not have it all start at the left. You don't need a comma after class and you delete the first "class" and the "just" strings.

vikinatora (approved):

src/main/java/com/limechain/utils/scale/ScaleUtils.java
David-Petrov: Done. 

Also, while writing the custom ScaleDecodingException, I noticed there was an existing ScaleEncodingException in `network.protocol.warp.exception`, moved it into `utils.scale.exceptions`.
If nobody has anything against this, we can resolve this conversation and merge the PR.

georg-getz (approved):

PR #255, opened 2023-12-01
Transform trie_structure to database node operations

### LINKED ISSUE #243: Transform `trie_structure` to database node operations
https://github.com/smol-dot/smoldot/blob/377fdf027862b3d39f65b6883bea0edb96208f7b/full-node/src/lib.rs#L952-L988
- [ ] Implement algorithm and structs required

ablax:
- [x] Implement algorithm and structs required

Closes #243

### DIFF
--- a/src/main/java/com/limechain/client/FullNode.java
+++ b/src/main/java/com/limechain/client/FullNode.java
@@ -6,11 +6,17 @@ import com.limechain.storage.KVRepository;
 import com.limechain.storage.block.BlockStateHelper;
 import com.limechain.sync.warpsync.WarpSyncMachine;
 import com.limechain.trie.structure.TrieStructure;
+import com.limechain.trie.structure.nibble.Nibble;
+import com.limechain.trie.structure.node.InsertTrieNode;
+import com.limechain.trie.structure.node.TrieNodeIndex;
+import com.limechain.trie.structure.node.handle.NodeHandle;
 import lombok.SneakyThrows;
 import lombok.extern.java.Log;
 import org.javatuples.Pair;
 
 import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Optional;
 import java.util.logging.Level;
 
@@ -33,7 +39,8 @@ public class FullNode implements HostNode {
             // do nothing?
         } else {
             // do the initial one-time population of the database with the genesis storage
-            initializeDatabaseStorageTrie();
+            TrieStructure<Pair<Optional<byte[]>, byte[]>> trieStructure = initializeDatabaseStorageTrie();
+            buildTrieNodesIterator(trieStructure);
         }
 
         // Start network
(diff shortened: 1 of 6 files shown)


src/main/java/com/limechain/client/FullNode.java
```diff
@@ -58,8 +65,62 @@ public void start() {
...
+     * @param trieStructure The trie structure containing the nodes. This structure should be
+     *                      a {@link TrieStructure} with user data of type {@link Pair<Optional<byte[]>, byte[]>},
+     *                      where the first element of the pair is an optional storage value (byte array),
+     *                      and the second element is the merkle value (byte array).
+     * @return A list of {@link InsertTrieNode} objects representing the nodes in the given trie structure.
+     * @throws IllegalStateException if the user data in the trie structure is empty or null, which
+     *                               indicates an invalid state for the trie nodes.
+     */
+    public List<InsertTrieNode> buildTrieNodesIterator(
+            TrieStructure<Pair<Optional<byte[]>, byte[]>> trieStructure) {
```
David-Petrov: As far as I gather, we're mixing an `Optional<byte[]>` for the storage value and then a pure `byte[]` for the merkle value, which we check for `null` to determine whether the field is present or not... Any particular reason for that decision? 😄

src/main/java/com/limechain/trie/structure/node/InsertTrieNode.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Represents a node in a trie structure for insertion into a database.
+ * This class encapsulates the storage value, merkle value, children's merkle values,
+ * and partial key nibbles of a trie node.
+ */
+@AllArgsConstructor
+@Getter
+
+public class InsertTrieNode {
```
David-Petrov, line 17: [Possibly a nitpick] Maybe this could become a record if it's a mere data holder?

src/main/java/com/limechain/trie/structure/node/handle/NodeHandle.java
```diff
@@ -25,4 +26,8 @@ public List<Nibble> getPartialKey() {
         //TODO: implement
         throw new NotImplementedException("Not yet implemented.");
     }
+
+    public T getUserData() {
+        return trieStructure.nodeAtIndex(new TrieNodeIndex(nodeIndex)).getUserData();
```
David-Petrov: That seems to be an infinite recursion... we'll have to refactor the access modifiers a bit, so that a `NodeHandle` has direct access to the `TrieStructure`'s underlying data structure, containing the actual structural nodes.

David-Petrov: For now we can do something like
```java
    return trieStructure.getUserDataAtIndex(new TrieNodeIndex(nodeIndex));
```
and perhaps we could also store the actual integer wrapped in a `NodeIndex` within the `NodeHandle`... or maybe abolish the notion of a `NodeIndex` if it becomes too messy.

David-Petrov (approved):
LGTM.

georg-getz (approved):

PR #261, opened 2023-12-13
Implement slab

bokoto000:


georg-getz (approved):

src/main/java/com/limechain/trie/structure/slab/Slab.java
```diff
@@ -16,70 +15,157 @@
...
+        }
+        size++;
         return index;
     }
 
+    /**
+     * Removes the element at the specified index from the slab.
+     *
+     * @param index the index of the element to be removed
+     * @return the element previously at the specified index
```
David-Petrov, line 83: We should consider whether we want to throw if queried for an invalid index or always return null.
Because right now:
- we return null if the index has been used once, but freed later;
- we throw an `IndexOutOfBoundsException` if the index is out of bounds for the underlying `storage`.

I'd say the stance of this data structure is rather: "I provide you with valid indices, if you try to use me with an index (int) not explicitly guaranteed to be valid by me, it's an end-of-world scenario and I'll throw", but we might also go the other way and simply return null on invalid indices (thus users can't insert `null` as a value, which is already set in the contract of `add()`). 

Although this contradicts the contract for `get()`... 😄

src/main/java/com/limechain/trie/structure/slab/Slab.java
```diff
@@ -16,70 +15,157 @@
...
+ * @param <T> the type of elements held in this slab
+ */
 public class Slab<T> implements Iterable<Pair<Integer, T>> {
     private ArrayList<T> storage;
     private int newIndex;
     private Queue<Integer> freeIndices;
+    private int size;
 
+    /**
+     * Constructs an empty Slab with default initial capacity of 0.
```
David-Petrov: Maybe the default capacity should be sth like `10` or `13`, similar to that Java's devs have decided for `ArrayList`?

src/main/java/com/limechain/trie/structure/slab/Slab.java
David-Petrov: Now that I think about it, better return null if the index is invalid. Stick to the notion of "null means no value for the given index, i.e. invalid index".

src/main/java/com/limechain/trie/structure/slab/Slab.java
bokoto000: I've set it to 10

src/main/java/com/limechain/trie/structure/slab/Slab.java
```diff
@@ -94,13 +95,23 @@ public T remove(int index) {
...
      *
-     * @param index the index of the element to return
-     * @return the element at the specified index, or null if no element is present
+     * @param index The index of the element to be retrieved.
+     * @return The element at the specified index.
+     * @throws InvalidSlabIndexException If the index is out of bounds or if the value at the index is null.
+     *                                   This exception wraps the original {@code IndexOutOfBoundsException}
+     *                                   when the index is out of bounds.
      */
     public @Nullable T get(int index) {
```
David-Petrov: This must become `@NotNull T`

David-Petrov (approved):

PR #265, opened 2023-12-22
Implement Trie Structure

David-Petrov:
Quick summary:
- Nibble;
- Nibbles: a convenience representation for any "sequence of nibbles"-like data structure source;
- a small change in `build.gradle.kts` to enable preview features (we want the pattern matching in switch expressions/statements);
- the essential TrieStructure, including TrieNode, the various NodeHandles, etc.;
- a debugging helper, the dot serializer for the trie;
- tests for the TrieStructure.


src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
```diff
@@ -0,0 +1,281 @@
...
+//  but it's what I've reached as a most convenient option to use outside.
+//  Basically a List<Nibble> with additional methods.
+
+// TODO:
+//  As a List<Nibble> implementor, Nibbles is currently mutable, which I severely dislike,
+//  since this mandates the developer to think about where to copy and where passing by reference is sufficient.
+/**
+ * Convenience wrapper for any 'sequence of Nibble'-like structure
+ */
+public class Nibbles implements List<Nibble>, RandomAccess, Comparable<List<Nibble>> {
```
David-Petrov: This Nibbles class is the only remaining "big mess", since I've not yet come to a final decision for its design. Refer to the todos for the two main dilemmas, but in quick summary:
- do we want to `implements List<Nibble>`, `extends List<Nibble>` or none at all, perhaps restricting ourselves to a more generic interface like `implements Collection<Nibble>` or `implements Iterable<Nibble>` (initially, it was simply the latter, but a lot of boilerplate piled up outside); 
- do we want this to be a mutable or immutable structure? Immutability brings about a lot of benefits like eliminating the issue of ownership and the fear of outside mutations while handling a given sequence of nibbles in some part of the project. As a tradeoff, we lose the java convenience of obeying a generic interface like `List` or `Collection`, which immensely reduces boilerplate and makes the class really blissful to use... but those interfaces mandate mutability.

src/main/java/com/limechain/trie/structure/nibble/NibblesCollector.java
```diff
@@ -0,0 +1,46 @@
...
+        return (list1, list2) -> {
+            list1.addAll(list2);
+            return list1;
+        };
+    }
+
+    @Override
+    public Function<List<Nibble>, Nibbles> finisher() {
+        // TODO: Avoid this unnecessary copying by introducing "lightweight by-reference only" constructors to Nibbles
+        return Nibbles::new;
```
David-Petrov, line 37: Again, on the topic of mutability. The current state for now is: everytime we construct a nibble, it clones the data structure passed in order to assure ownership of the nibbles. This might cause inefficiencies in places where copying isn't necessary, like when we're constructing the outer collection of nibbles for the sole purpose of wrapping it within a `Nibbles` instance (like it happens here)... remains to be decided after we decide what we want `Nibbles` to be.

src/main/java/com/limechain/trie/structure/BranchNodeHandle.java
```diff
@@ -0,0 +1,41 @@
...
+     * This doesn't change the trie structure.
+     * @return the newly created {@link StorageNodeHandle}
+     * @implNote After calling this method, {@code this} instance is no longer valid,
+     *           since the underlying node has been converted to a storage node,
+     *           thus it's no longer a branch node.
+     *           Any further usage of this {@link BranchNodeHandle} would be invalid.
+     * @throws IllegalStateException if this branch node handle has already been consumed,
+     *                               i.e. the referenced node has been converted to a storage node.
+     */
+    public StorageNodeHandle<T> convertToStorageNode() {
```
David-Petrov, line 32: That's a direct issue when designing Java code influenced by Rust's tooling, since in Java we can't instruct the JVM to "consume" `this` and disallow any further usage of the branch node handle. 

Consider whether that's a compromise we're willing to allow (refer to the javadoc). :)

src/main/java/com/limechain/trie/structure/NodeHandle.java
```diff
@@ -0,0 +1,115 @@
...
+    protected final int rawNodeIndex;
+
+    protected NodeHandle(TrieStructure<T> trieStructure, int rawNodeIndex) {
+        super(trieStructure);
+        this.rawNodeIndex = rawNodeIndex;
+    }
+
+    //NOTE:
+    // Later on, we might switch from optionals to nullable return types
+    // Depends on whether Optionals will be more helpful or more annoying
```
David-Petrov, line 31: This note applies also to TrieStructure's public interface. In general, currently we've got a mixture of both approaches: Optionals and nullable return types.

The original idea was:
- in places where "returning nothing" is a more valid and common use case, I've gone with Optional
- methods which are expected to be called when the user site is informed what needs to happen (i.e. a non-null value is expected to exist, missing values are rather an exception), I've gone for the @Nullable return type. A bit annoying is that this annotation enforces null-checks on user site, but I find that rather useful (we want to communicate to the user, that we know, and it's intentional, that his value could be missing).

But what we ended up with is rather a mess, so I'm thinking of unifying the interface in favor of one of both alternatives.

P.S. For easier public interface inspection, command + fn + f12 in IntelliJ.

src/main/java/com/limechain/trie/structure/TrieNodeIndex.java
```diff
@@ -0,0 +1,24 @@
+package com.limechain.trie.structure;
+
+import lombok.AccessLevel;
+import lombok.Data;
+import lombok.Setter;
+
+//TODO:
+// Consider making this data class an inner class of TrieStructure to guarantee
+// that the index has been generated by the same TrieStructure instance (with a runtime check, ofc)
```
David-Petrov: This idea with the "validity by construction" fell apart, since the `TrieNodeIndex` could always have been generated by one instance of the TrieStructure, but then used in another. But that's a very intentional invalid usage, so we're simply allowing the unexpected behaviour / exceptions if this happens. I think that's ok?

P.S. Also, methods in TrieStructure accepting `TrieNodeIndex` as an index are returning nullable stuff, since that's an example of "rather commonly expected value", so bothering with optionals would be too much... I think?

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -1,30 +1,480 @@
...
+                }
+
+                yield new Vacant<>(this, key, closestAncestorIndex);
+            }
+        };
+    }
+    // NOTE:
+    //  Those (six) iteration methods could be reduced, or at least live outside this class,
+    //  but I find them all reasonably useful and couldn't decide for anything particularly cleaner.
+
```
David-Petrov: This note is rather significant: I'm open for suggestions. I still think this abundance of iteration methods is really useful, but it clutters the public interface with repetitive methods.

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -1,30 +1,480 @@
...
+            for (Nibble nibble : current.partialKey) {
+                if (!keyIter.hasNext() || !keyIter.next().equals(nibble)) {
+                    return new ExistingNodeInnerResult.NotFound(closestAncestor);
+                }
+            }
+
+            // At this point, the tree traversal cursor (the already consumed part of the `key` iterator)
+            // exactly matches `current`.
+
+            // TODO: clone the `key` iterator more sanely... we're currently copying the whole underlying data source
```
David-Petrov: To my surprise, I haven't found a more elegant solution to this problem. Really, what we need is simply to copy the iterator itself, but Java doesn't seem to provide an easy way to do it as it's not the intended use of iterators... so I'm just copying the remaining nibbles everytime anew, just to start iterating them from where we've left off.

P.S. Originally the constructor of `NotFound.ClosestAncestor` accepted the iterator only, since that suffices, but since we were copying anyway, I made it own the entire underlying `Nibbles` collection.

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -1,30 +1,480 @@
...
+            // Check if the parents match
+            {
+                var thisNodeParent = thisNode.parent;
+                var otherNodeParent = otherNode.parent;
+
+                //NOTE: this is awful, so a quick explainer:
+                //  We want to return false in all cases except:
+                //    - both parents are null;
+                //    - both parents are not null and the two nodes' child indices within them are the same
+                //  Reference: https://github.com/smol-dot/smoldot/blob/200214a571af30b5fa3997aea988451adc235ed0/lib/src/trie/trie_structure.rs#L615
```
David-Petrov: I'm open for suggestions here, that's really ugly but I have no better ideas.

src/main/java/com/limechain/trie/structure/Vacant.java
```diff
@@ -0,0 +1,482 @@
...
+                assert 1 == Arrays.stream(this.branchChildrenIndices).filter(Objects::nonNull).count()
+                    : "The branch node we're about to insert must have exactly one child (the node that previously existed before this insertion)";
+
+                // Insert the intermediate branch node
+                int newBranchNodeIndex = this.trieStructure.nodes.add(new TrieNode<>(
+                    this.branchParent,
+                    this.branchPartialKey,
+                    // NOTE: Crucial mutability moment, this has to be a copy of the array.
+                    //  I'll never get those three hours of my life back :)
+                    //  Manage your ownership, kids.
```
David-Petrov, line 444: I would remove this note, but I really want to emphasize the utmost importance of cloning the array here, since I had a very turbulent 2-3 hours debugging mutability issues since I had forgotten to clone it and was updating the same reference a few lines below. I'm open for suggestions on how to better annotate it, with no comment it appears totally random and inconspicuous.

build.gradle.kts
```diff
@@ -21,6 +21,15 @@ group = "com.limechain"
...
 
+tasks.withType<JavaCompile> {
+    options.compilerArgs.add("--enable-preview")
+}
+
+tasks.withType<JavaExec> {
+    jvmArgs("--enable-preview")
+}
+
+
```
georg-getz: remove empty line

src/main/java/com/limechain/client/FullNode.java
```diff
@@ -117,9 +133,110 @@ public List<InsertTrieNode> buildTrieNodesIterator(
...
+                    if (storageValueHashed.isPresent()) {
+                        storageValue = new StorageValue(storageValueHashed.get(), true);
+                    } else if (maybeStorageValue.isPresent()) {
+                        storageValue = new StorageValue(maybeStorageValue.get(), false);
+                    }
+                }
+
+                DecodedNode<Nibbles, List<Byte>> decoded = new DecodedNode<Nibbles, List<Byte>>(
+                    // TODO:
+                    //  All of this ugly boilerplate for a simple List<Optional<byte[]>> to List<Byte>[] conversion...
```
georg-getz: nit: it's transformed to `List<Byte[]>` not to `List<Byte>[]`

src/main/java/com/limechain/trie/NodeVariant.java
```diff
@@ -32,4 +34,16 @@ public enum NodeVariant {
     public int getPartialKeyLengthHeaderMask() {
         return this.mask ^ 0xFF;
     }
+
+    // NOTE:
+    //  Technically, for NodeVariant.COMPACT_ENCODING this should return `None`, but we still return 0...
+    //  since that's the result of the computation, but it's semantically invalid
```
georg-getz: Can't we add an if returning a `null` in that case?

src/main/java/com/limechain/trie/structure/BranchNodeHandle.java
```diff
@@ -0,0 +1,41 @@
...
+ * i.e. in the purely "structural necessity branch node" sense.
+ */
+public final class BranchNodeHandle<T> extends NodeHandle<T> {
+
+    // NOTE: An ugly workaround to simulate the notion of "consuming" this branch node handle
+    private boolean consumed;
+    BranchNodeHandle(TrieStructure<T> trieStructure, int nodeIndex) {
+        super(trieStructure, nodeIndex);
+        consumed = false;
+    }
```
georg-getz, line 13: Is this intentionally left as package-private?

src/main/java/com/limechain/trie/structure/BranchNodeHandle.java
georg-getz: Seems fine to me.

src/main/java/com/limechain/trie/structure/NodeHandle.java
georg-getz: This should be a whole team discussion. The reason I like Optionals is because the Rust implementations have a lot of them and it's way easier to translate them into Java Optional. The reason I like `nulls` is because they are usually easier to write and the `@Nullable` annotation makes it very hard to forget to check a nullable field for null. I'm also not sure whether completely removing either option is a good solution. We can discuss this on tomorrow's daily for example.

src/main/java/com/limechain/trie/structure/NodeHandle.java
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * @return true if the node has a storage value
+     */
+    public boolean hasStorageValue() {
+        // NOTE: We could do away with a simple `instanceof` check, but this is more explicit
+        return switch (this) {
+            case StorageNodeHandle<T> ignored -> true;
+            case BranchNodeHandle<T> ignored -> false;
+        };
+    }
```
georg-getz: `instanceof` generally indicates bad design. Wouldn't it be possible for this function to be made abstract and to have `StorageNodeHandle` and the branch one override it?

src/main/java/com/limechain/trie/structure/NodeHandle.java
```diff
@@ -0,0 +1,115 @@
...
+     * a {@link StorageNodeHandle} (if hasStorageValue is true) or a {@link BranchNodeHandle} (if false)
+     * @param hasStorageValue whether the node has a storage value
+     * @return the corresponding node handle constructor
+     */
+    static <T> BiFunction<TrieStructure<T>, Integer, NodeHandle<T>> getConstructor(boolean hasStorageValue) {
+        return hasStorageValue
+            ? StorageNodeHandle::new
+            : BranchNodeHandle::new;
+    }
+}
```
georg-getz: This whole functions seems strange to me, can't we just have a check for whether the boolean is true/false and invoke the necessary constructor based on that? Meaning getting completely rid of this function.

src/main/java/com/limechain/trie/structure/TrieNodeIndex.java
georg-getz: If this has no other use then maybe it would be better if it's an innner class to `TrieStructure`

src/main/java/com/limechain/trie/structure/TrieNodeIndex.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+@Data
+public final class TrieNodeIndex {
+    @Setter(AccessLevel.NONE)
+    private int value;
+
+    TrieNodeIndex(int value) {
+        this.value = value;
+    }
+}
```
georg-getz: If you make value `final` you can remove the constructor since `@Data` contains `@RequiredArgsConstructor`

src/main/java/com/limechain/trie/structure/TrieStructure.java
georg-getz: 2 things:

1. Do all of these need to be public? `streamUnordered` is only used privately for example.
2. Is performance the only benefit to the unordered functions? Meaning do we actually have anywhere in the code where we want the elements to be unordered? My guess is that the performance benefit might not be significant at all and if that's the case we could remove all the unordered functions.

src/main/java/com/limechain/trie/structure/TrieStructure.java
georg-getz: I don't know if there is a more "sane" approach for copying iterators. You can do a quick search about it but if nothing comes up just delete the todo.

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -1,30 +1,480 @@
...
+
+//        // Alternatively (and more functionally :D):
+//        return Stream.iterate(this.getNodeAtIndexInner(targetNodeIndex).parentIndex,
+//                Objects::nonNull,
+//                index -> this.getNodeAtIndexInner(index.getValue0()).parentIndex)
+//            .map(Pair::getValue0)
+//            .collect(Collectors.collectingAndThen(Collectors.toList(), list -> {
+//                Collections.reverse(list);
+//                return list;
+//            }));
```
georg-getz: Is this forgotten or just given as an example of how else it could look like?

src/main/java/com/limechain/trie/structure/TrieStructure.java
georg-getz: Just split it up a bit into more variables and it will become very easy to read:
```
boolean areBothNull = ...
boolean areBothNotNullWithSameChildIndex = ...
if (areBothNull || areBothNotNullWithSameChildIndex) doSmth() else false
```

src/main/java/com/limechain/trie/structure/Vacant.java
```diff
@@ -0,0 +1,482 @@
...
+                    throw new IllegalStateException("Unreachable state reached: ancestor found in an empty trie.");
+                }
+
+                // ... and no ancestor has been rightfully found, simply insert one node (will effectively be the root)
+                // This is kind of a special case that we handle by returning early.
+                return new PrepareInsert.One<>(
+                    this.trieStructure,
+                    null,
+                    key.copy(),
+                    new Integer[16]
```
georg-getz: You can take the 16 and make it a variable since it's used so often in this class. It's also a bit of a magic number

src/main/java/com/limechain/trie/structure/Vacant.java
georg-getz: What reference is being updated? `branchChildrenIndices`  seems to stay the same throughout.

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
georg-getz: I think we need a separate sync do discuss this. Is it possible to still implement `List<Nibble>` while just changing the functions to create the list as an unmodifiable one like so for example:
`Collections.unmodifiableList();` instead of `List.of();`

src/main/java/com/limechain/trie/structure/nibble/NibblesCollector.java
georg-getz: This is also going in the above mentioned sync.

src/main/java/com/limechain/utils/scale/ScaleUtils.java
```diff
@@ -69,5 +77,41 @@ public byte[] encode(byte[][] values) throws IOException {
...
+            return encodeAsList(ScaleCodecWriter::directWrite, bytes);
+        }
+
+        public byte[] encodeAsListOfListsOfBytes(Collection<Collection<Byte>> values) {
+            return encodeAsList(new CollectionWriter<>(ScaleCodecWriter::directWrite), values);
+        }
+
+//        public byte[] encodeListOfListsOfBytes(List<List<Byte>> values) {
+//            return encodeAsList(new ListWriter<>(ScaleCodecWriter::directWrite), values);
+//        }
```
georg-getz: Is this intentional?

src/main/java/com/limechain/utils/scale/writers/CollectionWriter.java
```diff
@@ -0,0 +1,24 @@
...
+import java.io.IOException;
+import java.util.Collection;
+
+public class CollectionWriter<T> implements ScaleWriter<Collection<T>> {
+    private final ScaleWriter<T> scaleWriter;
+
+    public CollectionWriter(ScaleWriter<T> scaleWriter) {
+        this.scaleWriter = scaleWriter;
+    }
+    @Override
```
georg-getz: you can leave an empty line before `@Override`

src/main/java/com/limechain/utils/scale/writers/IterableWriter.java
```diff
@@ -0,0 +1,24 @@
...
+import java.io.IOException;
+import java.util.Iterator;
+
+public class IterableWriter<T> implements ScaleWriter<Iterable<T>> {
+    private final ScaleWriter<T> scaleWriter;
+
+    public IterableWriter(ScaleWriter<T> scaleWriter) {
+        this.scaleWriter = scaleWriter;
+    }
+    @Override
```
georg-getz: same line comment

src/test/java/com/limechain/trie/structure/TrieStructureTest.java
```diff
@@ -0,0 +1,265 @@
...
...iIAIiIAIiIAKDR0CC/OCdE1kkAiIgAiIgAiIgAiIgAp4EJMh70lRZIiACIiACIiACIiACIlCDgAT5GvC0qwiIgAiIgAiIgAiIgAgMAQEJ8kNwkmSiCIiACIiACIiACIjAaBCQID8a51m1FAEREAEREAEREAERGF0CtQT5Cy+8MEyZMmV06anmIiACIiACIiACIiACIuBI4N///d/DvffeG2644Yaw2mqrOZbce1GzZs0KdBAsv/zy4eijj+69AO0hAiIgAiIgAiIgAiIgAiIwjsBbb70VNtpoo1RXf+ONN8ZttxVzdZJkX+aff/6Akq8kAiIgAiIgAiIgAiIgAiLgT2CQBHn/2qlEERABERABERABERABERABAt17FuTXXnvt8L73vU/0REAEREAEREAEREAEREAEHAgQlf7SSy8NVIT8QgstlEbKO1RPRYiACIiACIiACIiACIjAyBP44x//GGbMmNFfhPxrr70WFlhggZGHKAAiIAIiIAIiIAIiIAIi4EFAc8h7UFQZIiACIiACIiACIiACIjC4BGrNIS9BfnBPrCwTAREQAREQAREQAREYPgIS5IfvnMliERABERABERABERABEeiFgAT5XmgprwiIgAiIgAiIgAiIgAg0SECCfINwVbQIiIAIiIAIiIAIiIAIDAABCfIDcBJkggiIgAiIgAiIgAiIgAhAQIK8rgMREAEREAEREAEREAERmLMJSJCfs8+vaicCIiACIiACIiACIjBEBCTID9HJkqkiIAIiIAIiIAIiIAIi0AcBCfJ9QNMuIiACIiACIiACIiACItAEAQnyTVBVmSIgAiIgAiIgAiIgAiIwOAQkyA/OuZAlIiACIiACIiACIiACI05AgvyIXwCqvgiIgAiIgAiIgAiIwBxPQIL8HH+KVUEREAEREAEREAEREIFhISBBfljOlOwUAREQAREQAREQAREQgf4ISJDvj5v2EgEREAEREAEREAEREAF3AhLk3ZGqQBEQAREQAREQAREQAREYKAIS5AfqdMgYERABERABERABERCBUSYgQX6Uz77qLgIiIAIiIAIiIAIiMAoEJMiPwllWHUVABERABERABERABIaCgAT5oThNMlIEREAEREAEREAEREAE+iYgQb5vdNpRBERABERABERABERABHwJSJD35anSREAEREAEREAEREAERGDQCEiQH7Q
```
georg-getz: why? 😄

georg-getz: the image is present in 2 places

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
David-Petrov: > I think we need a separate sync do discuss this. Is it possible to still implement `List<Nibble>` while just changing the functions to create the list as an unmodifiable one like so for example: `Collections.unmodifiableList();` instead of `List.of();`

But the interface requires mutating methods like `public void add(int index, Nibble element)` for example... do you mean we go a similar way like Java's unmodifiable lists and simply throw runtime exceptions when attempting mutations? If so, that's an option worth considering.

David-Petrov: Another suggestion which I allow myself to think is viable in hindsight (coming from @stoyan-lime), is to simply abolish the whole idea for this convenience wrapper class and flow along with a simple `List<Nibble>`. 😄 Again, this will result in a lot of conversions outside, but if we manage to keep consistent use of `List`s overall (instead of arrays or other data structures), this might turn out simpler... and yes, we'd lose the custom methods, but we could still go with static utility methods which might even remain in this class (thus turning `Nibbles` into a utility class).

src/main/java/com/limechain/trie/structure/BranchNodeHandle.java
David-Petrov: > Seems fine to me.

Okay, in this case if we agree it's ok that's the approach I'm taking for `Vacant` as well. I'm awaiting final confirmation, still.

src/main/java/com/limechain/trie/structure/TrieNodeIndex.java
David-Petrov: > If this has no other use then maybe it would be better if it's an innner class to TrieStructure

Does `TrieStructure.NodeIndex` sound good? Another reason I've moved it away is to simplify the FQN of this class (since it's a rather significant part of the public interface and outside usage), while preserving enough of the context. Also, I've sought to keep the inner body of the `TrieStructure` class minimal as to avoid ending up with a 3000 lines single class, which is also the reason to the shift of "private as package-private" (i.e. all the classes in the `trie.structure` package are semantically private parts of the `TrieStructure`, just decomposed for readability along the separation of concerns).

src/main/java/com/limechain/trie/structure/TrieStructure.java
David-Petrov: I've already searched for it, the simple answer is "iterators are not meant to be copied", except that's exactly what I want. So the most commonly proposed solution is namely that: collect the remainder of the iterator into a list and continue iterating from the new list 😄 

I'm appending this concern as a little clarifier to the underlying note, since that's a possible future refactor point I think is worth considering.

src/main/java/com/limechain/trie/structure/TrieNodeIndex.java
David-Petrov: Good catch, that was part of my uncommited small fixes from Monday. :D

Only a single note: making it final is what I wanted, which removes the need for the `@Setter` annotation. But the constructor needs to remain, since it's package-private (only the TrieStructure constructs the indices, but everyone outside is intended to use them). This is also the reason it's not a `record`.

src/main/java/com/limechain/trie/structure/TrieStructure.java
David-Petrov: A little bit of both, but let's say prompting for another opinion was the reason it was not immediately deleted in the initial clean-up. I accept votes on which one is to remain. 😄

src/main/java/com/limechain/trie/structure/TrieStructure.java
David-Petrov: 1. It's for completeness. In principle, we could only leave the `stream_` versions of the methods, but when you want to use the TrieStructure as an iterable in an enhanced for-loop for example, you'd have to do it like:
```java
Iterable<TrieNodeIndex> orderedIndices = () -> trieStructure.streamOrdered().iterator();
for (TrieNodeIndex index : orderedIndices) {
    // some body...
}
```
(since java disallows expressions after the `:` in enhanced for loops), which is just not convenient. So it's a tradeoff between convenience of usage and cluttering the `TrieStructure`'s body.

2. Honestly, I don't have exact measures on the efficiency gains, but it seems there would be a rather significant difference, since traversing the trie in order is not trivial (and hence, I suspect, a rather expensive operation), although with the same (linear) complexity. 
The main arguments in favour of preserving the unordered iteration is: yes, there'll be a point when we'll use it - for proof encoding. The reference for this is the current state of Smoldot, which was also the reason these methods exist in the first place. Depends on how far we'd want to stick out the similarities between us and Smoldot. 😄

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -1,30 +1,483 @@
...
+                }
+            }
+
+            if (!thisNode.partialKey.equals(otherNode.partialKey)) {
+                return false;
+            }
+        }
+    }
+
+    // TODO: Figure out how to sensibly deprecate this or whether it should be equivalent to structurallyEquals
```
David-Petrov: This overridden deprecated `equals` is to be removed, but I'm not sure how. I suspect my initial idea to dissuade the users to ever use it is rather stupid.

The idea is: we want to have two separate "equals" methods - one is `structurallyEquals`, the other doesn't exist yet, but will include comparison of the `userData` as well. 

Figure out a decent decision.

src/main/java/com/limechain/trie/structure/TrieStructure.java
David-Petrov: Good one. I've still left the clarifying comment above the block, I reckoned a natural language explanation could still be helpful.

src/test/java/com/limechain/trie/structure/TrieStructureTest.java
David-Petrov: > why? 😄

I forgot to remove this note, it was just a reaction. 😄  I'm kind of very keen on the visual capabilities of JavaDoc, and I think that's a sensible use case.

David-Petrov: > the image is present in 2 places

If you mean the other test above, it's a different picture. If you don't mean this, I kind of don't get the comment. 

But yeah, the idea being that images visualise what the expected trie at the end looks like, as an additional clarification to the assertions.

src/main/java/com/limechain/utils/scale/ScaleUtils.java
David-Petrov: `ScaleUtils` remains to be cleaned up, this is not intentional.

src/main/java/com/limechain/trie/structure/Vacant.java
David-Petrov: I thought about this, but honestly, in the context of looking through this class and what it does (collecting data to insert nodes), the plain number was a much more clear indicator of "ah yes, those are some node's children" (since I know they're exactly 16) than any variable name I could think of. The literal `16` in and of itself seemed like a fundamental enough name (again, in this context 😄).

What would a reasonable name for this magic number be? `TRIE_NODE_CHILDREN_COUNT`?

src/main/java/com/limechain/trie/structure/NodeHandle.java
David-Petrov: Yeah, well... At the cost of this one-liner method, we can change the previous approach (of what you suggested) from:
```java
        NodeHandle<T> nodeHandle;
        if (node.hasStorageValue) {
            nodeHandle = new StorageNodeHandle<T>(this, nodeIndex);
        } else {
            nodeHandle = new BranchNodeHandle<T>(this, nodeIndex);
        }
```
or using the ternary
```java
        NodeHandle<T> nodeHandle = node.hasStorageValue 
            ? new StorageNodeHandle<T>(this, nodeIndex)
            : new BranchNodeHandle<T>(this, nodeIndex);
```
to (the to me more concise):
```java
NodeHandle<T> nodeHandle = NodeHandle.<T>getConstructor(node.hasStorageValue).apply(this, nodeIndex)
```

For a total of 3 occurrences (for now), I thought that was a sensible solution (my functional-paradigm biases aside). But yeah, ultimately not that big of a deal. 😄

src/main/java/com/limechain/trie/structure/NodeHandle.java
David-Petrov: Yes, we totally could. Let's say that's the Rust influence taking over (plus the design around sealed classes allows for this alternative). Do we want to generally switch to abstract methods? 
- If so, this would apply to a few other places, too. In general, I tend to like this approach more, since abstract methods tend to add indirection (at the very least while inspecting code), and in this particular case: no sensible user would ask the concrete child class `StorageNodeHandle` if it has a storage value - it's kind of obvious, that's its whole point. 😄 So this method only makes sense for the parent class `NodeHandle`.
- But if it's only for this particular occurence (in order to avoid `instanceof` / pattern matching on `this`), I find a good reason for the abstract method.

src/main/java/com/limechain/trie/structure/BranchNodeHandle.java
David-Petrov: > Is this intentionally left as package-private?

Yes, since only the trie structure can construct its handles, whilst they're intended to be used outside by the user.
For the detailed clarification of why some things are package-private: [reference](https://github.com/LimeChain/Fruzhin/pull/265#discussion_r1442929751).

src/main/java/com/limechain/client/FullNode.java
David-Petrov: This will be changed soon as changes are incoming from Bobi's PR regarding the DecodedNode. But I don't think you're right though... 😄 Or I don't get what you mean.

src/main/java/com/limechain/trie/NodeVariant.java
David-Petrov: Well, that would include switching the return type from `int` to `Integer`... Do you think it's worth it?
Personally, I think a better alternative (if any) to this method would be to extract it away from `NodeVariant` (i.e. to not be a member method) to its single use case which kind of knows that `COMPACT_ENCODING` is not an option. 😄

src/main/java/com/limechain/trie/structure/BranchNodeHandle.java
georg-getz: Confirmation given

src/main/java/com/limechain/trie/structure/TrieNodeIndex.java
georg-getz: You can leave it as is then.

src/main/java/com/limechain/trie/structure/TrieStructure.java
georg-getz: I think it's fine to leave it as it is then. There might be multiple functions but all of them have small bodies so it's not taking that much space.

src/main/java/com/limechain/client/FullNode.java
georg-getz: I'm talking about the comment. It is saying that it's a conversion from
`List<Optional<byte[]>>` to `List<Byte>[]`                           whereas it should say
`List<Optional<byte[]>>` to `List<Byte[]>`

when you see "nit:" in a PR comment it means it's something almost irrelevant but still worth correcting

src/main/java/com/limechain/trie/NodeVariant.java
georg-getz: Sure, I didn't know this only has a singular use case.

src/main/java/com/limechain/trie/structure/NodeHandle.java
georg-getz: I think making it abstract is the best approach here. You mention that there would be no point calling `StorageNodeHandle.hasStorageValue()` but this implies that the reference we are working with would be of the child class which should (probably) **never** be the case. In Java we always want to use parent references when creating objects so that we can easily use polymorphism(e.g. `NodeHandle nv = new StorageNodeHandle()`).

src/main/java/com/limechain/trie/structure/NodeHandle.java
georg-getz: Ok you can leave it as is.

src/main/java/com/limechain/trie/structure/TrieStructure.java
georg-getz: I'm fine with either, just make sure to delete one of them.

src/main/java/com/limechain/trie/structure/Vacant.java
georg-getz: Yea, that seems like a fine name.

src/main/java/com/limechain/trie/structure/TrieStructure.java
georg-getz: Can't we override the equals method to have the body of `structurallyEquals` (or the second one since it's checking for more fields) for example?

src/main/java/com/limechain/trie/structure/TrieStructure.java
David-Petrov: I've removed the commented out version.

On a separate note, the `LinkedList` business got me thinking - the actual return type of this method could very well be a `Deque`, thus easily providing traversal capabilities in both directions (as of right now, we use `nodePath` only to traverse from root to target, so that's what it originally returns). So we gain on intentionality.
Using a Deque however totally removes the need to revert the list at the end, so that's also a benefit in simplicity.

Feel free to inspect the commit 335295f5 and resolve this convo if all is fine.

src/main/java/com/limechain/trie/structure/Vacant.java
David-Petrov: The variable itself doesn't, but exactly in this line we're passing it as the children indices for the new branch node, crucially, without having the new storage node inserted. That's why copying the old snapshot is very important, since two lines below we're also adding the freshly created storage node to its parent's children indices array:
https://github.com/LimeChain/Fruzhin/blob/0b505ce950e737ce64882c305834b045cab27045/src/main/java/com/limechain/trie/structure/Vacant.java#L459-L461

Thus, if we didn't copy, this mutates the original variable as well, which in turn breaks the loop where we're updating its **old** children:
https://github.com/LimeChain/Fruzhin/blob/0b505ce950e737ce64882c305834b045cab27045/src/main/java/com/limechain/trie/structure/Vacant.java#L463

The crucial difference between the old children and this new storage node child is their partial keys, which got messed up - a single nibble missing in seemingly random places, which was the result of the above-mentioned lack of copying.

src/main/java/com/limechain/trie/NodeVariant.java
David-Petrov: Well, after getting to actually refactoring this, it turned out we don't need this method at all, since what I'm using it for outside is already exactly solved by the already existing method `NodeVariant::getPartialKeyLengthHeaderMask`. You can inspect the final version here: 598e6e9, and resolve this if all is okay.

src/main/java/com/limechain/trie/structure/TrieStructure.java
David-Petrov: > Can't we override the equals method to have the body of `structurallyEquals` (or the second one since it's checking for more fields) for example?

Of course we can... but do we want? 😅 I'm personally against the existence of a method called `equals()` in a scenario where no singular sensible equality is possible, I'd prefer different names for explicitness. But Java will always require an `equals()`, so it's our call. I'm leaving this up to you - easiest solution would be to implement it as `structurallyEquals`. Do we proceed that way?

src/main/java/com/limechain/trie/structure/NodeHandle.java
David-Petrov: That's generally true, but keep in mind we're going for `sealed class`es in the whole `Entry` hierarchy below, including `NodeHandle`s. So, given a NodeHandle, we can always exhaustively pattern match which actual child class this instance is of (as seen from the method implementation in question). In this case, we lose nothing from the polymorphism, and (what I meant above) I'd even argue we're gaining on minimalism, since we're not introducing any methods into `Storage/BranchNodeHandle` that don't make sense for the respective child class itself (thus adding to their public interfaces what only makes sense for `NodeHandle`'s public interface). 😄
What we lose, well, is this pattern matching which might seem a bit obfuscating (this `ignored` necessity in Java kind of bugs me) depending on perceptions of readability and habits.

Anyway, I just wanted to make sure we're on the same page on that detail. Ultimately, I don't mind either that much (well, maybe I've got a slight bias against abstract methods, especially in such situations where the parent implementor can know it all 😄) given the relatively little relevance of this detail here in particular, but it's a good general discussion.

Just confirm and I'll refactor, unless I've managed to win you over. 😄

P.S. On a separate note, this method (in its current version) should probably be marked `final` as well. 🤔

src/main/java/com/limechain/trie/structure/BranchNodeHandle.java
David-Petrov: Okay, so the case for `Vacant` turns out to be a rather more complicated one, since to invalidate the instance itself, it'll have to happen from within `PrepareInsert.One/.Two`. I feel this kinda stinks, so I'd suggest we make ourselves happy with only the `@implNote`. On the other hand, if we really want to be defensive about this, it's going to take a little more reference cluttering, which might damage readability... I'm not sure it's worth it, since such cases are really intentional misuse. 😄

src/main/java/com/limechain/trie/NodeVariant.java
georg-getz: nice

src/main/java/com/limechain/trie/structure/TrieStructure.java
georg-getz: I think `structurallyEquals` is always going to be a separate function so you can just continue adding it, but add a note or todo for when the second function is implemented. There I think it would make more sense to use the equals function.

src/main/java/com/limechain/utils/scale/ScaleUtils.java
David-Petrov: I've done all the refactoring necessary, but afterwards I thought about the breaking nature of some of my changes (main suspects: method renamings). Merging the refactored code in such a state would break the build on `dev`, so I'd propose we skip this part until we merge and apply the changes afterwards in a new PR to `dev`.

src/main/java/com/limechain/trie/structure/TrieStructure.java
David-Petrov: I'm not sure what implementing `equals` in either way would imply for the advisably corresponding `hashCode`. I feel like currently no sensible `hashCode` implementation could exist (without going in a bit more depth for unused stuff right now), so overriding `equals` with different behaviour, but not `hashChode` would break the contract on the notion of 'equality'. This makes me think we should leave `equals` deprecated.

src/main/java/com/limechain/utils/scale/ScaleUtils.java
David-Petrov: That's what we'll do.

src/main/java/com/limechain/trie/structure/TrieStructure.java
David-Petrov: I see three options:
1) throw `UnsupportedOperationException` from both `equals` and `hashCode`;
2) just leave them as the default behaviour (i.e. don't override);
3) actually figure out what would make sense for `equals` and `hashCode`. Let's say `equals` is the easier part. But this is work I'm willing to put off for later.

I'm going with option 2.

David-Petrov: Regarding option 3 (the actual good approach):
we could make `equals` a superset of `structurallyEquals`, but including comparison of `userData`s. Easy way to achieve this:
1) implement `equals` for `TrieNode`'s, basically similar to a record's `equals`, comparing fields;
2) within the TrieStructure, iterate nodes lexicographically and compare the collections obtained.

But this'll remain for later. I'm intentionally not resolving this conversation for traceability's sake.

src/main/java/com/limechain/trie/structure/NodeHandle.java
David-Petrov: We left this thread intentionally unresolved for later traceability.

PR #267, opened 2024-01-04
Feat/implement node decoding

bokoto000:


src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
```diff
@@ -228,4 +240,221 @@ public byte[] calculateMerkleValue(Function<byte[], byte[]> hashFunction, boolea
...
+        if (encoded == null || encoded.length == 0) {
+            throw new NodeDecodingException("Invalid node value: it's empty");
+        }
+
+        int firstByte = reader.readByte() & 0xFF;
+        boolean hasChildren;
+        Boolean storageValueHashed;
+        int pkLenFirstByteBits;
+
+        //https://spec.polkadot.network/#defn-node-header
```
vikinatora: Let's keep code style consistent and extract parts of the decode logic into separate functions just like in `encode` above.
image

vikinatora:
```suggestion
        // https://spec.polkadot.network/#defn-node-header
```

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
```diff
@@ -228,4 +240,221 @@ public byte[] calculateMerkleValue(Function<byte[], byte[]> hashFunction, boolea
...
+
+        int childrenBitmap = extractChildrenBitmap(reader, hasChildren);
+
+        byte[] storageValue = extractStorageValue(reader, storageValueHashed);
+
+        List<List<Byte>> children = extractChildren(reader, childrenBitmap);
+        Nibbles partialKey = Nibbles.fromLEBytes(partialKeyLEBytes);
+        if (pkLen % 2 == 1) {
+            /// This is for situations where the partial key contains a `0` prefix that exists for
+            /// alignment but doesn't actually represent a nibble.
```
vikinatora: Please use consistent comment style
```suggestion
            // This is for situations where the partial key contains a `0` prefix that exists for
            // alignment but doesn't actually represent a nibble.
```

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
```diff
@@ -228,4 +240,221 @@ public byte[] calculateMerkleValue(Function<byte[], byte[]> hashFunction, boolea
...
+
+    /**
+     * Extracts the partial key from a node in a trie structure using a ScaleCodecReader.
+     *
+     * @param reader The ScaleCodecReader used to read the encoded node data.
+     * @param pkLen  The length of the partial key in nibbles.
+     * @return byte[] The byte array containing the partial key.
+     * @throws NodeDecodingException If there are no more bytes to read, or if the partial key has invalid padding.
+     */
+    public static byte[] extractPartialKey(ScaleCodecReader reader, int pkLen) {
```
vikinatora: Should this function be public?

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
```diff
@@ -228,4 +240,221 @@ public byte[] calculateMerkleValue(Function<byte[], byte[]> hashFunction, boolea
...
+
+    /**
+     * Extracts the children bitmap from a node in a trie structure.
+     *
+     * @param reader      The ScaleCodecReader used to read the encoded node data.
+     * @param hasChildren A boolean indicating whether the node has children.
+     * @return int The children bitmap as an integer. If the node has no children, returns 0.
+     * @throws NodeDecodingException If there are no more bytes to read, or if the children bitmap is zero.
+     */
+    public static int extractChildrenBitmap(ScaleCodecReader reader, boolean hasChildren) {
```
vikinatora: Should this be public static ?

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
```diff
@@ -228,4 +240,221 @@ public byte[] calculateMerkleValue(Function<byte[], byte[]> hashFunction, boolea
...
+
+    /**
+     * Extracts the storage value from a node in a trie structure.
+     *
+     * @param reader             The ScaleCodecReader used to read the encoded node data.
+     * @param storageValueHashed A Boolean indicating whether the storage value is hashed. Can be null, true, or false.
+     * @return byte[] The storage value as a byte array, or null if there is no storage value.
+     * @throws NodeDecodingException If there are no more bytes to read from the reader.
+     */
+    public static byte[] extractStorageValue(ScaleCodecReader reader, Boolean storageValueHashed) {
```
vikinatora: Same as above

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
```diff
@@ -228,4 +240,221 @@ public byte[] calculateMerkleValue(Function<byte[], byte[]> hashFunction, boolea
...
+
+    /**
+     * Extracts the children of a node in a trie structure.
+     *
+     * @param reader         The ScaleCodecReader used to read the encoded node data.
+     * @param childrenBitmap An integer bitmap representing the presence of children. Each bit corresponds to a child.
+     * @return List<List < Byte>> A list of children, where each child is represented as a list of bytes.
+     * @throws NodeDecodingException If the length of any child's data exceeds the maximum allowed size.
+     */
+    public static List<List<Byte>> extractChildren(ScaleCodecReader reader, int childrenBitmap) {
```
vikinatora: Same as above

vikinatora (changes requested):

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
```diff
@@ -67,6 +68,15 @@ public static Nibbles of(int... nibbles) {
...
     }
 
+    public static Nibbles fromLEBytes(byte[] bytes) {
+        Nibbles nibbles = new Nibbles(new byte[]{});
+        for (int i = 0; i < bytes.length; i++) {
+            nibbles.add(Nibble.fromInt((bytes[i] >> 4 & 0xf)));
+            nibbles.add(Nibble.fromInt((bytes[i] & 0xf)));
+        }
+        return nibbles;
+    }
```
David-Petrov: That's what `BytesToNibbles` is for. Ideally, I'd suggest we keep the Nibbles class itself behaving like a collection, thus extract generational logic outside... or at least that was my intention. You could replace any invocations of `Nibbles.fromLEBytes(arr)` with `new Nibbles(new BytesToNibbles(arr))` for now. 

There will be additional refactoring to the entire Nibbles class anyway, it's a bit messy as of right now :D

src/main/java/com/limechain/client/FullNode.java
```diff
@@ -223,7 +223,7 @@ TrieStructure<Pair<Optional<byte[]>, Optional<byte[]>>> initializeDatabaseStorag
                     // TODO:
                     //  All of this ugly boilerplate for a simple List<Optional<byte[]>> to List<Byte>[] conversion...
                     //  figure out a better representation
-                    children.stream().map(p -> p.map(ba -> new ArrayList<>(List.of(ArrayUtils.toObject(ba)))).orElse(null)).toArray(ArrayList[]::new),
+                        List.of(children.stream().map(p -> p.map(ba -> new ArrayList<>(List.of(ArrayUtils.toObject(ba)))).orElse(null)).toArray(ArrayList[]::new)),
```
David-Petrov: [Nitpick]
```suggestion
children.stream().map(p -> p.map(ba -> new ArrayList<>(List.of(ArrayUtils.toObject(ba)))).orElse(null)).toList()
```

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
```diff
@@ -228,4 +240,221 @@ public byte[] calculateMerkleValue(Function<byte[], byte[]> hashFunction, boolea
...
+    /***
+     * Decodes a node value found in a proof into its components.
+     * This can decode nodes no matter their version or hash algorithm.
+     *
+     * @param encoded The byte array representing the encoded node.
+     * @return DecodedNode<Nibbles, List < Byte>> The decoded node with its partial key, children, and storage value.
+     * @throws NodeDecodingException If the encoded array is null, empty, or invalid.
+     * @throws IllegalArgumentException If 'encoded' is null.
+     */
+    public static DecodedNode decode(byte[] encoded) {
```
David-Petrov: I think this return type shouldn't be raw, we should specify the generic parameters since `decode` is a constructing method. On that note, perhaps `decode` could also become generic, but for now that might not be needed at all.

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
```diff
@@ -228,4 +240,221 @@ public byte[] calculateMerkleValue(Function<byte[], byte[]> hashFunction, boolea
 
         return nodeValue;
     }
+
+    /***
+     * Decodes a node value found in a proof into its components.
+     * This can decode nodes no matter their version or hash algorithm.
+     *
+     * @param encoded The byte array representing the encoded node.
+     * @return DecodedNode<Nibbles, List < Byte>> The decoded node with its partial key, children, and storage value.
```
David-Petrov: [Possible nitpick]
Wrap those generics in a {@code ... } to avoid the HTML trouble with the angle brackets in the JavaDoc :D

```suggestion
     * @return {@code DecodedNode<Nibbles, List <Byte>>} The decoded node...
```

Although I generally stray away from repeating the return type in the JavaDoc, it's kinda visible from the signature.

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
David-Petrov: We reached the conclusion it's okay as is, as far as possible what's extractable has been extracted.

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
bokoto000: Changed them to private but I'd like to leave them static as they do not need an instance

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
bokoto000: The decoding logic in smoldot is not generic as the children are saved as byte array and the partial key is saved as nibbles so I suggest we keep this non-generic.

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
```diff
@@ -228,4 +239,218 @@ public byte[] calculateMerkleValue(Function<byte[], byte[]> hashFunction, boolea
...
+     */
+    public static DecodedNode<Nibbles, List<Byte>> decode(byte[] encoded) {
+        ScaleCodecReader reader = new ScaleCodecReader(encoded);
+        if (encoded == null || encoded.length == 0) {
+            throw new NodeDecodingException("Invalid node value: it's empty");
+        }
+
+        int firstByte = reader.readByte() & 0xFF;
+        boolean hasChildren;
+        Boolean storageValueHashed;
```
David-Petrov, line 260: [Possible nitpick] I'd add a NOTE for some kind of intentionality indication that `null` is a semantically valid value of this Boolean. I guess you've considered the `Optional<Boolean>` approach, although in Java that's understandably controversial and I can see why you've gone for the simpler nullable type :D

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
```diff
@@ -228,4 +239,218 @@ public byte[] calculateMerkleValue(Function<byte[], byte[]> hashFunction, boolea
...
+
+    /**
+     * Extracts the storage value from a node in a trie structure.
+     *
+     * @param reader             The ScaleCodecReader used to read the encoded node data.
+     * @param storageValueHashed A Boolean indicating whether the storage value is hashed. Can be null, true, or false.
+     * @return byte[] The storage value as a byte array, or null if there is no storage value.
+     * @throws NodeDecodingException If there are no more bytes to read from the reader.
+     */
+    private static byte[] decodeStorageValue(ScaleCodecReader reader, Boolean storageValueHashed) {
```
David-Petrov: Again, for intentionality (although mentioned in the Javadoc, but could be clarified a bit more):

```suggestion
    private static byte[] decodeStorageValue(ScaleCodecReader reader, @Nullable Boolean storageValueHashed)
```

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
```diff
@@ -228,4 +239,218 @@ public byte[] calculateMerkleValue(Function<byte[], byte[]> hashFunction, boolea
...
+
+        return storageValue;
+    }
+
+    /**
+     * Extracts the children of a node in a trie structure.
+     *
+     * @param reader         The ScaleCodecReader used to read the encoded node data.
+     * @param childrenBitmap An integer bitmap representing the presence of children. Each bit corresponds to a child.
+     * @return List<List < Byte>> A list of children, where each child is represented as a list of bytes.
```
David-Petrov: [Nitpick] `* @return {@code List<List <Byte>>} ...`

src/test/java/com/limechain/trie/structure/decoded/node/DecodedNodeTest.java
```diff
@@ -0,0 +1,65 @@
...
+                77, 26, (byte) 166, 1, 52, (byte) 179, (byte)173, 43, 89, (byte) 219, 104, (byte) 196, (byte) 190,
+                (byte) 208, (byte) 128, (byte) 135, (byte) 177, 13, (byte) 185, 111, (byte) 175};
+        List<Byte> boxedExpectedEncoded = Arrays.asList(ArrayUtils.toObject(expectedEncoded));
+        List<Byte> boxedExpectedChild1 = Arrays.asList(ArrayUtils.toObject(expectedChild1));
+        List<Byte> boxedExpectedChild2 = Arrays.asList(ArrayUtils.toObject(expectedChild2));
+        StorageValue expectedStorageValue = new StorageValue(new byte[]{}, false);
+        int expectedNotNullChildren = 2;
+        Nibbles expectedNibbles = Nibbles.EMPTY;
+        expectedNibbles.add(Nibble.fromByte((byte) 0x6));
+        expectedNibbles.add(Nibble.fromByte((byte) 0x3));
```
David-Petrov: Unfortunately, due to the currently unfinished bad design of Nibbles, this operation mutates the public static variable `Nibbles.EMPTY`, which will be deprecated anyway pretty soon. For this exact testing purpose however, we've got this convenience method (renaming of the variable included :D):
```suggestion
Nibbles expectedPartialKey = Nibbles.fromHexString("63");
```

src/test/java/com/limechain/trie/structure/decoded/node/DecodedNodeTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+        DecodedNode<Nibbles, List<Byte>> decoded = DecodedNode.decode(expectedEncoded);
+        assertNotNull(decoded);
+
+        assertArrayEquals(boxedExpectedEncoded.toArray(), decoded.encode()
+                .flatMap(Collection::stream).collect(Collectors.toList()).toArray());
```
David-Petrov: [Nitpick] Simply `equals()` works good enough for lists, I'd do sth like:

```suggestion
        List<Byte> actualEncoded = decoded.encode().flatMap(Collection::stream).toList();
        assertEquals(boxedExpectedEncoded, actualEncoded);
```

src/test/java/com/limechain/trie/structure/decoded/node/DecodedNodeTest.java
```diff
@@ -0,0 +1,65 @@
...
+        assertEquals(expectedNibbles, decoded.getPartialKey());
+
+        StorageValue storageValue = decoded.getStorageValue();
+        assertEquals(expectedStorageValue, storageValue);
+
+        List<List<Byte>> children = decoded.getChildren();
+        List<List<Byte>> notNullChildren = children.stream()
+                .filter(element -> element != null)
+                .collect(Collectors.toList());
+        assertEquals(expectedNotNullChildren, notNullChildren.size());
```
David-Petrov: [Totally a nitpick] If you happen to like it more:
```suggestion
        int notNullChildrenCount = children.stream().filter(Objects::nonNull).toList().size();
        assertEquals(expectedNotNullChildrenCount, notNullChildrenCount);
```

src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java
```diff
@@ -228,4 +239,219 @@ public byte[] calculateMerkleValue(Function<byte[], byte[]> hashFunction, boolea
...
+     * Calculates the length of the partial key in nibbles
+     *
+     * @param reader             ScaleCodecReader used to read the encoded node data.
+     * @param firstByte          The first byte of the encoded node data.
+     * @param pkLenFirstByteBits The number of bits in the first byte dedicated to encoding the partial key length.
+     * @return int The length of the partial key.
+     * @throws NodeDecodingException If there's an error in decoding, such as if the partial key length is too short,
+     *                               or if there's an overflow in calculating the partial key length.
+     */
+    private static int getPkLen(ScaleCodecReader reader, int firstByte, int pkLenFirstByteBits) {
```
David-Petrov: Perhaps rename this method to `decodePkLen` for consistency?

David-Petrov (approved):

PR #269, opened 2024-01-08
Clean up node encoding

David-Petrov:


David-Petrov:
https://github.com/LimeChain/Fruzhin/blob/1d6063f408e04e4e7da814f1a88d454250558f1e/src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java#L70-L87

I find this awful to read, so I'm open for suggestions on how to approach it better.

David-Petrov:
https://github.com/LimeChain/Fruzhin/blob/1d6063f408e04e4e7da814f1a88d454250558f1e/src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java#L128-L136

I'm sincerely confused by this little detail, but I guess we could simply disregard this TODO (or turn it into a NOTE for intentionality purposes). Any opinions are welcome.

David-Petrov:
https://github.com/LimeChain/Fruzhin/blob/1d6063f408e04e4e7da814f1a88d454250558f1e/src/main/java/com/limechain/trie/structure/decoded/node/DecodedNode.java#L138-L154

That's another weird instance of "I have no idea why we're doing this that way".

David-Petrov:
Generally, one last thing remains suspicious: too little SCALE encoding is happening overall when encoding the node to obtain it's `node_value`, compared to what the spec prescribes. Especially for [the subvalue part](https://spec.polkadot.network/chap-state#defn-node-subvalue), which has also previously been noticed to be incorrect. So if we're okay with being able to replicate what happens in Smoldot for example, then we're good with that and could also not bother.

georg-getz (approved):

PR #271, opened 2024-01-11
Refactor nibbles

David-Petrov:
refactor(nibbles)!: 

* make Nibbles immutable
* turn constructors into public factory methods for Nibbles that copy data on construction to gain ownership, instead leaving the constructor package protected in order to still make us of internal list's mutability when modifying the Nibbles.
* `Nibbles implements Iterable<Nibble>` instead of Collection or List
* remove generic partial key type argument from `DecodedNode` since in Java's eyes `Nibbles` is no more a collection, thus we can't access its `size()` method generically. Instead, we simply specify that generic type to Nibbles
* make `NibblesToBytes` a static utility class

### DIFF
--- a/src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
+++ b/src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
@@ -1,86 +1,78 @@
 package com.limechain.trie.structure.nibble;
 
 import org.apache.commons.collections4.IteratorUtils;
-import org.apache.commons.lang3.ArrayUtils;
 import org.jetbrains.annotations.NotNull;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
-import java.util.ListIterator;
 import java.util.Objects;
 import java.util.RandomAccess;
 import java.util.stream.Collector;
 import java.util.stream.Stream;
-
-// TODO:
-//  Should this implement List or just Iterable... or maybe extend List<Nibble>?
-//  This current version contains a huge amount of boilerplate,
-//  but it's what I've reached as a most convenient option to use outside.
-//  Basically a List<Nibble> with additional methods.
-
-// TODO:
-//  As a List<Nibble> implementor, Nibbles is currently mutable, which I severely dislike,
-//  since this mandates the developer to think about where to copy and where passing by reference is sufficient.
+import java.util.stream.StreamSupport;
 
 /**
- * Convenience wrapper for any 'sequence of Nibble'-like structure
+ * Convenience wrapper for any 'sequence of Nibble'-like structure.
+ * It's immutable and its public static factory methods eagerly copy passed data (nibbles) to obtain ownership.
  */
-public class Nibbles implements List<Nibble>, RandomAccess, Comparable<List<Nibble>> {
-    public static final Nibbles EMPTY = new Nibbles(new byte[]{});
+public class Nibbles implements Iterable<Nibble>, RandomAccess, Comparable<Iterable<Nibble>> {
+    /**
+     * A sequence of zero nibbles, i.e. empty
+     */
+    public static final Nibbles EMPTY = new Nibbles(List.of());
 
     /**
-     * Don't mutate! :D
+     * A sequence of all nibbles in order, from 0 to F
      */
-    public static final Nibbles ALL = new Nibbles(Nibble.all());
-
-    public static Stream<Nibble> all() {
-        return ALL.stream();
-    }
+    public static final Nibbles ALL = Nibble.all().collect(NibblesCollector.toNibbles());
 
     private final List<Nibble> nibbles;
 
-    // TODO: Maybe convert all those overloaded constructors to static generational methods
-    public Nibbles(Nibble nibble) {
-        this(List.of(nibble));
+    public static Nibbles of(Nibble nibble) {
+        return Nibbles.of(List.of(nibble));
     }
 
-    public Nibbles(Nibble[] nibbles) {
-        this(Arrays.asList(nibbles));
+    public static Nibbles of(Nibble[] nibbles) {
+        return Nibbles.of(Arrays.asList(nibbles));
     }
 
-    public Nibbles(byte[] nibbles) {
-        this(Arrays.stream(ArrayUtils.toObject(nibbles)).map(Nibble::fromByte).toList());
+    public static Nibbles of(Collection<Nibble> nibbles) {
+        return new Nibbles(new ArrayList<>(nibbles));
     }
 
-    public Nibbles(Iterable<Nibble> nibbles) {
-        this(nibbles.iterator());
+    public static Nibbles of(Stream<Nibble> nibbles) {
+        return Nibbles.of(nibbles.iterator());
     }
 
-    public Nibbles(Stream<Nibble> nibbles) {
-        this(nibbles.iterator());
+    public static Nibbles of(Iterable<Nibble> nibbles) {
+        return Nibbles.of(nibbles.iterator());
     }
 
-    public static Nibbles of(int... nibbles) {
-        return new Nibbles(Arrays.stream(nibbles).mapToObj(Nibble::fromInt));
+    public static Nibbles of(Iterator<Nibble> nibbles) {
+        return new Nibbles(IteratorUtils.toList(nibbles));
     }
 
-    public Nibbles(Iterator<Nibble> nibbles) {
-        // TODO:
-        //  Reconsider the internal representation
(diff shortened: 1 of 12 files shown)


src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
```diff
@@ -92,191 +84,148 @@ public boolean startsWith(Nibbles prefix) {
...
+     * Adds a new nibble to the nibbles.
+     * @param nibble the new nibble to add
+     * @return the new Nibbles after insertion
+     * @implNote will create a new modified copy of the nibbles, does not mutate the instance invoked on
+     */
+    public Nibbles add(Nibble nibble) {
+        List<Nibble> newNibbles = new ArrayList<>(this.nibbles);
+        newNibbles.add(nibble);
+        return new Nibbles(newNibbles);
     }
```
David-Petrov, line 102: I don't like how mutable this method name sounds, but I decided it should be named after what it actually does and not after how it certainly behaves. Resolve if you don't take issue in that.

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
```diff
@@ -92,191 +84,148 @@ public boolean startsWith(Nibbles prefix) {
...
         return nibbles.get(index);
     }
 
-    @Override
-    public Nibble set(int index, Nibble element) {
-        return nibbles.get(index);
+    // NOTE:
+    //  Since Nibbles is immutable, we don't actually need to explicitly copy
+    //  This method could go away, passing by reference could suffice, although going back to mutability
+    //  if later decided would become rather difficult.
```
David-Petrov: That's an interesting conundrum, take a look at the note.

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
```diff
@@ -92,191 +84,148 @@ public boolean startsWith(Nibbles prefix) {
...
+     * Takes only the first n nibbles from the beginning.
+     * @param n the number of nibbles to take
+     * @return a new Nibbles limited to only the first n leading nibbles
+     * @throws IndexOutOfBoundsException if
+     *         ({@code n < 0 || n > size})
+     */
+    public Nibbles take(int n) {
+        // NOTE: We might want to not throw, but return the same Nibbles instead if (n > size).
+        return new Nibbles(this.nibbles.subList(0, n));
     }
```
David-Petrov: I couldn't decide on the behaviour I prefer for these two (although it doesn't matter as much, but still).
Since these methods come from rather abundant outside usage of `.stream().skip(n).collect(NibblesCollector.toNibbles())`, it makes sense for them to behave like `Stream`'s skip / take (hence not throwing anything). 
On the other hand, turning those into methods of Nibbles allows us to make use of the underlying list and only copy a view of the sublist we're interested in and simply wrap it into a new nibbles (without unnecessary copying). But `subList` throws exceptions, so it made natural sense to do this as well.

Ultimately, we could add a simple check for the boundaries and act stream-likey while still taking advantage of `subList`.

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
```diff
@@ -92,191 +84,148 @@ public boolean startsWith(Nibbles prefix) {
...
-
-    @Override
-    public boolean addAll(int index, @NotNull Collection<? extends Nibble> c) {
-        return nibbles.addAll(c);
-    }
-
-    @Override
-    public boolean removeAll(@NotNull Collection<?> c) {
-        return nibbles.removeAll(c);
+    private static String toLowerHexString(Iterable<Nibble> nibbles) {
```
David-Petrov: I don't like this method being in here, but couldn't figure out a better place. If you think it's ok, let it be.

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
```diff
@@ -92,191 +84,148 @@ public boolean startsWith(Nibbles prefix) {
...
-    public void add(int index, Nibble element) {
-        nibbles.add(index, element);
+    /**
+     * Drops a number of nibbles from the beginning
+     * @param n the number of nibbles to skip
+     * @return a new Nibbles without the first n leading nibbles
+     * @throws IndexOutOfBoundsException if
+     *         ({@code n < 0 || n > size})
+     */
+    public Nibbles drop(int n) {
```
bokoto000, line 149: I smell some Haskell-ish naming here. Would subNibbles(int from) sound better?

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
```diff
@@ -92,191 +84,148 @@ public boolean startsWith(Nibbles prefix) {
...
-    public Nibble remove(int index) {
-        return nibbles.remove(index);
+    /**
+     * Takes only the first n nibbles from the beginning.
+     * @param n the number of nibbles to take
+     * @return a new Nibbles limited to only the first n leading nibbles
+     * @throws IndexOutOfBoundsException if
+     *         ({@code n < 0 || n > size})
+     */
+    public Nibbles take(int n) {
```
bokoto000, line 160: Again smells like Haskell. Would subNibbles(int to) sound more Java-ish?

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
stoyan-lime: You can move it to a utility class for nibble operations (NibbleConverter or something similar?). You can also move the methods from NibblesToBytes there.

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
```diff
@@ -1,86 +1,78 @@
...
 
-    public Nibbles(Stream<Nibble> nibbles) {
-        this(nibbles.iterator());
+    public static Nibbles of(Iterator<Nibble> nibbles) {
+        return new Nibbles(IteratorUtils.toList(nibbles));
     }
 
-    public static Nibbles of(int... nibbles) {
-        return new Nibbles(Arrays.stream(nibbles).mapToObj(Nibble::fromInt));
+    Nibbles(List<Nibble> nibbles) {
```
stoyan-lime, line 59: You can make the constructor private and create a public `of(List)` method to keep the class consistent.

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
georg-getz: I'm fine with them throwing exceptions.

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
georg-getz: Why would mutability become difficult to switch to?

David-Petrov: Because someone would have to track the ownership anew, which is currently given by `copy`. Everywhere we now copy, the new owner would want to own the data and probably mutate it at some point. These mutations generate new instances now, which is the easy part to refactor (when switching to mutability of Nibbles) - make them mutate the old instance, but that would cause changes in previous owners too, since by removing `copy` we will have passed by reference only.

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
georg-getz: We decided that `BytesToNibbles` and the reverse one could be in a single utility class where it would also make sense to put this.

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
David-Petrov: We decided we'll leave `copy` as a shallow implementation returning `this` only for explicitness and traceability of ownership's sake.

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
David-Petrov: I thought about something similar to Java's `subList` naming, but honestly, I'd prefer the more explicit verbs. Stream API's `skip` and `limit` were also considered, but they seem kind of arbitrary to me (I don't like `limit` but I can see where the Java authors have come from). Initially, they were `take` and `skip`, but then `drop` clicked more with me 😄 
Feel free to discuss it with the team, ultimately I'm up for anything.

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
David-Petrov: Take a look and tell me whether you like it: a3f60bfc6ac55054ac4f19506131a7652eca6ca9
Resolve if considered finished.

PR #276, opened 2024-01-24
Trie initialization and storage

### LINKED ISSUE #245: Create trie initialization workflow
Create a workflow that ties together the following algorithms in a similar fashion to `open_database`:
https://github.com/smol-dot/smoldot/blob/377fdf027862b3d39f65b6883bea0edb96208f7b/full-node/src/lib.rs#L804-L809
- #242 
- #243
- #244

stoyan-lime:
Closes #245


src/main/java/com/limechain/trie/structure/InsertTrieBuilder.java
```diff
@@ -0,0 +1,135 @@
...
+    public final static int STATE_VERSION = 1; // TODO: Figure out where we'll fetch this state version from
+
+    private TrieStructure<NodeData> trieStructure;
+
+    public InsertTrieBuilder initializeTrieStructure(Map<String, String> mainStorage) {
+        TrieStructure<NodeData> trieStructure = buildTrieStructure(mainStorage);
+
+        List<TrieNodeIndex> nodeIndices = trieStructure.streamOrdered().toList();
+
+        for(TrieNodeIndex index : Lists.reverse(nodeIndices)) {
```
georg-getz: space after for

georg-getz (approved):

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -88,6 +88,30 @@ public Entry<T> node(Nibbles key) {
...
+                    .insert(nodeData);
+            case BranchNodeHandle<T> handle -> {
+                handle.setUserData(nodeData);
+                handle.convertToStorageNode();
+            }
+            case StorageNodeHandle<T> __ -> {
+                // We have a duplicate entry:
+                // a second value corresponding to an already inserted key from the genesis storage.
+                // NOTE: don't throw?
+                throw new IllegalStateException("Duplicate key in genesis storage (raw.top).");
```
David-Petrov: This line is the exact reason this was initially not a method of the trie structure, but left to manage caller-site. Anyhow, I'd at least change the message, if we want to bind this behaviour (throwing an exception) to the method. Although it might make sense to overwrite the value instead.
Generally, I was not sure whether we want this to be a method, although I also think it's locally inconvenient to repeat outside each time.

src/main/java/com/limechain/trie/structure/nibble/Nibbles.java
```diff
@@ -54,6 +54,10 @@ public static Nibbles of(Iterator<Nibble> nibbles) {
         return new Nibbles(IteratorUtils.toList(nibbles));
     }
 
+    public static Nibbles of(byte[] bytes) {
```
David-Petrov: I'd change this method name to `fromBytes`, since those `of` factory methods have the semantics of 'construct me from this representation of nibbles', but here an additional iteration logic (splitting bytes in half) occurs.

David-Petrov:
I think those newly added classes in the `trie.structure` package (namely `InsertTrieBuilder`, `NodeData`, and `TrieBuildException`) should be separated into a new, `database`-related package, since they're essentially not a part of the workings of the trie structure package, but instead related to our database persistence logic. Something like `trie.structure.database` as a new, but most importantly - separate, package could suffice for now. 

Same goes for the stuff inside the `trie.structure.node` package. Could go in `trie.structure.database.node`. 

Main point - I think it should be clear those are database serialisation / persistence related stuff.

src/main/java/com/limechain/client/FullNode.java
```diff
@@ -7,39 +7,26 @@
...
 import java.util.logging.Level;
-import java.util.stream.Collectors;
-import java.util.stream.IntStream;
 
 @Log
 public class FullNode implements HostNode {
+    private static final String TRIE_NODE_PREFIX = "tn:";
+    private static final String TRIE_NODE_STORAGE_PREFIX = "tns:";
+    private static final String TRIE_NODE_CHILD_PREFIX = "tnc:";
+
```
vikinatora, line 29: (no action required) nit: there are many different prefix conventions throught the code. We should probably settle for one at some point.
image

stoyan-lime: I think we can even move them to an enum or class and access them from there. I opened a separate issue for that: #278

src/main/java/com/limechain/trie/structure/TrieStructure.java
David-Petrov: Maybe also add a `@throws` in the JavaDoc to indicate the current behaviour to the API user.

David-Petrov (approved):

PR #282, opened 2024-01-25
Process genesis hashes and replace the existing ones

### LINKED ISSUE #214: Process genesis hashes and replace the existing hard-coded ones

ablax:
- [x] Process genesis hashes and replace the existing ones
- [x] Currently fallback to old method until trie merkle value calculation is fixed
Closes #214

### DIFF
--- a/src/main/java/com/limechain/constants/GenesisBlockHash.java
+++ b/src/main/java/com/limechain/constants/GenesisBlockHash.java
@@ -1,16 +1,62 @@
 package com.limechain.constants;
 
+import com.limechain.chain.ChainService;
+import com.limechain.network.protocol.warp.dto.BlockHeader;
+import com.limechain.network.protocol.warp.dto.HeaderDigest;
+import com.limechain.trie.structure.NodeHandle;
+import com.limechain.trie.structure.TrieStructure;
+import com.limechain.trie.structure.database.InsertTrieBuilder;
+import com.limechain.trie.structure.database.NodeData;
+import com.limechain.utils.HashUtils;
 import io.emeraldpay.polkaj.types.Hash256;
-import lombok.experimental.UtilityClass;
+import lombok.Getter;
+import org.apache.tomcat.util.buf.HexUtils;
+import org.springframework.stereotype.Component;
 
-@UtilityClass
+import java.math.BigInteger;
+import java.util.Map;
+
+@Component
+@Getter
 public class GenesisBlockHash {
-    public static final Hash256 POLKADOT =
-            Hash256.from("91b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3");
-    public static final Hash256 WESTEND =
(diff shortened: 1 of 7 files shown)


src/main/java/com/limechain/constants/GenesisBlockHash.java
```diff
@@ -1,16 +1,81 @@
...
+    }
+
+    byte[] getMerkleValue(TrieStructure<NodeData> trieStructure) {
+        //TODO: GENESIS STATE ROOT MERKLE VALUE MUST BE
+        // '0x29d0d972cd27cbc511e9589fcb7a4506d5eb6a9e8df205f00472e5ab354a4e17' for the polkadot chain
+        return trieStructure.getRootNode()
+                .map(NodeHandle::getUserData)
+                .map(NodeData::getMerkleValue)
+                .orElseThrow(() -> new RuntimeException("Root node merkle value not found"));
+    }
```
georg-getz, line 71: I don't understand this comment

src/main/java/com/limechain/trie/structure/database/InsertTrieBuilder.java
```diff
@@ -56,22 +56,25 @@ private TrieStructure<NodeData> buildTrieStructure(Map<String, String> mainStora
 
     private void updateMerkleValue(NodeHandle<NodeData> nodeHandle) {
         NodeData userData = nodeHandle.getUserData();
+//        if (userData == null) {
+//            return;
+//        }
```
georg-getz: is this intentional?

src/main/java/com/limechain/trie/structure/database/InsertTrieBuilder.java
```diff
@@ -11,18 +11,18 @@
...
+import lombok.Getter;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.stream.IntStream;
 
 public class InsertTrieBuilder {
```
David-Petrov, line 21: Leave a big todo above this class - it should be decomposed into two separate classes (along the semantics of "trie builder" and "database inserter" 😄), but this is soon to come in a major fix alongside the entire storage host api additions.

src/main/java/com/limechain/constants/GenesisBlockHash.java
ablax: Currently, the gerMerkleValue function is returning the wrong value, because of the wrong implementation in the merkle value calculation. The 0x29d0... is the expected value, for the genesis files root merkle value calculation. I added it in this todo because we were initially unsure which value to expect.

David-Petrov: We're calculating the wrong root hash - something is not right, but we'll figure it out later. That's the issue we discussed this morning on the daily.

src/main/java/com/limechain/trie/structure/database/InsertTrieBuilder.java
ablax: not rly

georg-getz (approved):

PR #291, opened 2024-01-31
Add full/warp sync configuration on fullnode

### LINKED ISSUE #219: Add configuration to start full node using either warp or full sync

ablax:
- [x] add full/warp sync option
Closes #219

### DIFF
--- a/src/main/java/com/limechain/cli/Cli.java
+++ b/src/main/java/com/limechain/cli/Cli.java
@@ -28,6 +28,7 @@ public class Cli {
     private static final String DB_RECREATE = "db-recreate";
     private static final String NODE_MODE = "node-mode";
     private static final String NO_LEGACY_PROTOCOLS = "no-legacy-protocols";
+    private static final String SYNC_MODE = "sync-mode";
 
     /**
      * Holds CLI options
@@ -60,8 +61,9 @@ public class Cli {
             //       what does running the node in NodeMode NONE mean?
             String nodeMode = cmd.getOptionValue(NODE_MODE, NodeRole.FULL.toString());
             boolean noLgacyProtocols = cmd.hasOption(NO_LEGACY_PROTOCOLS);
+            String syncMode = cmd.getOptionValue(SYNC_MODE, "full");
 
-            return new CliArguments(network, dbPath, dbRecreate, nodeKey, nodeMode, noLgacyProtocols);
+            return new CliArguments(network, dbPath, dbRecreate, nodeKey, nodeMode, noLgacyProtocols, syncMode);
         } catch (ParseException e) {
             formatter.printHelp("Specify the network name - " + String.join(", ", validChains), options);
             throw new CliArgsParseException("Failed to parse cli arguments", e);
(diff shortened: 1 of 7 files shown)


src/main/java/com/limechain/client/FullNode.java
```diff
@@ -69,7 +71,13 @@ public void start() {
...
         // Start syncing
         log.log(Level.INFO, "Node successfully connected to a peer! Sync can start!");
-        AppBean.getBean(WarpSyncMachine.class).start();
+
+        CliArguments args = AppBean.getBean(CliArguments.class);
+        if ("full".equals(args.syncMode().toLowerCase())) {
+            AppBean.getBean(FullSyncMachine.class).start();
+        } else {
+            AppBean.getBean(WarpSyncMachine.class).start();
+        }
```
vikinatora: (no action required) 

1. Maybe use a switch statement
2. I believe it'd be better to have some validation whether the sync mode is a valid string **before** finding peers.

ablax: Currently it fallbacks to warp if the input is invalid, but yes it would be troublesome if someone just mistyped full and it fallbacks to warp

src/main/java/com/limechain/cli/Cli.java
```diff
@@ -40,6 +43,15 @@ public Cli() {
         this.options = buildOptions();
     }
 
+    @NotNull
+    private static SyncMode parseSyncMode(CommandLine cmd) throws ParseException {
+        try {
+            return SyncMode.valueOf(cmd.getOptionValue(SYNC_MODE, "warp").toUpperCase());
+        } catch (IllegalArgumentException e) {
+            throw new ParseException("Invalid sync mode provided, valid values - WARP or FULL");
```
georg-getz: I don't like this exception being thrown for 2 reasons:

1. It's not ours. I would expect this to be coming from apache somewhere.
2. It's a checked exception meaning it forces us to handle it with `throws` clauses further coupling us with it.

If you are going to throw something here let it be our own unchecked exception.

src/main/java/com/limechain/cli/CliArguments.java
```diff
@@ -3,12 +3,13 @@
...
+ * @param network           the network
+ * @param dbPath            the DB path
+ * @param dbRecreate        flag for recreating the database for current network
+ * @param nodeKey           HEX for secret Ed25519 key
+ * @param noLegacyProtocols flag for disabling legacy protocols
+ * @param syncMode          the sync mode
  */
 public record CliArguments(String network, String dbPath, boolean dbRecreate, String nodeKey, String nodeRole,
-                           boolean noLegacyProtocols) {
+                           boolean noLegacyProtocols, com.limechain.sync.SyncMode syncMode) {
```
georg-getz: Any reason for this to not be an import?

src/main/java/com/limechain/client/FullNode.java
```diff
@@ -69,7 +71,14 @@ public void start() {
...
         // Start syncing
         log.log(Level.INFO, "Node successfully connected to a peer! Sync can start!");
-        AppBean.getBean(WarpSyncMachine.class).start();
+
+        CliArguments args = AppBean.getBean(CliArguments.class);
+        switch (args.syncMode()){
+            case FULL -> AppBean.getBean(FullSyncMachine.class).start();
+            case WARP -> AppBean.getBean(WarpSyncMachine.class).start();
+            default -> AppBean.getBean(WarpSyncMachine.class).start(); //This statement should be impossible to reach
+            //but it's here just in case
```
georg-getz: If the default should be unreachable I think it would be best if you make it throw an exception and have its message contain the `args.syncMode()` info.

georg-getz (approved):

PR #292, opened 2024-02-01
Fix genesis root hash

### LINKED ISSUE #289: Figure out (and fix) why we're calculating the wrong genesis root hash

David-Petrov:
In summary:
- fixes state root hash calculations;
- finishes up what @ablax had started with the `GenesisBlockHash`;
- some moving and repackaging;
- splitting of `InsertTrieBuilder` into a `TrieStructureFactory` and `InsertTrieBuilder` (the proper database related part of it);
- in-depth refactor and extension of `ChainSpec` (in-memory interpretation of the spec + JSON (de/)serializability + `implements Serializable` completely);
- fix for an unrelated issue regarding Spring beans (refer [13f935e](https://github.com/LimeChain/Fruzhin/pull/292/commits/13f935e557bc56721735b805f671b44c66056709))

Closes #289


src/main/java/com/limechain/chain/spec/ChainSpec.java
```diff
@@ -0,0 +1,25 @@
+package com.limechain.chain.spec;
+
+import lombok.Getter;
+
+// Serves as a wrapper around the raw JSON parsed chain spec
```
stoyan-lime: Use javadoc instead of comments.

src/main/java/com/limechain/chain/spec/Genesis.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+
+public class Genesis {
+    private final Map<ByteString, ByteString> top;
+    // TODO: Add child storage support
+    // private Map<ByteString, ...?> childrenDefault;
```
stoyan-lime: Remove commented field.

src/main/java/com/limechain/client/FullNode.java
```diff
@@ -39,17 +43,20 @@ public class FullNode implements HostNode {
...
-                    .build();
+            Genesis genesisStorage = loadGenesisStorage();
 
-            insertStorage(db, trie, InsertTrieBuilder.STATE_VERSION); //Todo: calculate state version
+            // TODO: Next up, manage to fetch the runtime version
+            StateVersion stateVersion = StateVersion.V0;
+
+            TrieStructure<NodeData> trie = TrieStructureFactory.buildFromKVPs(genesisStorage.getTop(), stateVersion);
+            List<InsertTrieNode> dbSerializedTrieNodes = new InsertTrieBuilder(trie).build();
+            insertStorage(db, dbSerializedTrieNodes, InsertTrieBuilder.STATE_VERSION); //Todo: calculate state version
```
stoyan-lime: The state version passed here is different than the one set above.

David-Petrov: True, fixed this.

src/main/java/com/limechain/chain/spec/Genesis.java
```diff
@@ -0,0 +1,46 @@
...
+public class Genesis {
+    private final Map<ByteString, ByteString> top;
+    // NOTE:
+    //  Whenever we want to add child storage support, it's going to be here
+    //  maybe something like... private Map<ByteString, ...?> childrenDefault;
+
+    Genesis(RawChainSpec rawChainSpec) {
+        Map<ByteString, ByteString> parsedTopStorage = new HashMap<>();
+
+        Function<String, ByteString> parser =
```
stoyan-lime, line 43: You can extract this in a private method.

src/main/java/com/limechain/chain/ChainService.java
```diff
@@ -50,29 +47,29 @@ protected void initialize(HostConfig hostConfig) {
...
+            try {
+                this.chainSpec = ChainSpec.newFromJSON(hostConfig.getGenesisPath());
+                log.log(Level.INFO, "✅️Loaded chain spec from JSON");
 
-            repository.save(DBConstants.GENESIS_KEY, this.getGenesis());
-            log.log(Level.FINE, "Saved chain spec to database");
-        } catch (IOException e) {
-            throw new ChainServiceInitializationException(e);
+                repository.save(DBConstants.GENESIS_KEY, this.chainSpec);
+                log.log(Level.FINE, "Saved chain spec to database");
```
georg-getz: I'm thinking that we almost always want to be avoiding anything below `INFO` level. The main idea of fine/finer/finest is to be used for logs that would spam the output and we only very occasionally would like to see being printed in order to debug, whereas messages like this one I think should be at least `INFO` level

src/test/java/com/limechain/chain/StateRootHashesIntegrationTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.Objects;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class StateRootHashesIntegrationTest {
+
+    // TODO: Refactor these hardcoded paths after we refactor to extract a logical ChainSpec module
```
georg-getz: It is actually not uncommon for tests to have hardcoded setups like this one. Given that it doesn't take that much space this TODO could be omitted.

src/main/java/com/limechain/chain/ChainService.java
David-Petrov: No probs, I've not changed that. 😄 Will now.

src/test/java/com/limechain/chain/StateRootHashesIntegrationTest.java
David-Petrov: I was bothered by the fact that tests share the same resources as the production environment... Ideally, I wanted to abstract the actual paths and bind them to the enums, but the problem is that they're configurable via `application.properties`, which isn't loaded at the time the enum instances are initialized. 

Anyways, to avoid the shared resources, perhaps cloning only the genesis data into separate files under `test/resources` and using that independent dump could be an option? Genesis data never changes, after all... and decoupling the test resources from the production resources seems like a fine idea. 

Anyway, I've removed the todo. If we decide it's worth it, we might go that extra bit.

georg-getz: The concern you're raising has a way bigger scope than just this test so the TODO would be better positioned more outwardly. What I mean is that maybe it's not a bad idea for us to have a TODO.md where have improvement suggestions such as this one that's maybe in root folder so it's more easily viewable by all. 

Anyway this would require hearing the other team members opinions so it's a discussion item for tomorrow given that we are past working hours.

georg-getz (approved):

stoyan-lime (approved):

PR #300, opened 2024-02-09
Load runtime version on runtime build

### LINKED ISSUE #290: Manage to obtain `runtime_version` (thus state version as its subfield) from genesis runtime
- [ x ] successfully build genesis runtime
- [  ] differentiate between custom sections and a `Core_version` runtime call

David-Petrov:
- add functionality to first try to parse the runtime version directly from the wasm blob's custom sections and if this fails (sections not present), fall back to calling `Core_version` on the built runtime. In this, fix the previously existing parsing from custom sections;

- move the essence of shared memory reading into the `Runtime` class (i.e. the method `getDataFromMemory`), was previously in `HostApi`

- use the state version field of the runtime version for the genesis trie

- add tests (actually, only a single working one, since the runtime builder tests turned out more complicated than expected)

Closes #290


src/test/java/com/limechain/runtime/RuntimeBuilderTest.java
```diff
@@ -0,0 +1,30 @@
...
+import java.io.InputStream;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mockStatic;
+
+class RuntimeBuilderTest {
+
+    @Test
+    @Disabled("Very lengthy, needs either refactoring in the code being tested or a lot of effort into mocking")
+    void buildsRuntimeWithVersionInCustomSection() throws IOException {
```
David-Petrov: We might as well delete this test, since it'll be a cumbersome task to write properly.
Main issue is the dependency on the spring context which we have in the very constructor of a `Runtime` instance.
A few approach options are possible:
Ideally, we'd want an end-to-end test with the real blob, so we could:
1) try to mock out the entire `HostApi` implementations;
2) try to stub with an empty implementation of all imports;
3) give up on the end to end / integration test and simply test what's relevant to the `RuntimeBuilder` (which is in sync with unit testing, but we lose the actual check that our logic works with a real blob)

The frustrating part is that the currently written logic actually works, but for now only inspected whilst debugging in production... some kind of automation and a regression test that we won't break it later would be very fit at this point.

src/main/java/com/limechain/runtime/WasmSectionUtils.java
```diff
@@ -0,0 +1,124 @@
...
+import com.limechain.runtime.version.RuntimeVersion;
+import com.limechain.utils.ByteArrayUtils;
+import io.emeraldpay.polkaj.scale.ScaleCodecReader;
+import lombok.experimental.UtilityClass;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Arrays;
+import java.util.Objects;
+
+// TODO: [Testability concern] Maybe make this class non-static?
```
georg-getz: But it already is non static, do you mean removing the `@UtilityClass` annotation?

src/main/java/com/limechain/runtime/WasmSectionUtils.java
```diff
@@ -0,0 +1,124 @@
...
+    private static final byte[] RUNTIME_APIS_SECTION_NAME = "runtime_apis".getBytes();
+
+    /**
+     * Parses the runtime version if both wasm custom sections ("runtime_apis" and "runtime_version") are present.
+     * @param wasmBinary the wasm blob
+     * @return the parsed RuntimeVersion if both custom sections are present, null otherwise
+     */
+    @Nullable
+    public RuntimeVersion parseRuntimeVersionFromCustomSections(byte[] wasmBinary) {
+        // byte value of \0asm concatenated with 0x1, 0x0, 0x0, 0x0 from smoldot runtime_version.rs#97
```
georg-getz: Not a good practice to point to a concrete line in another repo since it can easily be changed making this comment false.

src/main/java/com/limechain/runtime/WasmSectionUtils.java
```diff
@@ -0,0 +1,124 @@
...
+
+            // Read section size (as varint)
+            int sectionSize = 0;
+            int shift = 0;
+            byte byteRead;
+            do {
+                byteRead = wasmBytes[offset++];
+                sectionSize |= (byteRead & 0x7f) << shift;
+                shift += 7;
+            } while (byteRead < 0);
```
georg-getz, line 96: Can you add a small comment explaining this do while? For example where is the 7 coming from?

src/main/java/com/limechain/runtime/WasmSectionUtils.java
David-Petrov: Yes, exactly. The two opposing points are:
1) pro: static methods are more cumbersome to mock; mocking an instance is easier (in tests, which don't exist yet)
2) con: logically, it's a static utility class (by its purpose), so instantiating it everywhere might be considered "annoying boilerplate" or "unnecessary" since it doesn't hold any significant state.

src/main/java/com/limechain/runtime/version/ApiVersions.java
```diff
@@ -0,0 +1,66 @@
...
+        ScaleReader<ApiVersion> apiVersionReader = new ApiVersionReader();
+        List<ApiVersion> apiVersionsRaw = new LinkedList<>();
+
+        //TODO: Think about checking if the length is divisible by the fixed byte size of an encoded api version
+        while (reader.hasNext()) {
+            try {
+                apiVersionsRaw.add(reader.read(apiVersionReader));
+            } catch (RuntimeException e) {
+                break; // Presume end of iteration (all is ok), although other exceptions are possible
+            }
```
georg-getz: It's probably good to have a log here, if the exception is expected the log could be set to a lower logging level, but it's good to have some info about the exception.

src/main/java/com/limechain/runtime/WasmSectionUtils.java
David-Petrov: I agree, but I left this comment as it was since I was unable to find anything regarding that prefix in the spec / anywhere in an official source... I'm most probably missing something, but alternatively, we could link to a certain commit hash in the smoldot repo in order to not lose the motivation of this "magic prefix" 😄

src/main/java/com/limechain/runtime/WasmSectionUtils.java
David-Petrov: Again, that was magic to myself, too... I'm not sure where this comes from as I've not investigated it in depth, probably the reason I was hesitant to change it in the slightest 😄

src/test/java/com/limechain/runtime/RuntimeBuilderTest.java
georg-getz: Have you tried using the runtime that's in the examples folder of wasmer-java? It is very lightweight and only requires 2 function imports and a memory import.

src/main/java/com/limechain/runtime/WasmSectionUtils.java
georg-getz: I prefer leaving it in as changing code for ease of testing isn't always a good practice.

src/main/java/com/limechain/runtime/WasmSectionUtils.java
georg-getz: commit hash is a bit better for sure

src/test/java/com/limechain/runtime/RuntimeBuilderTest.java
David-Petrov: In a side discussion, we decided to scrape the test for now as out of scope. Added to `TODO.md` instead.

src/main/java/com/limechain/runtime/version/ApiVersions.java
David-Petrov: Actually, an exception was not expected per se, it's rather really an exceptional scenario, it was a miss on my part.
Take a look again if we're satisfied with this handling (javadoc improved as well): a037d61

georg-getz (approved):

ablax (approved):

stoyan-lime (approved):

PR #305, opened 2024-02-19
Var-int blocknumber scale

David-Petrov:
- extract the currently hardcoded size of 4 bytes into a const inside BlockHeader (not the best place)

- make a dedicated BlockNumberReader/Writer for handling VarUint64Reader/Writer in the concrete case of it being a block number

### DIFF
--- a/src/main/java/com/limechain/network/protocol/warp/dto/BlockHeader.java
+++ b/src/main/java/com/limechain/network/protocol/warp/dto/BlockHeader.java
@@ -1,21 +1,22 @@
 package com.limechain.network.protocol.warp.dto;
 
 import com.limechain.network.protocol.blockannounce.scale.BlockHeaderScaleWriter;
-import com.limechain.utils.scale.exceptions.ScaleEncodingException;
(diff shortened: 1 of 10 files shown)


src/main/java/com/limechain/network/protocol/warp/dto/BlockHeader.java
```diff
@@ -1,21 +1,22 @@
...
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
 import java.math.BigInteger;
 import java.util.Arrays;
 
 @Setter
 @Getter
 public class BlockHeader {
+    // TODO: Make this const configurable
+    public static final int BLOCK_NUMBER_SIZE = 4;
```
David-Petrov, line 17: I'm pretty sure this isn't where we want this const to reside, but temporarily, I found this the most sensible place.
See [the issue regarding this](https://github.com/orgs/LimeChain/projects/16?pane=issue&itemId=53479974).

georg-getz (approved):

ablax (approved):

PR #308, opened 2024-02-23
Add block trie accessor.

stoyan-lime:


src/main/java/com/limechain/storage/trie/TrieStorage.java
```diff
@@ -149,6 +149,7 @@ private TrieNodeData getNodeFromDb(TrieNodeData trieNode, byte[] key) {
         }
 
         int commonPrefix = ByteArrayUtils.commonPrefixLength(trieNode.getPartialKey(), key);
+        // TODO: is this right?
```
ablax: remove comment

src/main/java/com/limechain/runtime/hostapi/StorageHostFunctions.java
```diff
@@ -28,11 +30,11 @@
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 public class StorageHostFunctions {
     private final HostApi hostApi;
-    private final KVRepository<String, Object> repository;
+    private final BlockTrieAccessor repository;
 
     private StorageHostFunctions(final HostApi hostApi) {
         this.hostApi = hostApi;
-        this.repository = SyncedState.getInstance().getRepository();
+        this.repository = AccessorHolder.getInstance().getBlockTrieAccessor();
```
ablax, line 37: Do we assume that on every runtime initialization, the AccessHolder holds the trie accessor for the block required by the runtime?

src/main/java/com/limechain/storage/trie/TrieStorage.java
```diff
@@ -295,12 +296,23 @@ private void collectEntriesUpTo(TrieNodeData node, byte[] key, byte[] currentPat
...
 
         collectEntriesUpTo(childNode, Arrays.copyOfRange(key, 1 + commonPrefix, key.length), nextPath, entries);
     }
 
+    public List<StorageNode> loadChildren(byte[] parentMerkleValue) {
+        TrieNodeData parentNode = getTrieNodeFromMerkleValue(parentMerkleValue);
+        List<byte[]> childrenMerkleValues = parentNode.getChildrenMerkleValues();
+        List<StorageNode> result = new ArrayList<>();
+        for (byte[] childrenMerkleValue : childrenMerkleValues) {
+            TrieNodeData childNode = getTrieNodeFromMerkleValue(childrenMerkleValue);
```
ablax: childrenMerkleValue can and will be null in some cases.

georg-getz (approved):

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -382,7 +383,7 @@ public NodeHandle<T> nodeHandleAtIndex(@NotNull TrieNodeIndex nodeIndex) {
     }
 
     @NotNull
-    NodeHandle<T> nodeHandleAtIndexInner(int nodeIndex) {
+    public NodeHandle<T> nodeHandleAtIndexInner(int nodeIndex) {
```
David-Petrov: I can't find the reason this has been made public... maybe forgotten?

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -400,7 +401,7 @@ public Nibbles nodeFullKeyAtIndex(TrieNodeIndex nodeIndex) {
     }
 
     @NotNull
-    Nibbles nodeFullKeyAtIndexInner(int targetNodeIndex) {
+    public Nibbles nodeFullKeyAtIndexInner(int targetNodeIndex) {
```
David-Petrov: [Same as above] I can't find the reason this has been made public... maybe forgotten?

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -495,4 +496,52 @@ public <U> boolean structurallyEquals(TrieStructure<U> other) {
...
     }
+
+    public boolean removeNode(Nibbles key) {
+        return switch (existingNodeInner(key)) {
+            case ExistingNodeInnerResult.NotFound ignored -> false;
+            case ExistingNodeInnerResult.Found found -> removeNode(found.nodeIndex);
+        };
+    }
+
+    public boolean removeNode(int nodeIndex) {
```
David-Petrov: I would change this `nodeIndex` argument's type from `int` to `TrieNodeIndex`, since it's part of the public API.
As far as I could gather, it's singular usage outside the package allows for it (as was intended).

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -495,4 +496,52 @@ public <U> boolean structurallyEquals(TrieStructure<U> other) {
             }
         }
     }
+
+    public boolean removeNode(Nibbles key) {
```
David-Petrov: [Suggestion] This could become a method on the `NodeHandle` class, thus guaranteeing that it could only be called on existing nodes. Where I see a benefit in that (avoiding the uncertainty which necessitates the `boolean` return type), I might see why one could find it convoluted. Still, that's what I would've done, at least for consistency's sake (the whole idea of the entire `...Handle`s being namely that - mutating operations on nodes, so for me it semantically passes as a method there) :D

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -495,4 +496,52 @@ public <U> boolean structurallyEquals(TrieStructure<U> other) {
...
+        return switch (existingNodeInner(key)) {
+            case ExistingNodeInnerResult.NotFound ignored -> false;
+            case ExistingNodeInnerResult.Found found -> removeNode(found.nodeIndex);
+        };
+    }
+
+    public boolean removeNode(int nodeIndex) {
+        TrieNode<T> trieNode = getNodeAtIndexInner(nodeIndex);
+        TrieNode.Parent parent = trieNode.parent;
```
David-Petrov, line 516: Removing the root is a no-go? Is this intentional?

src/main/java/com/limechain/trie/BlockTrieAccessor.java
```diff
@@ -0,0 +1,197 @@
...
+import com.limechain.trie.structure.Entry;
+import com.limechain.trie.structure.NodeHandle;
+import com.limechain.trie.structure.TrieStructure;
+import com.limechain.trie.structure.Vacant;
+import com.limechain.trie.structure.database.NodeData;
+import com.limechain.trie.structure.nibble.Nibble;
+import com.limechain.trie.structure.nibble.Nibbles;
+import io.emeraldpay.polkaj.types.Hash256;
+import lombok.AllArgsConstructor;
+import org.apache.commons.lang3.NotImplementedException;
```
georg-getz: What's the argument for using this exception instead of a custom one? It does create a bit of coupling and this concrete exception has the method `getCode()` which should return the unimplemented code which we never set and will never use.

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -495,4 +496,52 @@ public <U> boolean structurallyEquals(TrieStructure<U> other) {
...
+
+        if (numberOfChildren == 0) {
+            parentNode.childrenIndices[parent.childIndexWithinParent().asInt()] = null;
+            nodes.remove(nodeIndex);
+        } else if (numberOfChildren == 1) {
+            parentNode.childrenIndices[parent.childIndexWithinParent().asInt()] = mergeParentIntoChild(trieNode);
+            nodes.remove(nodeIndex);
+        } else {
+            trieNode.hasStorageValue = false;
+            trieNode.userData = null;
```
David-Petrov, line 535: Does this behaviour technically correspond to what's in the method name?
I'd say we're not "deleting a node" per se here, rather "deleting the KV entry from this particular node". This confuses me a bit, although I trust you it's implementationally correct / necessary.

P.S. Is it possible this method gets called on a branch node, which already didn't have any storage value... thus rendering this method invocation a no-op (changing nothing)?

src/main/java/com/limechain/trie/AccessorHolder.java
```diff
@@ -0,0 +1,27 @@
...
+    private AccessorHolder() { }
+
+    public static AccessorHolder getInstance() {
+        if (instance == null) {
+            instance = new AccessorHolder();
+        }
+        return instance;
+    }
+
+    public BlockTrieAccessor setBlock(Hash256 blockHash) {
```
David-Petrov: [Nitpick] I'd rename this method to `setAtBlock` (or similar) to indicate that aside from the returned instance is a mutated `this`... or add a Javadoc. (I fear side effects)

src/main/java/com/limechain/trie/TrieStructureFactory.java
```diff
@@ -30,6 +30,7 @@ public class TrieStructureFactory {
      */
     public TrieStructure<NodeData> buildFromKVPs(Map<ByteString, ByteString> entries, StateVersion stateVersion) {
         TrieStructure<NodeData> trie = buildTrieStructure(entries);
+        // TODO: copy implementation and ignore already calculated merkleValues
```
David-Petrov: I suspect this TODO is outdated...?

src/main/java/com/limechain/runtime/hostapi/StorageHostFunctions.java
```diff
@@ -98,10 +100,10 @@ public List<ImportObject> buildFunctions() {
      * @param valuePointer a pointer-size containing the key.
      */
     public void extStorageSetVersion1(RuntimePointerSize keyPointer, RuntimePointerSize valuePointer) {
-        byte[] key = hostApi.getDataFromMemory(keyPointer);
+        Nibbles key = Nibbles.fromBytes(hostApi.getDataFromMemory(keyPointer));
```
David-Petrov, line 103: Have you verified this is the expected way to retrieve the key/value (I'm a bit suspicious of no SCALE being involved :D)? I know the spec is lacking in details here, so I'd suppose you've cross-referenced with other implementations?

src/main/java/com/limechain/runtime/hostapi/StorageHostFunctions.java
```diff
@@ -285,12 +287,23 @@ public RuntimePointerSize extStorageChangesRootVersion1(RuntimePointerSize paren
...
+        byte[] nextKey = repository.getNextKey(key)
+                .map(NibblesUtils::toBytesAppending)
+                .map(this::asByteArray)
+                .orElse(null);
 
         return hostApi.writeDataToMemory(scaleEncodedOption(nextKey));
     }
 
+    private byte[] asByteArray(List<Byte> bytes) {
```
David-Petrov, line 299: I could suggest [this approach](https://github.com/LimeChain/Fruzhin/blob/353de481b99e28490255b5c6d9a424f6203369ed/src/main/java/com/limechain/trie/dto/node/DecodedNode.java#L217) as an alternative to this helper method. Only as a possibility if you prefer it.

src/main/java/com/limechain/storage/KVRepository.java
```diff
@@ -44,7 +44,7 @@ public interface KVRepository<K, V> {
      * @param limit maximum keys to return
      * @return whether the delete operation was successful
      */
-    List<byte[]> findKeysByPrefix(String prefixSeek, int limit);
+    List<byte[]> findKeysByPrefix(K prefixSeek, int limit);
```
David-Petrov, line 47: Are we sure all generic `K`s implementing this interface would have a semantically valid notion of what a "prefix" is?
(Not saying it was previously better and I clearly see your use case, just that we could note this as a potential issue)

src/main/java/com/limechain/trie/BlockTrieAccessor.java
```diff
@@ -0,0 +1,197 @@
...
+                    .map(NodeHandle::getUserData)
+                    .map(NodeData::getMerkleValue)
+                    .map(closestMerkle -> trieStorage.entriesBetween(closestMerkle, key.toString()))
+                    .ifPresent(entries -> entries.forEach(e -> partialTrie.insertNode(e.key(), e.nodeData())));
+        }
+    }
+
+    private void loadChildren(Nibbles key) {
+        Entry<NodeData> entry = partialTrie.node(key);
+        if (entry instanceof NodeHandle<NodeData> nodeHandle) {
```
David-Petrov, line 92: For consistency (and further, my personal preference :D), we could change this `if.. instanceof ...` to a `switch`.

src/main/java/com/limechain/trie/BlockTrieAccessor.java
```diff
@@ -0,0 +1,197 @@
...
+        if (closestNode instanceof Vacant<NodeData> vacantNode) {
+            Optional.of(vacantNode)
+                    .map(Vacant::getClosestAncestorIndex)
+                    .map(partialTrie::nodeHandleAtIndex)
+                    .map(NodeHandle::getUserData)
+                    .map(NodeData::getMerkleValue)
+                    .map(closestMerkle -> trieStorage.entriesBetween(closestMerkle, key.toString()))
+                    .ifPresent(entries -> entries.forEach(e -> partialTrie.insertNode(e.key(), e.nodeData())));
+        }
+    }
```
ablax: We never load any nodes to the partialTrie and it is always empty, thus getClosestAncestorIndex returns no value.

src/main/java/com/limechain/runtime/hostapi/StorageHostFunctions.java
stoyan-lime: The suggested approach seems to traverse the array twice, doesn't it?

David-Petrov: Yes. Depends on whether we value code reduction over performance or not. Was just a suggestion, rather nitpicky. 😄

src/main/java/com/limechain/trie/BlockTrieAccessor.java
```diff
@@ -65,13 +64,25 @@ private void loadPathToKey(Nibbles key) {
...
-                    .ifPresent(entries -> entries.forEach(e -> partialTrie.insertNode(e.key(), e.nodeData())));
+                    .ifPresent(entries -> entries.forEach(
+                            storageNode -> partialTrie.insertNode(storageNode.key(), storageNode.nodeData()))
+                    );
         }
     }
 
+    private byte[] closestAncestorMerkleValue(NodeHandle<NodeData> nodeHandle) {
+        if (nodeHandle == null) {
+            return blockHash.getBytes();
```
ablax:
```suggestion
            return BlockState.getInstance().getHeader(blockHash).getStateRoot().getBytes();
```

ablax (approved):

src/main/java/com/limechain/storage/KVRepository.java
stoyan-lime: K should implement Serializable and/or Iterable interfaces. This is a large refactor which has been discussed and needs to happen in a separate task.

PR #312, opened 2024-03-06
Add  Zombienet tests, Prometheus server, CLI options

### LINKED ISSUE #226: Add test to showcase successful peer finding

### LINKED ISSUE #227: Add test to showcase successful block header verification(light client)

vikinatora:
This PR introduces several new features:
## 1. Zombienet tests to Fruzhin.

The following tests are added: 
  - Block header verification (closes #227)
  - Peer discovery (closes #226)

## 2. Prometheus metrics
Zombienet relies on Prometheus metrics to be emitted so that it can run tests successfully. That's the reason why I've added a simple Prometheus server configuration. In the next phase, these metrics can be extended to feature more metrics like in the other Hosts.

## 3. Substrate CLI parameters
Additionally, I had to add dummy placeholders for several cli arguments. Zombienet passes them to the Host when executing tests. There's no way to disable them currently. In the next phase(s), we should devote time to making sure we have the same cli interface as Substrate so that we have better compatability with Substrate tooling.


src/main/java/com/limechain/Main.java
```diff
@@ -41,10 +52,12 @@ public static void main(String[] args) {
 
         // Start the client
         // NOTE: This starts the beans the client would need - mutates the global context
+        prometheusServer.emitStartTime();
```
David-Petrov: This is plugged in between the comment and the line it relates to. I'd move it above or below.

David-Petrov (approved):

src/main/java/com/limechain/cli/Cli.java
```diff
@@ -155,6 +184,19 @@ private Options buildOptions() {
...
+        unsafeRpcExternal.setRequired(false);
+        noMdns.setRequired(false);
+        noTelemetry.setRequired(false);
+        prometheusExternal.setRequired(false);
+        prometheusPort.setRequired(false);
+        rpcPort.setRequired(false);
+        listenAddress.setRequired(false);
+        basePath.setRequired(false);
+
+
```
georg-getz: CEL

georg-getz (approved):

PR #318, opened 2024-03-25
Add child storage functionality

### LINKED ISSUE #216: Add child storage functionality

### LINKED ISSUE #217: Add child trie functionality

ablax:
- [x] Child trie functionality

Closes #216, #217


georg-getz (approved):

src/main/java/com/limechain/trie/TrieAccessor.java
```diff
@@ -0,0 +1,227 @@
...
+        throw new UnsupportedOperationException(TRANSACTIONS_NOT_SUPPORTED);
+
+    }
+
+    @Override
+    public void closeConnection() {
+        throw new UnsupportedOperationException(TRANSACTIONS_NOT_SUPPORTED);
+    }
+
+    public byte[] getMerkleRoot(StateVersion version) {
```
David-Petrov, line 220: I really don't like the state version being a public argument here, since this binds this class to always be used in a combination with a `Runtime` instance (which knows the state version). For now this concern might not require a given refactor, since this is a broader topic (hybrid state is also not handled, so we'll have to think about it); I'm just leaving this comment as a mental note somewhere.

src/main/java/com/limechain/trie/TrieAccessor.java
```diff
@@ -0,0 +1,227 @@
...
+        return new DeleteByPrefixResult(deleted, true);
+    }
+
+    public void persistAll() {
+        for (ChildTrieAccessor value : loadedChildTries.values()) value.persistAll();
+        loadedChildTries.clear();
+
+        if (partialTrie.isEmpty()) return;
+        lastRoot = getMerkleRoot(StateVersion.V1);
+        trieStorage.insertTrieStorage(partialTrie, StateVersion.V1);
```
David-Petrov, line 186: Those state versions will have to be refactored (won't be hardcoded to V1, but we'll have to figure out how to fetch and handle them). [Mental note, no changes required.]

src/main/java/com/limechain/runtime/hostapi/ChildStorageHostFunctions.java
```diff
@@ -257,36 +260,40 @@ public RuntimePointerSize extDefaultChildStorageGetVersion1(RuntimePointerSize c
...
     /**
      * Compute the child storage root.
      *
-     * @param version the state version
      * @return a pointer-size to a buffer containing the 256-bit Blake2 child storage root.
      */
-    public RuntimePointerSize extDefaultChildStorageRootVersion2(int version) {
-        // TODO: update to use state trie versions
-        return extDefaultChildStorageRootVersion1();
+    public RuntimePointerSize extDefaultChildStorageRootVersion1(RuntimePointerSize childStorageKeyPointer, StateVersion v0) {
```
David-Petrov: I'd remove the `Version1` suffix of this method name, since it handles both API endpoints.

David-Petrov (approved):

PR #361, opened 2024-04-29
chore!: update `Nabu`

David-Petrov:
# Description

- What does this PR do?

  - update dependency version
  - remove the `.jar` from VC and automate the build from GitHub

  Nabu was in very early stages of development when last included, so now we're catching up as a part of the improvements on our networking module.

  For the full context, bear in mind that the version of Nabu is the one from the commit directly after [this PR ](https://github.com/LimeChain/nabu/pull/19), so the context in the discussion there is relevant. As for the changes, nothing significant has been changed on the side of Fruzhin, aside from:
  - couple of new arguments as a result of changing interfaces in Nabu
  - couple of changes in `build.gradle.kts` regarding to the automation of this dependency resolution (and transitive dependency management).

## Checklist:

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -26,6 +26,7 @@ tasks.withType<JavaExec> {
 
 repositories {
     mavenCentral()
+    maven("https://jitpack.io")
 }
 
 dependencies {
@@ -63,6 +64,23 @@ dependencies {
     implementation("io.prometheus:prometheus-metrics-core:1.2.0")
     implementation("io.prometheus:prometheus-metrics-instrumentation-jvm:1.2.0")
     implementation("io.prometheus:prometheus-metrics-exporter-httpserver:1.0.0")
+
+    // NOTE:
+    //  We implicitly rely on Nabu's transitive dependency on Netty's public interfaces.
+    //  We could explicitly include Netty ourselves, but we prefer to make sure we use the same version as Nabu.
+
+    // NOTE:
+    //  For jitpack's syntax for GitHub version resolution, refer to: https://docs.jitpack.io/
+
+    // Nabu
+//    implementation("com.github.LimeChain:nabu:master-SNAPSHOT") // Uncomment for "most-recent on the master branch"
+    implementation("com.github.LimeChain:nabu:16c6586")
+
+    // Guava
+    implementation("com.google.guava:guava:33.1.0-jre")
+
+    // Apache commons
+    implementation("commons-io:commons-io:2.16.1")
 
 }
 
--- a/src/main/java/com/limechain/exception/NotImplementedException.java
+++ b/src/main/java/com/limechain/exception/NotImplementedException.java
@@ -1,0 +1,7 @@
+package com.limechain.exception;
+
+public class NotImplementedException extends RuntimeException {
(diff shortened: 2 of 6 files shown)


build.gradle.kts
```diff
@@ -64,6 +65,23 @@ dependencies {
...
+    //  We could explicitly include Netty ourselves, but we prefer to make sure we use the same version as Nabu.
+
+    // NOTE:
+    //  For jitpack's syntax for GitHub version resolution, refer to: https://docs.jitpack.io/
+
+    // Nabu
+//    implementation("com.github.LimeChain:nabu:master-SNAPSHOT") // Uncomment for "most-recent on the master branch"
+    implementation("com.github.LimeChain:nabu:16c6586")
+
+    // Guava
```
David-Petrov, line 79: Those two are nonetheless included as transitive from Nabu, but we don't use them in any way related to that re-exporting occurring, so I've made them explicit. Just pointing that out in case anyone's opinion wants to differ on that one 😃

build.gradle.kts
```diff
@@ -64,6 +65,23 @@ dependencies {
...
+    // NOTE:
+    //  We implicitly rely on Nabu's transitive dependency on Netty's public interfaces.
+    //  We could explicitly include Netty ourselves, but we prefer to make sure we use the same version as Nabu.
+
+    // NOTE:
+    //  For jitpack's syntax for GitHub version resolution, refer to: https://docs.jitpack.io/
+
+    // Nabu
+//    implementation("com.github.LimeChain:nabu:master-SNAPSHOT") // Uncomment for "most-recent on the master branch"
+    implementation("com.github.LimeChain:nabu:16c6586")
```
ablax, line 77: Don't we want the master one?

build.gradle.kts
David-Petrov, line 79: Those two are nonetheless included as transitive from Nabu, but we don't use them in any way related to that re-exporting occurring, so I've made them explicit. Just pointing that out in case anyone's opinion wants to differ on that one 😃

build.gradle.kts
```diff
@@ -64,6 +65,23 @@ dependencies {
...
+
+    // Guava
+    implementation("com.google.guava:guava:33.1.0-jre")
+
+    // Apache commons
+    implementation("commons-io:commons-io:2.16.1")
```
David-Petrov, line 83: Those two are nonetheless included as transitive from Nabu, but we don't use them in any way related to that re-exporting occurring, so I've made them explicit. Just pointing that out in case anyone's opinion wants to differ on that one 😃

build.gradle.kts
David-Petrov: Depends on how flaky we're willing to allow this dependency to be. In general, snapshots are useful for development (accounting for more rapid changes). In production however, I'd prefer to stick to very concrete versions (a commit hash, for example, à la `.lock` file 😄), since I deem dependency updating a very intentional and delicate process to be undertaken with much caution.

In our current situation however (us, managing our own fork), I guess we could go either way without much difference 😄

ablax (approved):

vikinatora (approved):

PR #362, opened 2024-04-30
feat(dot/sync): Load Trie from database, fix some host APIs

### LINKED ISSUE #352: Correctly load a block state
We are currently saving a valid block state but lack the of getting it and using it which would save a lot of syncing time.

ablax:
# Description

- What does this PR do?
- [x] Implement logic for loading the trie structure from the database
- [x] Fix ext_crypto_secp256k1_ecdsa_recover host api to unblock block execution
- [x] Fix the ext_storage_clear_prefix host api to unblock block execution
- [x] Fix the ext_storage_next_key host api to unblock block execution
    - The logic to be able to load the state trie from the database is required to be able to set a given block state before executing a block.
    - The other changes were required, because they were blocking the block execution and full sync

Fixes #352 

## Checklist:

### DIFF
--- a/src/main/java/com/limechain/storage/trie/TrieStorage.java
+++ b/src/main/java/com/limechain/storage/trie/TrieStorage.java
@@ -515,4 +515,55 @@ public class TrieStorage {
 
         db.save(key, storageValue);
     }
+
+    /**
+     * Loads the trie structure from the database starting from a given root.
+     *
+     * @param trieRoot The Merkle root of the trie to load.
+     * @return The reconstructed trie structure, or null if the root is not found.
+     */
+    public TrieStructure<NodeData> loadTrieStructure(byte[] trieRoot) {
+        if (trieRoot == null || trieRoot.length == 0) {
+            log.warning("Invalid or empty trie root provided.");
+            return null;
+        }
+
+        TrieNodeData rootNode = getTrieNodeFromMerkleValue(trieRoot);
+        if (rootNode == null) {
+            log.warning("No trie node found for the given root.");
+            return null;
+        }
+
+        TrieStructure<NodeData> trie = new TrieStructure<>();
+        loadSubTrie(trie, rootNode, trieRoot, Nibbles.EMPTY);
+        return trie;
+    }
+
+    /**
+     * Recursively loads the sub-trie from the database and adds it to the given trie structure.
+     *
+     * @param trie            The trie structure being constructed.
+     * @param currentNodeData The current node data being processed.
+     * @param merkleValue     The Merkle value of the current node.
+     * @param currentPath     The nibble path to the current node.
+     */
+    private void loadSubTrie(TrieStructure<NodeData> trie, TrieNodeData currentNodeData, byte[] merkleValue,
+                             Nibbles currentPath) {
+        trie.insertNode(currentPath,
+                new NodeData(currentNodeData.getValue() == null ? currentNodeData.getTrieRootRef() :
+        currentNodeData.getValue(), merkleValue));
+
+        // Recursively load children and construct the trie
+        List<byte[]> childrenMerkleValues = currentNodeData.getChildrenMerkleValues();
+        for (int i = 0; i < childrenMerkleValues.size(); i++) {
+            byte[] childMerkleValue = childrenMerkleValues.get(i);
+            if (childMerkleValue != null) {
+                TrieNodeData childNodeData = getTrieNodeFromMerkleValue(childMerkleValue);
+                if (childNodeData != null) {
+                    Nibbles childPath = currentPath.add(Nibble.fromInt(i)).addAll(childNodeData.getPartialKey());
+                    loadSubTrie(trie, childNodeData, childMerkleValue, childPath);
+                }
+            }
+        }
+    }
 }
--- a/src/main/java/com/limechain/storage/trie/TrieStorage.java
+++ b/src/main/java/com/limechain/storage/trie/TrieStorage.java
@@ -507,6 +507,7 @@ public class TrieStorage {
         List<byte[]> childrenMerkleValues = trieNode.childrenMerkleValues();
 
         TrieNodeData storageValue = new TrieNodeData(
+                trieNode.isBranch(),
                 trieNode.partialKeyNibbles(),
                 childrenMerkleValues,
                 trieNode.isReferenceValue() ? null : trieNode.storageValue(),
@@ -556,9 +557,12 @@ public class TrieStorage {
         // utilising our knowledge of which node has a storage value anywhere in our algorithm, which is a grand
         // coincidence.
 
-        trie.insertNode(currentPath,
-                new NodeData(currentNodeData.getValue() == null ? currentNodeData.getTrieRootRef() :
-        currentNodeData.getValue(), merkleValue));
+        if (currentNodeData.isBranchNode()) {
+            trie.insertBranch(currentPath, new NodeData(null, merkleValue));
+        } else {
+            byte[] value = currentNodeData.getValue() == null ? currentNodeData.getTrieRootRef() : currentNodeData.getValue();
+            trie.insertNode(currentPath, new NodeData(value, merkleValue));
+        }
 
         // Recursively load children and construct the trie
         List<byte[]> childrenMerkleValues = currentNodeData.getChildrenMerkleValues();
--- a/src/main/java/com/limechain/storage/trie/TrieStorage.java
+++ b/src/main/java/com/limechain/storage/trie/TrieStorage.java
@@ -549,6 +549,13 @@ public class TrieStorage {
      */
     private void loadSubTrie(TrieStructure<NodeData> trie, TrieNodeData currentNodeData, byte[] merkleValue,
                              Nibbles currentPath) {
+        // TODO Known issue:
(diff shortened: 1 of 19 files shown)


src/main/java/com/limechain/trie/BlockTrieAccessor.java
```diff
@@ -9,12 +9,17 @@
...
+     * @param stateRootHash the state root hash of the block whose trie is to be accessed
+     */
+    public BlockTrieAccessor(Hash256 stateRootHash) {
+        super(stateRootHash.getBytes());
 
         BlockState blockState = BlockState.getInstance();
-        if (blockState.isInitialized() && blockState.hasHeader(blockHash)) {
-            super.lastRoot = blockState.getHeader(blockHash).getStateRoot().getBytes();
+        if (blockState.isInitialized() && blockState.hasHeader(stateRootHash)) {
+            super.lastRoot = blockState.getHeader(stateRootHash).getStateRoot().getBytes();
```
David-Petrov: I think either the argument is misnamed (should remain block hash) or `blockState.getHeader` does something different from what's described (it should expect the block hash).

ablax: Indeed it is supposed to be blockHash, the confusion came from the superclass' constructor.

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -501,6 +506,37 @@ public boolean removeValueAtKey(Nibbles key) {
         };
     }
 
+    public int clearNodeValueRecursive(TrieNodeIndex nodeIndex, Long limit) {
+        return clearNodeValueRecursive(nodeIndex.getValue(), limit);
+    }
+
+    public int clearNodeValueRecursive(int nodeIndex, Long limit) {
```
David-Petrov: This shouldn't be public, the design premise being that the public interface only exposes access through `TrieNodeIndex`es and no raw `int`s.

src/main/java/com/limechain/trie/TrieAccessor.java
```diff
@@ -172,7 +150,31 @@ public void persistUpdates() {
 
     @Override
     public Optional<Nibbles> getNextKey(Nibbles key) {
-        return Optional.ofNullable(trieStorage.getNextKeyByMerkleValue(lastRoot, key));
+        NodeHandle<NodeData> rootHandle = initialTrie.getRootNode().orElse(null);
+        return findNextKey(rootHandle, key, Nibbles.EMPTY);
+    }
+
+    private Optional<Nibbles> findNextKey(NodeHandle<NodeData> node, Nibbles prefix, Nibbles currentPath) {
```
David-Petrov, line 156: Just a question: due to issues with in-memory diffs mismatching the on-disk stored state, this implementation moves the entire logic of finding the next key onto the mutated in-memory trie, right?

ablax: with the current implementation, yes

David-Petrov:
A nasty duplication (+ mismatch) of logic has sneaked in at some point. What `TrieStructure::clearNodeValue` is doing right now seems to have been originally meant to be a private subpart of `TrieStructure::removeValueAtKey`. The latter method is now only used in a single test case (thus factually deprecated), but it properly accounts for the conditions under which to apply `clearNodeValue`. Now however, `clearNodeValue` is being used from within the `TrieAccessor` with the incorrect presumptions of what it does - instead of clearing a storage value from the storage, this method actually brute force deletes the node (no matter whether it has a storage value). This has probably resulted from our rather mixed usage of the trie structure as a granular structural representation of the tree AND a high-level manager of the actual state storage (think, simple KVPs). To avoid this, I would:
1. Either move the notion of `removing a KVP from the storage` to the `StorageNodeHandle` class to avoid such misconceptions and be sure I'm only able to call that method on an actual storage node within the `TrieStructure`;
2. Or fix the conditions under which it's being called... but that would need to happen every time on the caller's side (that's a single usage for now, here: https://github.com/LimeChain/Fruzhin/blob/00d07ab5fd23b203646566be8ad0523a4e64446e/src/main/java/com/limechain/trie/TrieAccessor.java#L99). Proper condition would be checking for an instance of `StorageNodeHandle`, since anything else doesn't represent a node with an actual storage value.
3. Very pressingly change some names around and streamline the process.

Well, all of that, of course, depends on how much we want to invest in this approach before refactoring, so you might want to disregard this entire situation as a temporary mess and my comments as nitpicks :D

src/main/java/com/limechain/storage/trie/TrieStorage.java
```diff
@@ -515,4 +515,55 @@ private void insertTrieNodeStorage(InsertTrieNode trieNode,
...
+     * Recursively loads the sub-trie from the database and adds it to the given trie structure.
+     *
+     * @param trie            The trie structure being constructed.
+     * @param currentNodeData The current node data being processed.
+     * @param merkleValue     The Merkle value of the current node.
+     * @param currentPath     The nibble path to the current node.
+     */
+    private void loadSubTrie(TrieStructure<NodeData> trie, TrieNodeData currentNodeData, byte[] merkleValue,
+                             Nibbles currentPath) {
+        trie.insertNode(currentPath,
```
David-Petrov: Known issue: we're inserting *all nodes* from the DB as key-value-pairs, thus marking a lot of structural branch nodes (i.e. branch nodes with no storage value) as actually having a storage value.

Miraculously though, we're still able to calculate proper state root hashes since we're not actually utilising our knowledge of which node has a storage value anywhere in our algorithm, which is a grand coincidence.

P.S. On another note, inserting `currentNodeData.getTrieRootRef()` instead will always result in weird behaviours where an actual branch node coming from the DB would be inserted as a storage node with a NodeData(value = null; merkleValue = {correctMerkleValue})... a weird form of invalid state, fortunately not a blocker for block execution with our current implementation... but again, by a great coincidence :D

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -494,18 +500,42 @@ public <U> boolean structurallyEquals(TrieStructure<U> other) {
...
+            if (childrenIndex != null) {
+                deleteNodesRecursively(childrenIndex, limit, deleted);
+            }
+        }
+
+        parentNode.childrenIndices[parent.childIndexWithinParent().asInt()] = null;
+        nodes.remove(nodeIndex);
+    }
+
+    private boolean clearNodeValue(int nodeIndex) {
```
David-Petrov: I'd rather call this `deleteNode` or `removeNode` since that's the actual brute force removing of the node with the index.

src/main/java/com/limechain/trie/TrieAccessor.java
```diff
@@ -93,17 +90,11 @@ public Optional<byte[]> findMerkleValue(Nibbles key) {
...
     @Override
     public boolean delete(Nibbles key) {
-        Entry<NodeData> node = initialTrie.node(key);
-        if (!(node instanceof Vacant<NodeData>)) {
-            NodeHandle<NodeData> nodeHandle = node.asNodeHandle();
-            initialTrie.clearNodeValue(nodeHandle.getNodeIndex());
-            return true;
-        }
-        return false;
+        return initialTrie.deleteNodeAt(key);
```
David-Petrov: Whilst `deleteNodeAt` seems an appropriate name for what the method does, we're still missing the check here for whether its an actual storage node at this key. With the current implementation we're able to accidentally remove a branch node with no storage value if it's full key happens to coincide with the one requested.

This could either be delegated to the TrieStructure as a responsibility (i.e. an actual `clearStorageValue` method which does the check), but I'd prefer to leave that at the caller as to not introduce further responsibilities to the `TrieStructure` class.

georg-getz (approved):

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -500,13 +525,26 @@ public <U> boolean structurallyEquals(TrieStructure<U> other) {
...
+            }
+            case BranchNodeHandle<T> branchNodeHandle -> false;
+            case Vacant<T> vacant -> false;
         };
     }
 
+    public boolean deleteInternalNodeAt(Nibbles key) {
+        Entry<T> entry = node(key);
+        if(!(entry instanceof Vacant<T>)){
+            deleteStorageNodeAt(entry.asNodeHandle().rawNodeIndex);
```
David-Petrov: This method we could rename to simply `deleteNodeAt` since it deletes the node with no checks.

src/main/java/com/limechain/trie/structure/TrieStructure.java
```diff
@@ -91,6 +91,31 @@ public Entry<T> node(Nibbles key) {
...
     }
 
+    /**
+     * Insert a branch node at the given key.
+     *
+     * @param key      partial key in nibbles
+     * @param nodeData data to insert
+     * @implSpec if a storage value is already present for the given key, it will be overwritten
+     */
+    public void insertBranch(Nibbles key, T nodeData) {
```
David-Petrov, line 104: Let's add a comment above this public method that it's a temporary hacky solution in order to properly load the trie structure from the DB... and not a very polished part of the public interface of the class.

David-Petrov (approved):
This should do it for now! 👍

PR #407, opened 2024-05-16
Improve local development process.

Zurcusa:
Improve local development process.

# Description

- What does this PR do?
Remove "raw" from genesis field in westend-local-example.json and update README.md with correct polkadot node port.

So that we have a more error free local development guide for newcomers.

By changing the local chain spec json and updating the README.md

## Checklist:


vikinatora:
We have to at least mention the step where the chain spec format received from the node doesn't exactly match the one Fruzhin expects. Ideally, we should have a script that does all the manual steps.

georg-getz (approved):

Zurcusa:
When using the existing `westend-local-example.json` as a template chain spec for the local node I kept receiving request timeouts during the RequestFragmentsState warp sync. Using the fetched chain spec and editing it so it complies with the project's requirements seems to take care of the problem (see [this](https://github.com/LimeChain/Fruzhin/pull/407/commits/3afde7594c9383d64815d08ac91a09da5b4d58d3) commit). In my opinion it would help avoid such issues in the future. This is temporary solution until an automation script is created.

PR #413, opened 2024-05-21
refactor(block execution)!: showcase a MWE with explicit `Context` object

David-Petrov:
# Description

Decouple the Spring app from runtime invocations. This PR showcases a variant of this decoupling using an explicit `Context` object to pass to the runtime builder, in which object all necessary dependencies for the host API endpoints are contained, thus granting the caller the ability to define all interactions from the runtime to the outside world through the behaviour of these dependencies.

The main goal of this MWE is to execute the first block in Kusama without ever starting the Spring app (i.e. in isolation). In more practical terms: to rewrite the previously existing `BlockExecutorTest`, and the new variant can be found in [BlockExecutorIsolatedTest](https://github.com/LimeChain/Fruzhin/blob/5d313edcb5d1c3ef99d0ff188d2f9096e9f0f7a3/src/test/java/com/limechain/sync/fullsync/BlockExecutorIsolatedTest.java#L41). The diff between the two tests is the starting point for grasping all the changes. 

The bulk of refactoring is isolated in the package `com.limechain.runtime.research.hybrid`, where I've copied a lot of existing classes in an isolated environment and made changes on the copies, since applying changes in the already existing production classes quickly proved to be non-viable for the purpose of achieving a MWE.

Breakdown on the main changes within `com.limechain.runtime.research.hybrid`:
- `org.wasmer.Memory` and `FreeingBumpHeapAllocator` have been put behind interfaces;
- `TrieAccessor`s are now injected the `TrieStorage` dependency on construction;
- regarding the host api:
  - an `Endpoint` enumeration of all host api endpoints has been introduced, holding "just data", i.e. the signatures of the host api endpoints;
  - a `HostApiImpl` class is introduced to explicitly hold implementations of the endpoints and to simplify creation of `ImportObject`s exhaustively;
- `Runtime` and `RuntimeBuilder` have been renamed to `WasmRuntimeInstace` and `WasmRuntimeInstanceBuilder` (names are WIP) and now obtain an explicit `Context` object on construction;

Other notable changes (outside the `hybrid` package) which were necessary to achieve this decoupling:
- remove the `BlockState` field of `TrieStorage`, as it breaks separation of concerns;
- `TrieStorage` is no longer a static singleton instance, as it is an implicit member of the `Context` (through the `TrieAccessor`);
- [SIDE FIX / QUESTIONABLE] The runtime builder is now used for [the runtimeVersionV1 endpoint](https://github.com/LimeChain/Fruzhin/blob/21547d7005434d58ac867b4bca050219a99987fc/src/main/java/com/limechain/runtime/hostapi/MiscellaneousHostFunctions.java#L106) instead of manual instantiation. More details at this commit: 21547d7005434d58ac867b4bca050219a99987fc.

### DIFF
--- a/src/main/java/com/limechain/runtime/research/hybrid/WasmRuntimeInstance.java
+++ b/src/main/java/com/limechain/runtime/research/hybrid/WasmRuntimeInstance.java
@@ -1,0 +1,117 @@
+package com.limechain.runtime.research.hybrid;
+
+import com.limechain.runtime.hostapi.WasmExports;
+import com.limechain.runtime.hostapi.dto.RuntimePointerSize;
+import com.limechain.runtime.research.hybrid.allocator.Allocator;
+import com.limechain.runtime.research.hybrid.allocator.freeingbumpheap.FreeingBumpHeapAllocator;
+import com.limechain.runtime.research.hybrid.context.Context;
+import com.limechain.runtime.research.hybrid.hostapi.MinimalHostapiImpl;
+import com.limechain.runtime.research.hybrid.hostapi.SharedMemory;
+import com.limechain.runtime.research.hybrid.memory.WasmMemory;
+import com.limechain.runtime.version.RuntimeVersion;
+import lombok.extern.java.Log;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.wasmer.ImportObject;
+import org.wasmer.Imports;
+import org.wasmer.Instance;
+import org.wasmer.Memory;
+import org.wasmer.Module;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Level;
+
+
+@Log
+public class WasmRuntimeInstance {
+    Module module; // used to open/close all its instances
+
+    Instance instance;
+    Context context;
+
+    Allocator allocator;
+
+    private RuntimeVersion version;
+
+    // TODO: Reconsider whether `Context` should be a constructor argument or (dependency injected)
+    //                       OR an argument of `call()` (dependency parameterized)
+    //  The former necessitates the explicit passing of a context for every build of the runtime (we can't build the underlying module only - we always instantiate and tie the instance to the context)
+    //          - In more lightweight cases (i.e. only build to fetch the version), we'll have to explicitly provide a "partial" context; "partial" meaning with some of the services inside being null (or partially implemented) since we know they are not going to be used)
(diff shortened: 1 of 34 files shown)


src/main/java/com/limechain/runtime/research/hybrid/WasmRuntimeInstance.java
```diff
@@ -0,0 +1,117 @@
...
+    //  The latter necessitates the explicit re-instantiating of the underlying `org.wasmer.Instance` for each RuntimeAPI call (they do that in Gossamer), but allows for easy caching of the compiled wasm module (thus we can pass that around, e.g. in the BlockState) and instantiate with a context only when invocation is necessary
+    public WasmRuntimeInstance(Module module, Context context) {
+        this.context = context;
+
+        SharedMemory sharedMemory = new SharedMemory(null, null);
+        context.setSharedMemory(sharedMemory);
+
+        // getting imports
+        ImportObject.MemoryImport memory = new ImportObject.MemoryImport("env", 22, false);
+        List<ImportObject> functionImports = new MinimalHostapiImpl(context).getFunctionImports(); // NOTE: We could even inject the "Implementation" object from outside, if we'd want more granularity
```
David-Petrov, line 50: The `MinimalHostApiImpl` is named so just for the sake of the MWE. As you can see, it's not a dependency of `WasmRuntimeInstance`, so you can think of it as *the* host api implementation.

Although, as noted in this very line, we can change around abstractions: instead of passing the `Context` to the runtime instance's constructor, we can directly pass a `HostApiImpl`, thus delegating the entire implementation of the endpoints to the user (more granular user control). This would allow for defining separate minimal implementations (such as `MinimalHostapiImpl`), for example for isolated test scenarios (simply calling `Core_version` springs to mind), so instead of hand-picking which parts of `Context` you can ignore (i.e. set to `null`) as not needed, you can make that explicit by providing a really stripped implementation on your own. Of course, for production, a full implementation will be provided, e.g. from within the `WasmRuntimeInstanceBuilder`. 
This approach might even prove more reasonable when migrating to a on-disk trie.

src/test/java/com/limechain/sync/fullsync/BlockExecutorIsolatedTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+import java.io.IOException;
+import java.math.BigInteger;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@Log
+class BlockExecutorIsolatedTest {
```
David-Petrov, line 39: With this test not including booting the Spring app we lose the Spring logger, so logs are a bit uglier (and all in red for some reason). Think about whether this is an issue we'd want to investigate.

David-Petrov:
Closing this as already implemented in #456.

David-Petrov:
Closing this as already implemented in #456.

PR #446, opened 2024-05-27
refactor: dynamically load memory import

### LINKED ISSUE #445: Adapt runtime module memory import logic so that it's not statically defined
Currently, when initializing a `Runtime` instance we load the memory import statically. This causes the initialization to fail when runtimes change their memory limitations. To improve that we need to load the memory import from the provided wasm runtime files instead.

Zurcusa:
# Description
Switch to dynamically loading the memory import from the provided wasm files. Also removes the manual parsing logic for the custom sections.

- What does this PR do?
Loads the memory import dynamically from the provided runtime wasm. Further Improves the custom section parsing.

Defining a static memory import causes runtime init failure when the runtimes change their minimum memory limits.

They're implemented with the help of [this parser.](https://github.com/dylibso/chicory/tree/main/wasm). The memory import is extracted before the initialization of the runtime instance and loaded in its imports.

Fixes #445 
## Checklist:

### DIFF
--- a/src/main/java/com/limechain/runtime/Runtime.java
+++ b/src/main/java/com/limechain/runtime/Runtime.java
@@ -15,7 +15,6 @@ import org.wasmer.Memory;
 import org.wasmer.Module;
 
 import java.nio.ByteBuffer;
-import java.util.List;
 import java.util.logging.Level;
 
 import static com.limechain.runtime.RuntimeBuilder.getImports;
--- a/src/main/java/com/limechain/runtime/Runtime.java
+++ b/src/main/java/com/limechain/runtime/Runtime.java
@@ -8,11 +8,14 @@ import lombok.Getter;
 import lombok.extern.java.Log;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.wasmer.ImportObject;
+import org.wasmer.Imports;
 import org.wasmer.Instance;
 import org.wasmer.Memory;
 import org.wasmer.Module;
 
 import java.nio.ByteBuffer;
+import java.util.List;
 import java.util.logging.Level;
 
 import static com.limechain.runtime.RuntimeBuilder.getImports;
@@ -25,9 +28,9 @@ public class Runtime {
     private final int heapPages;
     private final FreeingBumpHeapAllocator allocator;
 
-    public Runtime(Module module, int heapPages) {
+    public Runtime(Module module, ImportObject.MemoryImport memoryImport, int heapPages) {
         this.heapPages = heapPages;
-        this.instance = module.instantiate(getImports(module, this));
+        this.instance = module.instantiate(Imports.from(getImports(memoryImport, this), module));
         this.allocator = new FreeingBumpHeapAllocator(getHeapBase());
     }
 
--- a/src/main/java/com/limechain/runtime/WasmSectionUtils.java
+++ b/src/main/java/com/limechain/runtime/WasmSectionUtils.java
@@ -22,6 +22,7 @@ import java.util.Objects;
 public class WasmSectionUtils {
     private static final String RUNTIME_VERSION_SECTION_NAME = "runtime_version";
     private static final String RUNTIME_APIS_SECTION_NAME = "runtime_apis";
+    private static final int DEFAULT_MINIMUM_MEMORY_PAGES = 24;
 
     /**
      * Parses the runtime version if both wasm custom sections ("runtime_apis" and "runtime_version") are present.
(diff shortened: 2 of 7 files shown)


src/main/java/com/limechain/runtime/Runtime.java
```diff
@@ -25,9 +27,9 @@ public class Runtime {
     private final int heapPages;
     private final FreeingBumpHeapAllocator allocator;
 
-    public Runtime(Module module, int heapPages) {
+    public Runtime(Module module, ImportObject.MemoryImport memoryImport, int heapPages) {
```
Zurcusa, line 30: I couldn't find a more elegant way to pass this as we need it during the `module.instantiate` method. The responsibility to deal with wasm binary stays in the `RuntimeBuilder`  class.

src/main/java/com/limechain/runtime/WasmSectionUtils.java
```diff
@@ -1,124 +1,82 @@
...
-    public WasmCustomSection findRuntimeVersionCustomSection(byte[] wasmBinary) {
-        return WasmSectionUtils.findCustomSection(wasmBinary, RUNTIME_VERSION_SECTION_NAME);
+    public ImportObject.MemoryImport parseMemoryImportFromBinary(byte[] wasmBinary) {
+        ImportSection importSection = toModuleWithSections(wasmBinary, SectionId.IMPORT).importSection();
+        MemoryImport parsedMemoryImport = (MemoryImport) importSection.stream()
+                .filter(i -> i.name().equals("memory"))
+                .findFirst()
+                .orElse(null);
+        return new ImportObject.MemoryImport("env",
+                parsedMemoryImport != null ? parsedMemoryImport.limits().initialPages() : 24,
```
Zurcusa: A default value of 24 for initial pages seems like an adequate solution considering the current requirements in various runtimes.

vikinatora: I'd suggest putting it in a private final field for others that won't know the context of the number

David-Petrov: Is failing to parse the memory import from the binary an expected scenario or an exceptional case? I feel like, intuitively, we expect the memory import to always be present in the binary (since it throws if there's a mismatch) and failing to retrieve it should be a rather exceptional case, hence bothering with a default value wouldn't be a concern.
Anyway, I might be wrong. Depends on the answer of the first question. 😄

Zurcusa: Imo it should always be present as per [doc](https://spec.polkadot.network/chap-state#sect-memory-management):

> This memory should be made available to the Polkadot Runtime for import under the symbol name memory.

My goal was to keep the current implementation to some extent. The previous flow always provided a memory import (regardless of mismatched with runtime wasm or not), which meant that in the case of a missing `memory` import it should startup (and maybe fail if we don't allocate it enough memory for a startup). Imo we should keep this default import, but maybe notify with a log that such an import was not found in the runtime wasm.

David-Petrov: To my understanding, the cited doc states that we must, in fact, import "a memory", i.e. we must provide an `ImportObject.MemoryImport` when instantiating the module. What I meant was whether we would ever have to "guess a default number for the memory pages" since it seemed to me that the runtime's expectations will always be stated in the wasm blob. And if we can parse that information, then that's our correct and expected way to go.
As for keeping the current implementation - if it doesn't match the use case, I wouldn't strive to keep it. 😄
Anyways, the main question was about intent and how to best express it.

src/main/java/com/limechain/runtime/Runtime.java
David-Petrov: This will be dealt with in the real PR, after the refactor.

src/main/java/com/limechain/runtime/hostapi/MiscellaneousHostFunctions.java
```diff
@@ -110,7 +111,8 @@ public RuntimePointerSize runtimeVersionV1(RuntimePointerSize data) {
         try {
             // TODO: Refactor using RuntimeBuilder... when we're sure it works properly
             Module module = new Module(wasmBlob);
-            Runtime newRuntime = new Runtime(module, RuntimeBuilder.DEFAULT_HEAP_PAGES);
+            ImportObject.MemoryImport memoryImport = WasmSectionUtils.parseMemoryImportFromBinary(wasmBlob);
```
David-Petrov, line 114: This hunk will also not be needed in the real PR, after the refactor.

src/main/java/com/limechain/runtime/WasmSectionUtils.java
Zurcusa: It seems that throwing an exception would be a correct way to handle this case. Cases where memory import is missing from the binary should not happen often(never?), thus if it does there would probably be consequences on all other hosts as well. No need to handle it specifically in out imeplementation.

PR #447, opened 2024-05-29
feat: improve syncing startegies

ablax:
# Description

- What does this PR do?
'Get rid' of the old SyncedState class.
Create a new class (SyncState) that holds only relevant information about the current sync state and keeps it in the database. Automatically switch to full sync after warp sync. (Currently starts from the first block, but after trie version 1 implementation will update it to start from the latest warp sync block).

Fixes #335

## Checklist:


README.md
```diff
@@ -27,12 +27,7 @@ cd Fruzhin
 ### Java Version
 
 Fruzhin only works
-with [Java 21 Coretto](https://docs.aws.amazon.com/corretto/latest/corretto-21-ug/downloads-list.html). Using any other
-version may cause "cannot calculate secret" errors when running the node:
-
-```
-org.bouncycastle.tls.crypto.TlsCryptoException: cannot calculate secret
-```
+with [Java 21](https://docs.aws.amazon.com/corretto/latest/corretto-21-ug/downloads-list.html).
```
Zurcusa: Was anything changed? Or was it always the case that Fruzhin worked with all java distros.

src/main/java/com/limechain/chain/lightsyncstate/scale/AuthorityReader.java
```diff
@@ -4,11 +4,10 @@
 import io.emeraldpay.polkaj.scale.ScaleCodecReader;
 import io.emeraldpay.polkaj.scale.ScaleReader;
 import io.emeraldpay.polkaj.scale.reader.UInt64Reader;
-import io.emeraldpay.polkaj.types.Hash256;
 
 public class AuthorityReader implements ScaleReader<Authority> {
     @Override
     public Authority read(ScaleCodecReader reader) {
-        return new Authority(new Hash256(reader.readUint256()), new UInt64Reader().read(reader));
+        return new Authority(reader.readUint256(), new UInt64Reader().read(reader));
```
Zurcusa, line 11: I think we could add the UInt64Reader into the ScaleCodecReader and avoid all the initialization across the project. For that we'd need to build a local snapshot of the lib. Wdyt?

src/main/java/com/limechain/client/FullNode.java
```diff
@@ -56,9 +56,15 @@ public void start() {
         log.log(Level.INFO, "Node successfully connected to a peer! Sync can start!");
 
         CliArguments args = AppBean.getBean(CliArguments.class);
+        WarpSyncMachine warpSyncMachine = AppBean.getBean(WarpSyncMachine.class);
```
Zurcusa, line 59: Have we ever considered constructor or method param injection?

src/main/java/com/limechain/rpc/methods/system/SystemRPCImpl.java
```diff
@@ -40,7 +40,7 @@ public class SystemRPCImpl {
     private final SystemInfo systemInfo;
     private final Network network;
     private final WarpSyncMachine warpSync;
-    private final SyncedState syncedState = SyncedState.getInstance();
+    private final SyncState syncState;
```
Zurcusa, line 43: I see here we use constructor injection. Imo we should create some kind of code conventions that we should follow regarding injection, using lombok etc.

src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
Zurcusa, line 1: Here we even have a mixture between ctor injection and AppBean.getBean

src/main/java/com/limechain/storage/DBRepository.java
```diff
@@ -30,12 +31,12 @@ public class DBRepository implements KVRepository<String, Object> {
      * Main DB folder
      */
     private static final String FOLDER_NAME = "db";
+    public static final Charset UTF_8 = StandardCharsets.UTF_8;
```
Zurcusa: You could do a static import to avoid this:
`import static java.nio.charset.StandardCharsets.UTF_8;`

README.md
ablax: With Java 17 we had this issue, but I noticed that it is not reproducible with the different java 21 versions after the java update.

src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
ablax: In the HostFunctions class the Rutime constructor injected entity is not a bean and can have a lot of instances.
Currently we do not support constructor injection to HostApiFunctions class

src/main/java/com/limechain/client/FullNode.java
ablax: We cannot exactly inject to the FullNode class through constructor/param. Also the FullNode and/or LightClient classes are currently initialized in the same method in which the spring application is initialized, we cannot exactly constructor/method inject them.

src/main/java/com/limechain/chain/lightsyncstate/scale/AuthorityReader.java
ablax: We use differente readers other that UInt64Reader as well. We are able to do that, but we might want to add others as well. Currently to avoid a lot of initializations of readers, we decided to make most of our readers Signleton with in constructor initialized param for reused readers, but this one is not converted yet.

src/main/java/com/limechain/client/FullNode.java
ablax: This comment is valid for the (SyncedState now renamed to) WarpSyncState class tho. But I intentionally left it out with AppBean, because I will modify it with the next incoming PRs as well.

Zurcusa: Wouldn't @David-Petrov 's refactor split the spring init and make it possible for such injections? I'm just guessing. Will check when he creates the PR.

ablax: Not sure, can't tell. But guess not, as his refactor's maint concern is Runtime and context for runtime invocations

src/main/java/com/limechain/sync/warpsync/WarpSyncState.java
```diff
@@ -147,7 +96,8 @@ public void syncBlockAnnounce(BlockAnnounceMessage blockAnnounceMessage) {
      * @param peerId        sender of the message
      */
     public synchronized void syncCommit(CommitMessage commitMessage, PeerId peerId) {
-        if (commitMessage.getVote().getBlockNumber().compareTo(lastFinalizedBlockNumber) <= 0) {
+        SyncState snycState = AppBean.getBean(SyncState.class);
```
David-Petrov: Sneaky `snyc` 😄

.github/dependabot.yml
```diff
@@ -4,3 +4,4 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    open-pull-requests-limit: 10
```
ablax, line 7: Used to be 5 by default, but I have seen it miss some more important version updates (for example last week's spring version update), because of already maxed out PRs, so I decided it would not be a bad idea to increase the limit

src/main/java/com/limechain/storage/DBRepository.java
```diff
@@ -191,13 +187,6 @@ public void commitTransaction() {
         //TODO: implement
     }
 
-    private byte[] getPrefixedKey(String key) {
-        return chainPrefix.concat(key).getBytes();
-    }
-    public byte[] removePrefixFromKey(byte[] key) {
-        return Arrays.copyOfRange(key, chainPrefix.length(), key.length);
-    }
-
```
ablax, line 200: Was removed, because it's absolutely obsolete to add chain name as prefix, because we are using different db directories for the different chains

.github/dependabot.yml
David-Petrov: I agree with that.

src/main/java/com/limechain/chain/lightsyncstate/scale/AuthorityReader.java
David-Petrov: Modifying external libraries comes with the cost of then maintaining forks. I find the public API of `io.emeraldplay.polkaj.scale` fairly apt, so I'd say it's our responsibility to define and properly structure whatever en/decoding logic we need. And that's a valid point, scale en/decoding it's currently not very unified across the entire project.

src/main/java/com/limechain/network/Network.java
```diff
@@ -148,7 +146,7 @@ private void initializeProtocols(ChainService chainService, GenesisBlockHash gen
                 )
         );
 
-        if (nodeRole == NodeRole.FULL) {
+        if (nodeRole == NodeRole.AUTHORING) {
```
David-Petrov, line 155: This seems like a significant change. `NodeRole.FULL` turned out incorrect or why is that?

src/main/java/com/limechain/rpc/config/CommonConfig.java
```diff
@@ -73,17 +70,22 @@ public Network network(ChainService chainService, HostConfig hostConfig, KVRepos
...
+        return new FullSyncMachine(network, syncState);
     }
 
     @Bean
     public GenesisBlockHash genesisBlockHash(ChainService chainService) {
         return new GenesisBlockHash(chainService);
     }
+
+    @Bean
+    public SyncState syncState(GenesisBlockHash genesisBlockHash, KVRepository<String, Object> repository) {
```
David-Petrov: [NIT] I would put this bean above all beans that are using it, just for perceived ease of reading. :D

src/main/java/com/limechain/config/SystemInfo.java
```diff
@@ -16,20 +18,22 @@
...
+    private final BigInteger highestBlock;
 
-    public SystemInfo(HostConfig hostConfig, Network network) {
-        this.role = network.getNodeRole().name();
+    public SystemInfo(HostConfig hostConfig, Network network, SyncState syncState) {
+        this.nodeRole = network.getNodeRole();
         this.chain = hostConfig.getChain();
         this.dbPath = hostConfig.getRocksDbPath();
         this.hostIdentity = network.getHost().getPeerId().toString();
+        this.highestBlock = syncState.getLastFinalizedBlockNumber();
```
David-Petrov, line 35: I'm a tad confused by this. Is this class only used to print in the logs on node start up? Because we're setting the highest known block in the constructor with no way to later update that information, so it's not going to be up-to-date as soon as we execute one more block?

Also, I would spare requiring an entire `SyncState` if we only need the `highestBlock` from it, but again, I might be confused as to what the intention for this class is.

src/main/java/com/limechain/network/protocol/blockannounce/messages/BlockAnnounceHandshakeBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+public class BlockAnnounceHandshakeBuilder {
+
+    /**
+     * Creates a Block Announce handshake based on the latest finalized Host state
+     *
+     * @return our Block Announce handshake
+     */
+    public BlockAnnounceHandshake getBlockAnnounceHandshake() {
+        SyncState syncState = AppBean.getBean(SyncState.class);
+        SystemInfo systemInfo = AppBean.getBean(SystemInfo.class);
```
David-Petrov: I fear the reliance on those beans :D Because as seen from the outside, a user of this class would never know of these implicit dependencies. I'd suggest we make those dependencies explicit, by (for example) declaring classes that depend on beans as beans themselves (`@Service`, `@Component`, etc.). 
I feel like `AppBean.getBean` should generally disappear, but without going into too much refactoring - maybe we attempt to make this class a service, or sth... wdyt?

src/main/java/com/limechain/storage/block/SyncState.java
```diff
@@ -0,0 +1,97 @@
...
+import com.limechain.network.protocol.warp.dto.BlockHeader;
+import com.limechain.storage.KVRepository;
+import io.emeraldpay.polkaj.types.Hash256;
+import lombok.Getter;
+import lombok.Setter;
+
+import java.math.BigInteger;
+
+@Getter
+public class SyncState {
```
David-Petrov, line 18: As far as I could gather, the main point of the PR is to extract this `SyncState` as holding common syncing stuff regardless of the method of syncing, whilst `WarpSyncState` manages syncing details regarding warp syncing specifically.

If so, then I'd just ask - isn't `authoritySet` warp sync specific?
Also, I'd add a JavaDoc to describe the role and intentions behind the classes. (I suspect this might be part of the unfinishedness of the PR, but still, had to mention :D)

src/test/java/com/limechain/config/SystemInfoTest.java
```diff
@@ -25,15 +28,20 @@ void SystemInfo_SetsRole_NoCliOption() {
         when(network.getHost()).thenReturn(host);
         when(network.getNodeRole()).thenReturn(NodeRole.FULL);
 
-        SystemInfo systemInfo = new SystemInfo(hostConfig, network);
+        SyncState syncState = mock(SyncState.class);
+        when(syncState.getLastFinalizedBlockNumber()).thenReturn(BigInteger.ZERO);
+
+        SystemInfo systemInfo = new SystemInfo(hostConfig, network, syncState);
```
David-Petrov, line 34: Here, too. I don't think we need the entire `syncState` to mock a block number if we just inject the block number into `systemInfo`.

src/main/java/com/limechain/rpc/config/CommonConfig.java
```diff
@@ -50,10 +49,8 @@ public HostConfig hostConfig(CliArguments cliArgs) {
 
     @Bean
     public KVRepository<String, Object> repository(HostConfig hostConfig) {
-        DBRepository repository = DBInitializer.initialize(hostConfig.getRocksDbPath(),
+        return DBInitializer.initialize(hostConfig.getRocksDbPath(),
                 hostConfig.getChain(), hostConfig.isDbRecreate());
-        SyncedState.getInstance().setRepository(repository);
```
David-Petrov, line 56: I find this deletion a huge win, good stuff!

src/main/java/com/limechain/storage/block/SyncState.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public static final String LAST_FINALIZED_BLOCK_NUMBER = "ss::lastFinalizedBlockNumber";
+    public static final String LAST_FINALIZED_BLOCK_HASH = "ss::lastFinalizedBlockHash";
+    public static final String AUTHORITY_SET = "ss::authoritySet";
+    public static final String LATEST_ROUND = "ss::latestRound";
+    public static final String STATE_ROOT = "ss::stateRoot";
+    public static final String SET_ID = "ss::setId";
```
David-Petrov: Those aren't public, are they?

src/main/java/com/limechain/storage/DBRepository.java
```diff
@@ -16,11 +16,12 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.logging.Level;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
```
David-Petrov, line 23: We sure we don't want ASCII? It'd be weird to allow weird character stuff like diacritics in DB keys :D

Additionally, that's the reason why I think `String` is not the perfect public API for holding opaque `byte[]` data, since `String`s hold a lot of representational logic (encodings). That's more of a general remark, but it came up here.

src/main/java/com/limechain/storage/DBRepository.java
David-Petrov: I agree, this seems to be the case given the logic in the constructor of this class.

src/main/java/com/limechain/sync/fullsync/FullSyncMachine.java
```diff
@@ -56,6 +59,10 @@ public void start() {
         //  unless explicitly set via some of the "update..." methods
         this.networkService.updateCurrentSelectedPeerWithNextBootnode();
 
+        Hash256 lastFinelizedStateRoot = syncState.getStateRoot(); //to replace line 67
+        // Todo: fetch state of the latest finalized block and start executing blocks from there
```
David-Petrov: [NIT] All caps `TODO`? 

I'd suggest we introduce some kind of conventions on different kinds of `TODO`s (NOTE, SAFETY, INTENTIONALITY, etc...), so that we all see glowy highlights the same way.

src/main/java/com/limechain/sync/warpsync/action/VerifyJustificationAction.java
```diff
@@ -1,42 +1,49 @@
...
 @Log
-public class VerifyJustificationState implements WarpSyncState {
-    private final SyncedState syncedState = SyncedState.getInstance();
+public class VerifyJustificationAction implements WarpSyncAction {
+    private final WarpSyncState warpSyncState = WarpSyncState.getInstance();
+    private final SyncState syncState;
     private Exception error;
 
+    public VerifyJustificationAction() {
+        this.syncState = AppBean.getBean(SyncState.class);
```
David-Petrov, line 23: I honestly dislike fetching this `syncState` as a bean here. Feels like a responsibility (or an attribute) of the `WarpSyncMachine` itself is independently delegated into its implementation detail class `VerifyJustificationAction` without any clear traceability as to who's responsible for setting up initial conditions, etc.

I'm aware this is out of scope, and I can't propose a solution without digging deeper, but I felt obliged to mention it :D Wdyt?

src/main/java/com/limechain/sync/warpsync/WarpSyncMachine.java
```diff
@@ -3,107 +3,109 @@
...
 import java.util.Comparator;
+import java.util.List;
 import java.util.PriorityQueue;
 import java.util.Queue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.logging.Level;
 
 @Log
+@Getter
```
David-Petrov, line 29: You sure we really want public getters and setters for all fields?

README.md
```diff
@@ -27,12 +27,7 @@ cd Fruzhin
 ### Java Version
 
 Fruzhin only works
-with [Java 21 Coretto](https://docs.aws.amazon.com/corretto/latest/corretto-21-ug/downloads-list.html). Using any other
-version may cause "cannot calculate secret" errors when running the node:
-
-```
-org.bouncycastle.tls.crypto.TlsCryptoException: cannot calculate secret
-```
```
David-Petrov, line 35: You've removed the reason for why we only work with Coretto, but the README still says `Frzhin only works with...`. 

From what I've gathered from the discussions (can't find where now...) this issue no longer persists, right? So, do we still need Coretto only, or do we want to simply require "Java 21"?

src/test/java/com/limechain/network/protocol/sync/SyncTest.java
```diff
@@ -57,6 +57,7 @@ public void stopNode() {
     }
 
     @Test
+    @Disabled("Integration test")
```
David-Petrov: At this point, I'd put `@Disabled` on the whole class. These tests are heavily reliant on real-world networking, so they don't even have place here (can't be run in isolation), thus are flaky. Maybe we should think of how we want to approach integration tests?

src/test/java/com/limechain/sync/warpsync/action/RuntimeDownloadStateTest.java
```diff
@@ -24,11 +26,14 @@
 import static org.mockito.Mockito.when;
 
 @ExtendWith(MockitoExtension.class)
+@Disabled
```
David-Petrov: Why? Couldn't refactor, tests not needed?

src/test/java/com/limechain/sync/warpsync/WarpSyncActionTest.java
```diff
@@ -37,9 +38,10 @@
 
 @SuppressWarnings("unused")
 @ExtendWith(MockitoExtension.class)
-class SyncedStateTest {
+@Disabled
```
David-Petrov: Same here, why that?

David-Petrov (commented):
Overall a pretty decent removal of the old `SyncedState` class. With that being the main goal, it's achieved. Does this, however, conclude the whole task of improving the syncing strategy?

src/main/java/com/limechain/network/Network.java
ablax: We are not processing transaction messages as full node at all. But we will as an authoring node when producing blocks.

src/main/java/com/limechain/config/SystemInfo.java
ablax: It is only the node status on start up, we don't need to update the highest block later during the time the host is running.

src/main/java/com/limechain/storage/block/SyncState.java
ablax: I think it will be relevant for the block production as well.

src/main/java/com/limechain/storage/block/SyncState.java
ablax: I might move them to DBConstants class.

README.md
ablax: we want to simply require java 21

David-Petrov: Then I'd reword the "only works with" part; or remove the specific link to the Coretto distro.

src/main/java/com/limechain/chain/lightsyncstate/scale/AuthorityReader.java
Zurcusa: This might be an idea for the "tech debt" tickets I mentioned in the daily today. I'll keep a list of links to threads / chats that could be an idea to look at. I'll close the thread for the time being.

src/main/java/com/limechain/rpc/methods/system/SystemRPCImpl.java
Zurcusa: Did we agree on anything here?

David-Petrov: I'm not sure whether you mean to raise a particular concern against constructor injection.

Other than that, I'm on board with the code conventions part. I could also repeat myself in this reply with what I meant [here](https://github.com/LimeChain/Fruzhin/pull/456#discussion_r1624238868).

Zurcusa: Nevermind my last comment. I completely misread the changes... :)

src/main/java/com/limechain/config/SystemInfo.java
David-Petrov: How about clarifying that in the JavaDoc?

Actually, if this is the case, I don't see the reason this is a spring bean. It has essentially only one usage place - logging the system info on node start up.

Is there anything I'm missing, like a future consideration of sorts, or do you just deem it out of scope to clean up for now?

src/main/java/com/limechain/network/protocol/blockannounce/messages/BlockAnnounceHandshakeBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+
+        Hash256 genesisBlockHash = syncState.getGenesisBlockHash();
+        Hash256 lastFinalizedBlockHash = syncState.getLastFinalizedBlockHash();
+        BigInteger lastFinalizedBlockNumber = syncState.getLastFinalizedBlockNumber();
+
+        NodeRole nodeRole = systemInfo.getNodeRole();
```
David-Petrov: We generally use `HostConfig` instead of `SystemInfo` for the purpose of fetching the node role. 
In the spirit of getting rid of `SystemInfo` as a bean, I'd change that around here, too.

Edit: Depends on what `SystemInfo`'s future will look like. It's a bit confused as of its purpose and identity right now: [ref](https://github.com/LimeChain/Fruzhin/pull/447#discussion_r1629087060) 😄

src/main/java/com/limechain/config/SystemInfo.java
David-Petrov: Nevermind. I see `SystemInfo` being used in `SystemRPCImpl`. 

On the other side, the `highestBlock` feels injected there just for the sake of getting logged on node start-up. 

I don't like this class; this feels like more than one responsibility. It did confuse me 😅

David-Petrov (approved):
After a discussion, we've agreed that some imperfections discussed in this PR will be addressed in a follow-up PR. Approving for now to unblock the work with the core of essential changes.

Issue #376, opened 2024-05-13
Decouple Spring and Polkadot Runtime


nistanimirov:

nistanimirov:
Closed with the merge of PR #456

Issue #355, opened 2024-04-29
Block execution research


vikinatora:
Dedicate time to researching how to approach improving the existing block execution.

David-Petrov:
- [ ] [LocalStorage](https://github.com/LimeChain/Fruzhin/blob/e068f0a5cb2dd81e207119dc3673bb175eba85c3/src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java#L70) must be [fork-aware](https://spec.polkadot.network/chap-host-api#defn-offchain-local-storage), which ours is not. Aside from extracting this dependency, we must consider where the use site is going to be regarding fork awareness.
Gossamer solves this by instantiating a as a new in-memory DB, but I'm not sure when / how those get persisted if at all. Not that high priority for now as it's not been a blocker. 

- [ ] Take time go properly handle `per-node` state-version... or at least streamline hanging hardcoded `StateVersion`s into parameters. [Murad]

- [ ] Migrate trie from in-memory to on-disk. [Yordan to research & work on + support from David & Murad]

- [ ] After executing a block, the `BlockTrieAccessor` simply dumps all the trie nodes in the database (i.e. to even retrieve it from the DB later, we must at least know the root hash... to identify the trie) and the information about what root we've managed to calculate is lost (i.e. we have no link saying "this is the root hash for this block"). Has to do rather with the `BlockState` and block finalisation / handling of forks; BUT I've stumbled upon this being an issue whilst trying to write the assertion for the `kusamaFirstBlock` test case...: after we've executed the block, we can't actually "see what the result of the execution was", i.e. "which is the hash that had been calculated"... so my temporary workaround is to simply recalculate it again from the last state of the in-memory trie... which is not that cool. 😄 [David & Murad to sync]

nistanimirov:
Closing this Epic task, as all other sub-tasks were completed.

Issue #439, opened 2024-05-27
Identify the trie by root hash


nistanimirov:
> - [ ] After executing a block, the `BlockTrieAccessor` simply dumps all the trie nodes in the database (i.e. to even retrieve it from the DB later, we must at least know the root hash... to identify the trie) and the information about what root we've managed to calculate is lost (i.e. we have no link saying "this is the root hash for this block"). Has to do rather with the `BlockState` and block finalisation / handling of forks; BUT I've stumbled upon this being an issue whilst trying to write the assertion for the `kusamaFirstBlock` test case...: after we've executed the block, we can't actually "see what the result of the execution was", i.e. "which is the hash that had been calculated"... so my temporary workaround is to simply recalculate it again from the last state of the in-memory trie... which is not that cool. 😄 [David & Murad to sync] 

 _Originally posted by @David-Petrov in [#355](https://github.com/LimeChain/Fruzhin/issues/355#issuecomment-2109743987)_

PR #456, opened 2024-05-31
refactor(runtime)!: add runtime execution context

### LINKED ISSUE #375: Abolish all Singletons related to the HostAPI

David-Petrov:
# Description

A rather big refactor striving to make runtime invocations sandboxable, i.e. decouple them from the spring app context (dependencies on its beans) and any dependencies on global singletons. We do this by introducing an explicit `Context` object to manage those dependencies. It gets injected into the runtime instance on construction, thus enabling us to create instances with different execution contexts.

## Detailed description
A detailed description can be found in each of the commit's messages:
* refactor(TrieStorage)!: remove `BlockState` dependency

    The `TrieStorage` class should serve as an abstraction over the DB (`KVRepository db`) for trie-related queries (thus handling finalised blocks only) and must not depend on `BlockState` related logic (which generally manages unfinalised blocks also).

* refactor(TrieStorage)!: remove singleton
    
    `TrieStorage` will not have a static singleton, will be a bean in production instead, and otherwise independently instantiable with the `db` dependency now in its constructor rather than in an `initialize()` method.

* refactor(runtimeVersionV1): use the runtime builder instead of manually building the instance
    
    - additionally, change around the scale encoding of the `RuntimeVersion`
    
    The main benefit is we can take advantage of the wasm custom sections in newer runtimes instead of always calling `Core_version`. Although, for this particular hostapi endpoint, it could be more ergonomic to not tie ourselves to the builder... anyhow, did it.

* refactor!: add runtime execution context

  * `StateRpcImpl`: fix a bug in `stateGetReadProof` - block trie accessor must be initialized with state root for the block instead of the block hash
  
  * `runtime.allocator` and `runtime.memory` packages: abstract the `FreeingBumpHeapAllocator` behind an `Allocator` interface. Also, abstract the `org.wasmer.Memory` behind a `Memory` interface, with which the `Allocator` operates.
  
  * `runtime.hostapi` package:
    - an `Endpoint` enum was introduced to hold the signatures for all Host API endpoints. Think of this class as containing "just data" - minimally necessary to define what a "Host API endpoint" is.
    - an abstract `HostApi` class was introduced to define what an "implementation" of these endpoints must consist of. Basically, it requires an execution context be passed in order to even construct an implementation; and forces child classes (actual implementations) to provide a total list of import objects for every Endpoint.
    - `DefaultHostApi` is the default production implementation, which has been decomposed into the previously known `...HostFunctions` classes (as they're partitioned in the spec). All of those classes have been minimally changed just to adapt them to the new concept (no actual endpoint implementations have been severely changed within them). As a notable novelty, a `PartialHostApi` interface was introduced for ease during development and migration, but I find it apt and decided it's a nice addition. It's basic idea was similar to `HostApi`'s, but with the crucial difference that it explicitly allows implementations to be partial (i.e. not for all endpoints). For now, all its implementors are the parts of the `DefaultHostApi`, so it's simply an implementation detail of the `DefaultHostApi`, but could easily be promoted to the public interface of runtime instantiation.
    - Notable changes amongst the `...HostFunctions` classes:
      - Each of them now accepts the minimal necessary dependencies on construction (which they receive from `DefaultHostApi`'s `Context`) instead on relying on global static singletons or Spring app beans.
      - Writing/Reading shared memory is no longer a responsibility of `Runtime`, but has been extracted into a separate `SharedMemory` class doing only that.
      - Their constructors are intentionally made package protected (to emphasize their current encapsulation as an implementation detail of `DefaultHostApi`)
      - `OffchainHostFunctions` now operates with `OffchainStorages`, which will be described below (see `storage.offchain` package).
      - `StorageHostFunctions` experienced a little refactor in its `scaleEncodedOption` helper.
  * `runtime.Context` class: introduced (roughly) to:
    - hold the necessary dependencies from the outer world;
    - hold the `SharedMemory` which is an internal component of the runtime instance, but exposes "public" functionality (public to implementors of the host API endpoints)
    - cache the runtime version (and potentially other stuff in the future, which does not exactly belong as a field in `Runtime`)
    It is intended as an implementation detail of the `Runtime` class (not directly accessible to the outer world). It does have public getters since the current package decomposition inside `runtime` is more granular than java's opinion on nesting folders and visibility within nested packages, but it's not intended to be visible outside the `runtime` package. The only exception to this "intention" is the third overload of `RuntimeBuilder::buildRuntime` which for now (and that's not very nice) publicly exposes the context for the very eventual and rare case that someone would want to provide a custom implementation of the Host API (for which they'll need things like `SharedMemory`, which is internally constructed by `RuntimeBuilder` since it depends on the underlying `org.wasmer.Instance` for the memory export). That's a bit of mixture of intent, but I won't dig any deeper into that for now (not a problem).
  * `runtime.Runtime` class: Intended as a wrapper around the underlying `org.wasmer.Instance`, with added execution context (and, for now, a reference to the `org.wasmer.Module` it originated from... but I'm not sure that's necessary since it's only used in `.close()`). Anyways, it doesn't handle memory operations anymore, and is rather empty for now, but this is the place where the public interface for interactions with the instance would go (methods like `.executeBlock(Block block)` (in future), `getVersion()`, etc.).
    Note that its constructor is package private, since constructing it is a concern of:
  * `runtime.RuntimeBuilder` class: it's actually more of a `RuntimeFactory` in standard Java terms, but essentially it's the public constructor for runtime instances. Source code should be self-explanatory here.
  * `storage.offchain` package: the previous `OffchainStore` has been abstracted to an implementor of the new `BasicStorage` interface, which was introduced to more simply represent the necessary for `OffchainHostFunctions` dependencies, which reside collected inside `OffchainStorages`. That's rather a leftover from my initial prototyping, but I like the simplicity this introduces in the supply chain through the `Context`.
    - `compareAndSet` method was moved as default in the interface.
  * package `trie`:
    - `.*TrieAccessor`: `trieStorage` was made an explicit dependency injected on construction
    - `AccessorHolder`: obtaining the `TrieStorage` from as spring bean has been moved here (scope boundary for my refactor).
    
  ---- TESTS ----

  * `BlockExecutorTest`: a showcase of explicitly configured runtime instance executing a block in isolation. We want more of those.
    - `InMemoryDB` introduced as a light testing repository
  * `RuntimeWasmTest`: although disabled, rises an interesting question (bring up a separate discussion in the PR)
  * all other tests have been refactored to adapt to changes, nothing noteworthy there.
  
  ------

Anything not explicitly mentioned in the above list is trivial (or I've missed to mention it, so feel free to raise any noticed oddities).

Fixes #375, #376, and the first big half of #355 

---

For traceability's sake: this PR is the realisation of the POC showcased in #413.

## Checklist:


src/main/java/com/limechain/runtime/Runtime.java
```diff
@@ -1,34 +1,43 @@
...
-    private final Instance instance;
-    private final int heapPages;
-    private final FreeingBumpHeapAllocator allocator;
-
-    public Runtime(Module module, int heapPages) {
-        this.heapPages = heapPages;
-        this.instance = module.instantiate(getImports(module, this));
-        this.allocator = new FreeingBumpHeapAllocator(getHeapBase());
+    // TODO: Figure out whether we'll actually need this field
+    Module module; // used to open/close all of its own instances
```
David-Petrov, line 18: The `Module` field is left in here just for the sake of being closed in the `.close()` method. Initially, no such logic existed, but that's what I gathered from Gossamer; and since our `BlockTree` and its underlying `HashToMap` are modeled after Gossamer, I figured this might turn out to be what we want (closing the module as well). 

The main thing to consider is what this method actually does in `wasmer-java`, since there are significant differences in the public API and naming conventions between our `wasmer-java` and Gossamer's `wasmtime`. I haven't figured that out yet, so it's a TODO.

src/main/java/com/limechain/runtime/Runtime.java
```diff
@@ -1,34 +1,43 @@
...
+    //      "partial" meaning with some of the services inside being null (or partially implemented)
+    //      since we know they are not going to be used
+    //  The latter necessitates the explicit re-instantiating of the underlying `org.wasmer.Instance`
+    //      for each RuntimeAPI call (they do that in Gossamer),
+    //      but allows for easy caching of the compiled wasm module (thus we can pass that around, e.g. in the BlockState)
+    //      and instantiate with a context only when invocation is necessary.
+    //      This might prove more fitting, for example,
+    //      when executing consecutive blocks (between which there's no runtime updates)
+    //      without rebuilding the module and by only changing the context...
+    Runtime(Module module, Context context, Instance instance) {
```
David-Petrov, line 36: That's a rather significant one. It's a core design decision. I'm not sure, however, what's most fit for our needs, so for now we go the easy way - we need both the `byte[]` source and `Context` for instantiation; so a `Runtime` instance is bound with it's context for its entire lifetime. This would mean that even if there were no runtime updates between blocks, we'd have to re-instantiate the entire module in case we need a change in the context.

It's possible that's not optimal at some point, so it's free-to-refactor territory.

src/main/java/com/limechain/rpc/methods/state/StateRPCImpl.java
```diff
@@ -104,14 +107,18 @@ public List<String> stateGetKeysPaged(final String prefixHex, int limit, String
      * @param blockHashHex The block hash in hexadecimal format to scope the query.
      * @return The storage value in hexadecimal format for the given key within the specified block, or null if not found.
      */
+    // TODO: Figure out whether incoming requests will only ask for state in finalised block (i.e. db persisted blocks)
+    //  or also unfinalised blocks must be supported (i.e. these in BlockState.BlockTree)
+    //  For now we only handle persisted blocks (because we only read from the DB through trieStorage)
     public String stateGetStorage(final String keyHex, final String blockHashHex) {
```
David-Petrov, line 113: That's a mismatch in the implementation of this method: it fetches state root from `BlockState`, which also responds with unfinalised blocks, but then we query the `trieStorage`, which only handles persisted (finalised) blocks.

That's only an issue if an RPC request comes for an unfinalised block; so if that's never gonna happen, it's a no-issue (but still weird when inspecting due to this mismatch).

src/main/java/com/limechain/runtime/RuntimeBuilder.java
```diff
@@ -1,62 +1,150 @@
...
+     * @param code the runtime wasm bytecode
+     * @return a ready to execute `Runtime` instance
+     * @implNote Make sure the Spring context is properly initialized, as building the runtime this way relies on it.
      */
     public Runtime buildRuntime(byte[] code) {
+        var nodeRole = AppBean.getBean(HostConfig.class).getNodeRole();
+
+        var db = AppBean.getBean(KVRepository.class);
+        var localStorage = new OffchainStore(db, StorageKind.LOCAL);
+        var persistentStorage = new OffchainStore(db, StorageKind.PERSISTENT);
```
David-Petrov, line 63: That's a shortcut I took due to rushing out. The idea was to preserve the old interface (to not introduce too many rather out-of-scope smaller changes). In practice, this method packages "how we used to do it" and is a boundary of my refactor.

Originally, `RuntimeBuilder` is not the place for this convenience shortcut, since this method in its public API makes it dependent on the Spring context, and my intention for the builder was to decouple it (i.e. dependencies passed explicitly from outside). I'd simply move this somewhere else, but I couldn't find a "nice enough" place.

src/main/java/com/limechain/runtime/RuntimeBuilder.java
```diff
@@ -1,62 +1,150 @@
...
     /**
-     * Builds and returns a Runtime object for executing WebAssembly code.
+     * Builds and returns a ready-to-execute `Runtime` with dependencies from the global Spring context.
      *
-     * @param code The WebAssembly bytecode.
-     * @return A Runtime object.
+     * @param code the runtime wasm bytecode
+     * @return a ready to execute `Runtime` instance
+     * @implNote Make sure the Spring context is properly initialized, as building the runtime this way relies on it.
      */
```
David-Petrov, line 61: That's a shortcut I took due to rushing out. The idea was to preserve the old interface (to not introduce too many rather out-of-scope smaller changes). In practice, this method packages "how we used to do it" and is a boundary of my refactor.

By intent, `RuntimeBuilder` is not the place for this convenience shortcut, since this method in its public API makes it dependent on the Spring context, and my intention for the builder was to decouple it (i.e. dependencies passed explicitly from outside). I'd simply move this somewhere else, but I couldn't find a "nice enough" place.

src/main/java/com/limechain/runtime/RuntimeBuilder.java
```diff
@@ -1,62 +1,150 @@
...
     public Runtime buildRuntime(byte[] code) {
+        var nodeRole = AppBean.getBean(HostConfig.class).getNodeRole();
+
+        var db = AppBean.getBean(KVRepository.class);
+        var localStorage = new OffchainStore(db, StorageKind.LOCAL);
+        var persistentStorage = new OffchainStore(db, StorageKind.PERSISTENT);
+        // TODO:
+        //  This shouldn't be like that: base storage (for offchain_index api) is not the same as persistent storage.
+        //  Figure out the right approach, when it becomes relevant.
+        var baseStorage = persistentStorage;
```
David-Petrov, line 71: That's an out-of-scope one, but I noticed it.

The TODO regards the second of two issues; the first one being that our `localStorage` is acutally not fork-aware (just writes in the db) as of right now.

We discussed this is not a high priority for now, but I still feel the need to persist the knowledge of these issues somewhere... for when it becomes a problem (if at all).

src/main/java/com/limechain/runtime/RuntimeBuilder.java
```diff
@@ -1,62 +1,150 @@
...
+            sharedMemory,
+            null
+        );
+
+        // Construct the host API implementation
+        HostApi hostApi = hostApiProvider.apply(context);
+
+        // Build the imports
+        // TODO: Extract the memory import from the runtime, don't hardcode.
+        ImportObject.MemoryImport memory = new ImportObject.MemoryImport("env", 22, false);
```
David-Petrov, line 137: That's the place where @Zurcusa jumps in with #446.

src/main/java/com/limechain/runtime/RuntimeBuilder.java
```diff
@@ -72,31 +160,48 @@ private byte[] zstDecompressIfNecessary(byte[] code) {
...
+        @Nullable
+        KeyStore keyStore,
+        @Nullable
+        OffchainStorages offchainStorages,
+        boolean isValidator
+    ) {
+        public static final Config EMPTY = new Config(null, null, null, false);
     }
 
+    // TODO: Move this method somewhere else as it doesn't have to do with actually building a runtime instance
```
David-Petrov, line 218: This one's a big thorn for me. Has nothing to do with building a runtime instance; must be moved somewhere else. I've left this out for now as an out-of-scope change. 

That's essentially a static method. I would move it into `TrieDecoder` for example; or introduce a new utility class for that sole purpose; or move it closer to the user-site... not sure.

src/main/java/com/limechain/storage/offchain/BasicStorage.java
```diff
@@ -0,0 +1,35 @@
+package com.limechain.storage.offchain;
+
+import java.util.Arrays;
+
+/**
+ * A minimal interface for a basic key-value storage.
+ */
+public interface BasicStorage {
+    // TODO: Shouldn't these `key`s be byte[] instead of String? `String` is not a good container for raw `byte[]`s.
```
David-Petrov: For the purposes of pure storage interfacing, these are nothing more than `byte[]`.

I left them as `String`s since the underlying implementor `OffchainStore` uses our `KVRepository<String, Object>` and the easier way was to just declare the interface using Strings instead of changing a lot of `Strings` to `byte[]`s.  And not to litter the diff too much with side considerations. 

That's a small issue; introduced whilst refactoring priorly existing behaviour. Anyway, I would go on to fix this later.

src/test/java/com/limechain/runtime/RuntimeWasmTest.java
```diff
@@ -11,8 +10,13 @@
...
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
-@Disabled("This is an integration test")
+@Disabled("This is an integration test.")
 class RuntimeWasmTest {
+    // TODO:
+    //  This wasm is not a valid runtime since it doesn't obey the runtime API (doesn't export `Core_version`).
+    //  This test suite needs total refactoring, since our Runtime class models a wasm module obeying the contract
+    //  to call it a "Runtime", i.e. the entire runtime API, memory exports, etc. as per spec.
+    //  Currently, we don't have a smaller class modelling "an instance of an arbitrary wasm module"
```
David-Petrov: That's an important consideration, although it arises from this practically deprecated test.

I'd like to trace somewhere that we've made a conscious design decision on that - do we actually need appropriate infrastructure for things less than "a valid runtime wasm" as per spec?

This test, for example, currently fails because the provided source doesn't export `Core_version`, which we call on construction in order to cache the runtime's version in the context.

We might EITHER decide that the test is not necessary (although it points to an understandable desire to be able to test something smaller in isolation) OR decide that this test's desires are respectable, in which case we'd have to at least not rely on calling `Core_version` on instantiation... (to pass this test, at least), which then means that `RuntimeBuilder` is not actually a builder for runtimes only, but for any wasm module (breaks the spec contract), which then evolves into us introducing separate logic for handling wasm instances, which are not runtimes. We don't have that yet. The current design is concerned with building valid runtimes only.

src/test/java/com/limechain/sync/fullsync/BlockExecutorTest.java
```diff
@@ -23,68 +25,46 @@
...
+        var cfg = new RuntimeBuilder.Config(trieAccessor, null, null, false);
 
-        Runtime genesisRuntime = new RuntimeBuilder().buildRuntime(runtimeByteCode);
-        byte[] encodedUnsealedHeader =
-                ScaleUtils.Encode.encode(BlockHeaderScaleWriter.getInstance()::writeUnsealed, block.getHeader());
+        Runtime genesisRuntime = new RuntimeBuilder().buildRuntime(runtimeByteCode, cfg);
 
-        var encodedBody = ScaleUtils.Encode.encode(BlockBodyWriter.getInstance(), body);
+        // Now, prepare the arguments for the actual runtime calls
+        // TODO: Encapsulate these more properly
```
David-Petrov, line 66: The test is not yet finished (in the sense of being quality test code), but it otherwise does what's intended (in this sense, it's finished). 

The main issue is that it touches on not yet well structured logic (constructing runtime call arguments from a block; having a nice interface for common runtime invocations; etc.), which is out of scope for now.

Rushing out due to lack of time.

src/test/java/com/limechain/sync/fullsync/BlockExecutorTest.java
```diff
@@ -95,18 +75,43 @@ void kusamaFirstBlock() throws IOException {
...
 
         // Attempt to execute a block
-        log.info("Executing block");
-        byte[] executeBlockResponse = genesisRuntime.call("Core_execute_block", executeBlockParam);
+        assertDoesNotThrow(() -> genesisRuntime.call("Core_execute_block", executeBlockParam),
+            "Executing the first block of Kusama must not throw any exceptions.");
+
+        // Explicitly assert that we calculated the expected state root
+        // TODO: This recalculates the merkle root again... where do we fetch the actual result from... the db... and how?
+        byte[] actualRoot = trieAccessor.getMerkleRoot(StateVersion.V0);
```
David-Petrov, line 96: @ablax That's the issue I was asking you about. How would you do this; what's your take on that?

Formally described here [#439](https://github.com/LimeChain/Fruzhin/issues/439)
Touches on things I considered out of scope for now, so it's up for debate.

src/main/java/com/limechain/storage/trie/TrieStorage.java
```diff
@@ -560,7 +424,10 @@ private void loadSubTrie(TrieStructure<NodeData> trie, TrieNodeData currentNodeD
         if (currentNodeData.isBranchNode()) {
             trie.insertBranch(currentPath, new NodeData(null, merkleValue));
         } else {
-            byte[] value = currentNodeData.getValue() == null ? currentNodeData.getTrieRootRef() : currentNodeData.getValue();
+            byte[] value =
```
Zurcusa, line 427: It would be a good idea to keep code formatting in check. Afaik Sonarqube doesn't support that out of the box, but we could use something like [this](https://github.com/checkstyle/sonar-checkstyle)

src/main/java/com/limechain/rpc/methods/childstate/ChildStateRPCImpl.java
```diff
@@ -15,8 +16,9 @@
 
 @Service
 public class ChildStateRPCImpl {
+    @Autowired
```
Zurcusa, line 19: I suggest we strive away from field injection and use ctor one. See [this](https://www.baeldung.com/java-spring-field-injection-cons)

src/test/java/com/limechain/runtime/RuntimeWasmTest.java
vikinatora: I think `hello_world.wat` is not the proper wasm for this test. Valid runtime wasm files should be the standard files we use for any kind of runtime tests(unless the specific unit tests requires that).

src/main/java/com/limechain/runtime/RuntimeBuilder.java
```diff
@@ -1,62 +1,150 @@
...
+        // TODO:
+        //  This shouldn't be like that: base storage (for offchain_index api) is not the same as persistent storage.
+        //  Figure out the right approach, when it becomes relevant.
+        var baseStorage = persistentStorage;
+
+        RuntimeBuilder.Config cfg = new RuntimeBuilder.Config(
+            null,
+            null,
+            new OffchainStorages(localStorage, persistentStorage, baseStorage),
+            nodeRole == NodeRole.AUTHORING
```
Zurcusa, line 85: Is this meant to always be AUTHORING?

src/main/java/com/limechain/trie/BlockTrieAccessor.java
```diff
@@ -14,14 +15,14 @@ public class BlockTrieAccessor extends TrieAccessor {
      *
      * @param blockHash the block hash of the block whose trie is to be accessed
      */
-    public BlockTrieAccessor(Hash256 blockHash) {
-        super(BlockState.getInstance().isInitialized() ?
+    public BlockTrieAccessor(TrieStorage trieStorage, Hash256 blockHash) {
```
David-Petrov, line 18: The `BlockTrieAccessor` is now entirely obsolete IMO, since the problem it was created to solve (circumnavigate sprawling singletons) is solved in this PR. However, this class falls under general between-blocks logic, it's out of scope of the current PR.

src/main/java/com/limechain/runtime/hostapi/PartialHostApi.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Map;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/**
+ * An interface defining an implementation of a (not necessarily strict) subset of all {@link Endpoint}s.
+ * @apiNote Mainly intended for partitioning a total implementation into separate classes for readability
+ *          and separation of concerns.
+ */
+public interface PartialHostApi {
```
Zurcusa, line 15: Considering the `HostApi` and the `DefaultHostApi` this name is confusing. Maybe we could switch it to `PartialHostFunctions`

David-Petrov:
For some reason, not yet known to me, `BlockExecutorTest` doesn't pass in GitHub's test environment. We gotta fix this before merging.

src/test/java/com/limechain/runtime/RuntimeWasmTest.java
ablax: We should get rid of `hello_world.wasm`. `runtime.wasm` could be used instead - it is currently used for the MiscHostApi's getRuntimeVersion test

David-Petrov:
I also don't like the structure of the `runtime` package as-is. It was initially decomposed more sanely, but when I got to staging files I figured we'd lose the entire history (+ the PR's diffs would've become incomprehensible with all of the moved files), so I left this as a final step for a separate PR (only moving things around) after this one's done being reviewed.

ablax:
> For some reason, not yet known to me, `BlockExecutorTest` doesn't pass in GitHub's test environment. We gotta fix this before merging.

It's because of the wasmer-java not being able to be used in github action runner. That's the reason it was disabled in the first place

src/test/java/com/limechain/sync/fullsync/BlockExecutorTest.java
```diff
@@ -23,68 +25,46 @@
...
 import java.math.BigInteger;
 
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
+//@formatter:off
 @Log
-@ExtendWith(MockitoExtension.class)
-@Disabled("This test is currently broken in CI")
```
ablax, line 40: @David-Petrov ^

src/test/java/com/limechain/sync/fullsync/BlockExecutorTest.java
ablax: The assert on line 89 pretty much asserts that the expectedRoot is matching the actualRoot. Otherwise we do not purposefully store our own calculated merkleRoot anywhere. Maybe you can make the mainTrieRoot variable in TrieAccessor gettable, as it is updated on each getMerkleRoot call, and use it as the actualRoot.

src/main/java/com/limechain/runtime/RuntimeBuilder.java
ablax: No, it's a check for the isValidator boolean

src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
```diff
@@ -203,7 +195,7 @@ public RuntimePointerSize extOffchainNetworkState() {
         PeerId peerId = network.getHost().getPeerId();
```
David-Petrov, line 201: This `network` is a static singleton dependency I've missed; must be extracted in the context.

src/main/java/com/limechain/runtime/RuntimeBuilder.java
David-Petrov: After some looking into [Murki's PR](https://github.com/LimeChain/Fruzhin/pull/447), `WarpSyncState` seems like the place where this method should live.

src/main/java/com/limechain/runtime/Runtime.java
georg-getz: `close` drops the Module pointer. Would be good to know whether that's necessary or simply a small improvement over the garbage collection.

src/main/java/com/limechain/runtime/Runtime.java
georg-getz: You can try with the former approach and for the start not even have different Contexts but rather just a single full one. We might not always need the full context but it would be a good test to see how performant it would be.

src/main/java/com/limechain/runtime/Runtime.java
```diff
@@ -63,87 +72,40 @@ private byte[] callInner(String functionName, RuntimePointerSize parameterPtrSiz
...
-     *
-     * @param data data to be written
-     * @return a pointer size to the written data
-     */
-    public RuntimePointerSize writeDataToMemory(byte[] data) {
-        RuntimePointerSize allocatedPointer = allocate(data.length);
-        writeDataToMemory(data, allocatedPointer);
-        return allocatedPointer;
+    RuntimeVersion callCoreVersion() {
+        return ScaleUtils.Decode.decode(this.call("Core_version"), new RuntimeVersionReader());
```
georg-getz, line 86: I feel like "Core_version" "Core_execute_block" and the rest should be listed as constants somewhere.

src/main/java/com/limechain/runtime/Runtime.java
```diff
@@ -63,87 +72,40 @@ private byte[] callInner(String functionName, RuntimePointerSize parameterPtrSiz
...
      */
-    public void writeDataToMemory(byte[] data, RuntimePointerSize runtimePointerSize) {
-        ByteBuffer memoryBuffer = getMemory().buffer();
-        memoryBuffer.position(runtimePointerSize.pointer());
-        memoryBuffer.put(data, 0, runtimePointerSize.size());
+    public void close() {
+        // TODO:
+        //  We're not sure whether we should also close the module.
+        //  That's what Gossamer does though, but their wasm runtime could be doing things differently underneath.
+        this.module.close();
```
georg-getz, line 96: Gossamer are irrelevant here. The method deletes the module pointer which is created because it is created in the rust part of wasmer-java. Have to manually test impact.

src/main/java/com/limechain/runtime/Runtime.java
David-Petrov: I feel the same. Didn't know how far to push it in this PR, but I'd also introduce some kind of modelling for the Runtime API.

src/main/java/com/limechain/trie/BlockTrieAccessor.java
georg-getz: So it has to be fully deleted?

src/test/java/com/limechain/runtime/RuntimeWasmTest.java
georg-getz: hello_world is created from a simple rust file printing hello world. It's not made to be a polkadot runtime but rather used because of it's small size for testing the 2 exports that every wasm file exports - data end and heap base.

David-Petrov: So it doesn't require any of the imports from the Host API, right? If that's the case, isn't this a test of `org.wasmer.Instance` instead, because I see no other logic being tested on our side. If that's the case, I'd remove it completely.

src/test/java/com/limechain/sync/fullsync/BlockExecutorTest.java
David-Petrov: I would restrain myself from further refactoring `BlockTrieAccessor` here. If you really think we wouldn't need the actual computed root after a block to be accessible for us (as it's only ever a concern of the runtime), then I'm ok with removing the assertion completely. 

This, however, touches on another decision within `TrieAccessor` - it mutates itself on calling `getMerkleRoot`, becoming an accessor for the next block automatically. I really don't like this happening implicitly (at least because `getMerkleRoot` does a lot more than it says as a method), so I'd at least make this mutation explicit... for this though, some (at least internal) knowledge of what the next state root is would have to be persisted somewhere (either in the block's header, or as a variable on user-site).

Anyhow, that's a topic for a further PR.

src/main/java/com/limechain/trie/BlockTrieAccessor.java
David-Petrov: I've absolutely confused the two classes here, I meant to comment that for the `AccessorHolder`. Ignore this discussion, [here](https://github.com/LimeChain/Fruzhin/pull/456#discussion_r1624225562)'s what I actually meant.

src/main/java/com/limechain/trie/AccessorHolder.java
```diff
@@ -30,7 +32,9 @@ public static AccessorHolder getInstance() {
      * @return         The BlockTrieAccessor instance representing the block trie with the specified root hash.
      */
     public BlockTrieAccessor setToStateRoot(byte[] lastRoot) {
-        this.blockTrieAccessor = new BlockTrieAccessor(lastRoot);
+        TrieStorage trieStorage = AppBean.getBean(TrieStorage.class);
```
David-Petrov, line 35: That's a boundary for tapping into spring beans. And it's a weird piece of code: the static `AccessorHolder` singleton relies on the `TrieStorage` from the spring context.

However, the `AccessorHolder` is now entirely obsolete IMO, since the problem it was created to solve (circumnavigate sprawling singletons) is solved in this PR. However, this class falls under general between-blocks logic, it's out of scope of the current PR.

src/main/java/com/limechain/storage/trie/TrieStorage.java
David-Petrov: I also totally agree, and that's a good catch. @georg-getz let's all agree on some formal definition of shared formatting and make sure we're all using the same formatting configuration. 😄

src/main/java/com/limechain/rpc/methods/childstate/ChildStateRPCImpl.java
David-Petrov: I agree completely. In general, the project needs some sort of convention on how we'll do dependencies, because now it's a mixture of spring beans (mainly used outside of spring components), static singletons, and, well..., the approach introduced in this PR (explicit manual context definition and management). 

The mentioned changes were introduced as a minimal quick fix in order to just push the boundaries as little as necessary for the purposes of sandboxing the runtime invocations and keep this PR as narrowly focused as possible.

Honestly, it was one of those small needed changes that would snowball into an avalanche of refactoring, so I spared this effort for this PR, since it'd quickly pivot into the above mentioned separate topic (how we do dependencies in general).

src/main/java/com/limechain/runtime/hostapi/PartialHostApi.java
David-Petrov: The analog to `HostApi` is intentional here. `HostApi` is an abstraction for a total implementation of the API, whilst `PartialHostApi` is an abstraction for a potentially partial implementation. The only difference between the two being the strictness of expectation for the provided `FuncImport`s.

You're right, however, that naming is not consistent. Initially, all `...HostFunctions` classes were renamed to `...HostApi`, but in order to keep diffs readable and preserve git history, I finally kept to the old names.

Now it's a matter of choice - which naming we'd want to choose, and then unify names to that convention. Personally, I like `StorageHostApi` more than `StorageHostFunctions`... Or to be more precise, `DefaultStorageApi` would be the most consistent (having to do with the fact, that we also possibly allow for other implementation; although we probably won't need this abstraction much outside of testing).

src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java
David-Petrov: Fixed in 5cf793d.

src/main/java/com/limechain/runtime/Runtime.java
David-Petrov: That's essentially what we have now in the method discussed [here](https://github.com/LimeChain/Fruzhin/pull/456#discussion_r1622220414) - a convenience method for building with **the** full production context (spring beans and what not), if I understand your remark correctly. Supporting different contexts is not an issue; the question is **when** we want to actually bind the context to the underlying instance. So yeah, for now that happens on construction of a `Runtime` instance (instance in the sense of OOP terminology).

What do you mean with "how performant it would be" exactly - to see if not being able to cache the `Module` would be an issue with the current constraints?

src/main/java/com/limechain/rpc/methods/state/StateRPCImpl.java
David-Petrov: @ablax I think we talked about this... any opinions (or conclusions)?

src/main/java/com/limechain/runtime/Runtime.java
David-Petrov: In any case, I think this topic is fine as-is for now. I don't plan on refactoring in this regard any further in this PR.

However, I really don't want to lose the discussion on that topic, or the knowledge of the other alternatives. We okay with that tedious TODO or where do we delegate this long-term? (I want to delete the TODO as it's rather big for an inline comment)

src/main/java/com/limechain/storage/offchain/BasicStorage.java
David-Petrov: Fixed in b2b3de0.

src/main/java/com/limechain/runtime/hostapi/PartialHostApi.java
Zurcusa: > Now it's a matter of choice - which naming we'd want to choose, and then unify names to that convention.

This means we would switch all `...HostFunctions` to `...HostApi` if I understand correctly. If that is the case I am okay with that. Considering there are javadocs incode what you explained about `HostApi` vs `PartialHostApi` seems logical.

src/test/java/com/limechain/runtime/RuntimeWasmTest.java
georg-getz: Yeah, feel free to remove it

src/test/java/com/limechain/sync/fullsync/BlockExecutorTest.java
georg-getz: Not sure what the question is but you can modify/delete this test as you see fit.

src/main/java/com/limechain/storage/trie/TrieStorage.java
georg-getz: Yea we can use the sonar-checkstyle as it seems it would be in sync with sonar cloud

src/main/java/com/limechain/runtime/Runtime.java
georg-getz: Not sure what that modelling would look like so you might stick to the few exported constants for a start.

src/test/java/com/limechain/runtime/RuntimeWasmTest.java
David-Petrov: Fixed in [9778418](https://github.com/LimeChain/Fruzhin/pull/456/commits/9778418c89b310d2b68990c3a0cf570651e1175f).

src/main/java/com/limechain/runtime/Runtime.java
David-Petrov: That's exactly what I meant - constants for the runtime api endoints and methods in `Runtime` for calling them. For now, an `executeBlock` makes sense. Delegated to the follow-up PR.

src/test/java/com/limechain/sync/fullsync/BlockExecutorTest.java
David-Petrov: This test is the pinnacle of what we're striving for, so I wouldn't delete it ever. 😄 
It is just not "beautiful" since it's still in its sketch form.
Delegated to the follow-up PR.

src/main/java/com/limechain/runtime/Runtime.java
David-Petrov: Delegated for the follow-up PR.

src/main/java/com/limechain/rpc/methods/state/StateRPCImpl.java
ablax: Well currently we don't have the state for unfinalized block whatsoever. Maybe we may want to start executing blocks from blockannounce and keeping their states until a finalization and pruning occures, but I doubt anyone will be sending rpc requests for unfinalized blocks

src/test/java/com/limechain/sync/fullsync/BlockExecutorTest.java
David-Petrov: Delegated to the follow-up PR.

src/main/java/com/limechain/runtime/hostapi/PartialHostApi.java
David-Petrov: Delegated to the follow-up PR.

src/main/java/com/limechain/runtime/Runtime.java
David-Petrov: Virtually the same topic as https://github.com/LimeChain/Fruzhin/pull/456#discussion_r1622192678

src/main/java/com/limechain/trie/AccessorHolder.java
David-Petrov: Delegated to the follow-up PR.

src/main/java/com/limechain/storage/trie/TrieStorage.java
Zurcusa: Super. I will create an issue in the "no milestone" section regarding this.

src/test/java/com/limechain/sync/fullsync/BlockExecutorTest.java
David-Petrov: I see this as a major blocker. What this means for now is that GH will reject any kind of tests we'd like to add around runtime invocations; which was one of the main goals of this PR.

Zurcusa (approved):

src/main/java/com/limechain/runtime/Runtime.java
David-Petrov: After brief discussion, the comment was extracted into TODO.md. Fixed in a395403, in the follow-up PR.

src/main/java/com/limechain/runtime/RuntimeBuilder.java
David-Petrov: Fixed in 00d07ab0ad7c371f92c6874200986cb4c61fbe42

src/test/java/com/limechain/sync/fullsync/BlockExecutorTest.java
David-Petrov: Fixed in [e222d87](https://github.com/LimeChain/Fruzhin/pull/464/commits/e222d870a46afcb797ef1ab9b9517f2b8c8bd6dc)

src/main/java/com/limechain/runtime/Runtime.java
David-Petrov: Proposed fix in [e222d87](https://github.com/LimeChain/Fruzhin/pull/464/commits/e222d870a46afcb797ef1ab9b9517f2b8c8bd6dc)

You might wanna take a look and see if it's fine for now. Let's continue the convo there.

src/main/java/com/limechain/trie/AccessorHolder.java
David-Petrov: `AccessorHolder` was entirely removed as not needed. Fixed in [e222d87](https://github.com/LimeChain/Fruzhin/pull/464/commits/e222d870a46afcb797ef1ab9b9517f2b8c8bd6dc)

src/test/java/com/limechain/sync/fullsync/BlockExecutorTest.java
David-Petrov: After encapsulating things around block execution more properly, I felt like this test doesn't need to be so explicit about argument serialization and trie root assertions, so I dropped those.

Serializing the arguments for runtime invocations is now a concern of `Runtime`, and a private implementation detail (for now), hence generally untested. We'll test on a higher level for this one.

See e222d870a46af for the relevant changes.

src/main/java/com/limechain/runtime/RuntimeBuilder.java
David-Petrov: After decent considerations, this issue has been resolved (as reasonably as possible for the current scope) in 66991d20

src/main/java/com/limechain/runtime/RuntimeBuilder.java
David-Petrov: The TODO remains, as this issue is considered a lower priority for now.

src/main/java/com/limechain/runtime/Runtime.java
David-Petrov: This field remains for now. It seems, really, like it's not necessary, and rather a small improvement (optimisation for memory freeing), since `module.close()` essentially just drops the module pointer (and calls `nativeDrop` to drop the native object). In any case, bear this in mind if future considerations arise when refactoring - it's not the strictest of necessities.

No benchmarking has been done, however, so this is purely a conclusion from static inspections of the source code of `wasmer-java`.

Disclosed in 43db682b

src/test/java/com/limechain/sync/fullsync/BlockExecutorTest.java
David-Petrov: @Zurcusa you could link to this conversation and resolve it, when (and if) your PR resolves this issue. Thanks in advance 👍

PR #463, opened 2024-06-04
refactor: Dynamic memory import.

### LINKED ISSUE #445: Adapt runtime module memory import logic so that it's not statically defined

Zurcusa:
# Description
Switch to dynamically loading the memory import from the provided wasm files. Also removes the manual parsing logic for the custom sections.

- What does this PR do?
Loads the memory import dynamically from the provided runtime wasm. Further Improves the custom section parsing.

Defining a static memory import causes runtime init failure when the runtimes change their minimum memory limits.

They're implemented with the help of [this parser.](https://github.com/dylibso/chicory/tree/main/wasm). The memory import is extracted before the initialization of the runtime instance and loaded in its imports.

Fixes #445 
## Checklist:


David-Petrov:
I see the `WasmCustomSection` record previously used as a DTO for the API of `WasmSectionUtils` is no longer used. If you deem that the newly imported dependency `chicory` renders the old API obsolete, then I'd also remove `WasmCustomSection`. Seems to me that this is the case.

src/main/java/com/limechain/runtime/WasmSectionUtils.java
```diff
@@ -1,124 +1,120 @@
...
         }
 
         // Read as many api versions as there are (we don't know the length for some reason).
-        ApiVersions apiVersions = ApiVersions.decodeNoLength(runtimeApis.content());
+        ApiVersions apiVersions = ApiVersions.decodeNoLength(runtimeApis.bytes());
 
         // Construct the runtime version partially uninitialized,
         // and then immediately set its apis we've read from the custom section.
-        RuntimeVersion version = new RuntimeVersionReader().read(new ScaleCodecReader(runtimeVersion.content()));
+        RuntimeVersion version = new RuntimeVersionReader().read(new ScaleCodecReader(runtimeVersion.bytes()));
```
David-Petrov: Nit to my old self :smile: :
```suggestion
        RuntimeVersion version = ScaleUtils.Decode.decode(runtimeVersion.bytes(), new RuntimeVersionReader());
```

src/main/java/com/limechain/runtime/WasmSectionUtils.java
```diff
@@ -1,124 +1,120 @@
...
+        } else if (moduleWithSections.importSection() != null) {
+            Optional<Import> parsedMemoryImport = moduleWithSections.importSection().stream()
+                .filter(i -> i.name().equals(MEMORY_IMPORT_NAME))
+                .findFirst();
+            if (parsedMemoryImport.isPresent()) {
+                initialPagesLimit = ((MemoryImport) parsedMemoryImport.get()).limits().initialPages();
+            }
+        }
+
+        return new ImportObject.MemoryImport(ENV_MODULE_NAME, initialPagesLimit, false);
```
David-Petrov: I see we're able to obtain the `shared` attribute as well from the `MemoryLimits`. If it's not always false per spec, I'd also add that to our `ImportObject.MemoryImport`.

Even if it is always false per spec, it would be good to check that in actuality.

src/main/java/com/limechain/runtime/WasmSectionUtils.java
```diff
@@ -1,124 +1,120 @@
...
+     * Parses the import section of a wasm blob and extracts the "memory" import if present.
+     *
+     * @param wasmBinary the wasm blob
+     * @return the parsed {@link org.wasmer.ImportObject.MemoryImport} if present,
+     * a default one with 24 initial pages otherwise
      */
-    @Nullable
-    public WasmCustomSection findRuntimeApisCustomSection(byte[] wasmBinary) {
-        return findCustomSection(wasmBinary, RUNTIME_APIS_SECTION_NAME);
+    public ImportObject.MemoryImport parseMemoryFromBinary(byte[] wasmBinary) {
```
David-Petrov, line 71: Just a thought on concern separation: I think `WasmSectionUtils` shouldn't decide how to instantiate the `ImportObject.MemoryImport` in case this information couldn't be fetched from the custom sections.

In that line of thought, I'd probably make the return type `Optional` or `@Nullable` to indicate whether and import object could be parsed from custom section; and delegate the default instantiation to the `RuntimeBuilder`.

Depends on how much we'd want to think of `WasmSectionUtils` as a separate class vs. an implementation detail of `RuntimeBuilder`.

In any case, it is a little mixed as of right now (as it previously was... sort of).

src/test/resources/memory_default.json
David-Petrov, line 1: Nit: for small enough data, I'd prefer to insert it directly into the tests' source code as to maximise local readability.
Anyways, that's a rather small nit, depending also on whether we'd want to treat `src.test.resources` as a well structured folder, understandable at a glance 😄

David-Petrov (changes requested):
Aside from the few comments I've left, LGTM! 👍 
P. S. Maybe Sonarlint's got a point, too? 👀

src/main/java/com/limechain/runtime/WasmSectionUtils.java
Zurcusa: > Even if it is always false per spec, it would be good to check that in actuality.

I agree with you on this one.

Side note, in our host we're loading only a single instance of a WASM module at a time (old one is scrapped when new one needs to be instantiated). This means that a shared memory would not be useful/should not exist. The `shared` flag should be true when multiple modules use the same memory in order to avoid collisions.

src/main/java/com/limechain/runtime/WasmSectionUtils.java
Zurcusa: I just noticed the fallback to `Core_version` resides in the Builder. Let's keep that same approach for this as well. The `WasmSectionUtils` should only have the responsibility to parse and create objects if present in the WASM.

src/test/resources/memory_default.json
Zurcusa: Imo having JSONs in code is not ideal. We can easily create directories in the resource folder and maintain the readability.

Zurcusa: I added a method to compare objects with non overridden equals methods and inaccessible fields in the `TestUtils`. That solved the issue with the short jsons.

src/main/java/com/limechain/runtime/WasmSectionUtils.java
David-Petrov: I see, that's clear. Soon enough though, this observation
> in our host we're loading only a single instance of a WASM module at a time (old one is scrapped when new one needs to be instantiated)

might not be true. For now it is, but we'll see how this evolves.

David-Petrov (approved):
It's super!

PR #464, opened 2024-06-06
refactor!: runtime context follow-up

David-Petrov:
This is the follow-up PR of #456

Each commit in this PR more or less tackles an aforementioned dangling issue discussion from the previous PR.

I suggest this PR would be easier reviewed commit per commit, since the entirety of file changes doesn't necessarily paint a coherent picture.

---

 A detailed map to the previous issues:
* [a395403](https://github.com/LimeChain/Fruzhin/pull/464/commits/a395403f58fe3a9df1bbbde5610af02fc633c973) resolves https://github.com/LimeChain/Fruzhin/pull/456#discussion_r1631276229
* [e222d87](https://github.com/LimeChain/Fruzhin/pull/464/commits/e222d870a46afcb797ef1ab9b9517f2b8c8bd6dc) fixes 
  * https://github.com/LimeChain/Fruzhin/pull/456#discussion_r1623948424
  * https://github.com/LimeChain/Fruzhin/pull/456#discussion_r1624225562
  * https://github.com/LimeChain/Fruzhin/pull/456#discussion_r1622268179 
  * https://github.com/LimeChain/Fruzhin/pull/456#discussion_r1622273420 
* [66991d2](https://github.com/LimeChain/Fruzhin/pull/464/commits/66991d206b11d2a357fdafbcdd488a147aff0ab8) fixes https://github.com/LimeChain/Fruzhin/pull/456#discussion_r1622220414
* [43db682](https://github.com/LimeChain/Fruzhin/pull/464/commits/43db682bca254c7cc9cc438a565a8be2ea1e9f4e) resolves https://github.com/LimeChain/Fruzhin/pull/456#discussion_r1622192678


src/main/java/com/limechain/runtime/builder/RuntimeBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * Serves as a convenience Spring bean to package all Spring beans necessary
+ * for common use-cases regarding building runtime instances.
+ * @implNote Composes over {@link RuntimeFactory}, which is the lower-level granular API for runtime instantiation.
+ */
+@Component
+@RequiredArgsConstructor
+public class RuntimeBuilder {
```
David-Petrov, line 27: Any suggestions for a better name, given the purpose of this class?

src/main/java/com/limechain/rpc/server/RpcApp.java
```diff
@@ -21,7 +21,8 @@
     "com.limechain.rpc.config",
     "com.limechain.rpc.methods",
     "com.limechain.rpc.server",
-    "com.limechain.storage"
+    "com.limechain.storage",
+    "com.limechain.runtime.builder",
```
David-Petrov, line 25: This is a shortcut (doesn't necessarily scale well). I prioritized the explicit intent that `RuntimeBuilder` is not a class that is supposed be independently used, since it only makes sense as a spring bean... that's its whole purpose.

In any case, that's a collision of design principles - in the long term I'd EITHER get rid of this explicit component scan and migrate all actual beans into proper bean classes; OR stick to the current  paradigm and move the logic currently residing in `RuntimeBuilder` into wherever otherwise appropriate.

src/main/java/com/limechain/runtime/RuntimeEndpoint.java
```diff
@@ -0,0 +1,20 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+/**
+ * Holds all Runtime API endpoints' names.
+ */
+@Getter
+@AllArgsConstructor
+// TODO: Define all endpoints
```
David-Petrov: I could do this now, but since I'm not yet sure we're sticking to this design decision, I thought I'd wait on discussions.

Anyway, this is the class to incrementally add new endpoint definitions to, whenever necessary. In this sense, it's not yet finished.

src/main/java/com/limechain/rpc/server/RpcApp.java
David-Petrov: This also has to do with the existence of the `com.limechain.runtime.builder` package. I really don't like it as a subpackage of `com.limechain.runtime` as this `RuntimeBuilder` generally should live outside of it. I couldn't find a proper super-package to put it into, so that's why I sub-packaged.

src/main/java/com/limechain/runtime/RuntimeEndpoint.java
```diff
@@ -0,0 +1,20 @@
...
+public enum RuntimeEndpoint {
+    CORE_EXECUTE_BLOCK("Core_execute_block"),
+    CORE_VERSION("Core_version"),
+    BLOCKBUILDER_CHECK_INHERENTS("BlockBuilder_check_inherents"),
+    METADATA_METADATA("Metadata_metadata"),
+    ;
```
Zurcusa, line 18:
```suggestion
    METADATA_METADATA("Metadata_metadata");
```

Zurcusa: Hm I didn't know compiler doesn't complain about this. You can leave it if it's intentional.

src/main/java/com/limechain/runtime/RuntimeEndpoint.java
Zurcusa: I'm fine with this approach. However, I see some of the endpoints have parameters related to them. It could be a good idea to add a string field in this enum which would serve as a description of the endpoint/of the parameters required.

src/main/java/com/limechain/rpc/server/RpcApp.java
Zurcusa: In the case of `RpcApp` is there a reason why we have it under `com.limechain.rpc.server`? If we bring it a level higher  into the "root directory" of the rpc we can skip adding `com.limechain.rpc.XXX`. 
Regarding the other beans, if we do not have the `RuntimeBuilder` as a RPC specific class (under the rpc packages) I do not see an issue of scanning it's package separately. When we point to a specific package containing only a single file that does not bring much overhead.

src/main/java/com/limechain/runtime/RuntimeEndpoint.java
David-Petrov: I tend to do this for larger WIP enums, where new line additions are expected toward the end. It is purely a style preference - for ease of adding new lines / moving lines around, to avoid the diff with the semicolon, etc.

src/main/java/com/limechain/runtime/RuntimeEndpoint.java
David-Petrov: I'm not sure we'd want a string field, since this info is developer-relevant only and resides in the spec anyways. In this sense, we have no use for descriptions being available in the java runtime; I see Javadocs rather fit. 
However, on the matter of the different parameters for each endpoint, I see `Runtime` as the place for constructing methods nicely encapsulating this interface (the runtime API) with methods, which will have proper javadocs (see `Runtime::executeBlock` for example). So the `RuntimeEndpoint` serves as nothing more than an exhaustive identifier for all existing function names (and a type to identify an `endpoint` by), but without holding the entire signature for the API. 

I am, however, open for better naming...

src/main/java/com/limechain/rpc/server/RpcApp.java
David-Petrov: My issue is not with overhead; I'm rather wary of the potential packaging chaos this would introduce. It's not intuitive and doesn't really follow any kind of structural convention / principle.

Regarding your first question - I'm not sure there's any concrete design intention behind that or how up-to-date that is with the current state of the source base. I feel more like that's how it once started and it's just stuck around till now.

Anyway, I just wanted to point this out as a potential "not nice thing" that's currently happening. It will become more apparent if you think about all the places that rely on `AppBean.getBean` (meaning that those classes are essentially spring components), that are sprinkled around everywhere. This `RuntimeBuilder` is just one of them that I've made explicit in this PR.

src/main/java/com/limechain/sync/fullsync/FullSyncMachine.java
```diff
@@ -65,28 +65,31 @@ public void start() {
...
-        runtime = getRuntimeFromState();
+        runtime = buildRuntimeFromState(blockTrieAccessor);
         blockState.storeRuntime(highestFinalizedHeader.getHash(), runtime);
 
         int startNumber = highestFinalizedHeader
                 .getBlockNumber()
                 .add(BigInteger.ONE)
                 .intValue();
+
         int blocksToFetch = 100;
```
Zurcusa, line 85: This looks likes something that could be a constant.

David-Petrov: This entire class is far from its final form. I've abstained myself from refactoring beyond my scope, so the changes here are rather necessitated by the desire to polish the `Runtime` class's interface; I'd say dealing with this class any further falls under the umbrella of improving the syncing strategy.

Regarding your concrete comment; it shouldn't even be a constant IMO, but for now it is. 😄

ablax (approved):

src/main/java/com/limechain/runtime/builder/RuntimeBuilder.java
Zurcusa: `RuntimeConfig` || `RuntimeConfigurer` considering it holds beans needed for the base configuration of the runtime.

Zurcusa (approved):

src/main/java/com/limechain/runtime/RuntimeEndpoint.java
Zurcusa: I have no suggestions for better naming. I'm marking the thread as resolved. +1 for javadocs

src/main/java/com/limechain/rpc/server/RpcApp.java
Zurcusa: My comment was not solely for mentioning overhead. See [this](https://docs.spring.io/spring-boot/docs/1.5.14.RELEASE/reference/html/using-boot-structuring-your-code.html#using-boot-locating-the-main-class). Moving the main class would be a step to improve the "structure chaos".
In this specific case I do not see a problem with the builder. Each bean has it's own domain (to some extent), thus lives in a different package. If in the future we need to create more "builders" we can easily have a factory for builders and have that in a separate package. In that way we'd always go through a single class when requesting a specific component (builder).

src/main/java/com/limechain/runtime/builder/RuntimeBuilder.java
David-Petrov: Yeah, but it essentially produces runtime instances as its main goal... just does that using spring beans as a concretisation over what `RuntimeFactory` expects in broader terms. I figure the JavaDoc might be a bit misleading as well...

In actuality, it is a `SpringRuntimeFactory`, but I have an uncanny feeling about naming stuff after a particular framework. Maybe that's an indicator for multiple responsibilities or stinky design, I'm not sure... but I couldn't come up with anything better with the current state of affairs. 😄

Are we okay with `RuntimeConfigurer::buildRuntime`? 😄

src/main/java/com/limechain/rpc/server/RpcApp.java
David-Petrov: I completely agree with you. I think you're saying what I implicitly meant with my EITHER clause above, so we're more or less on the same page 😄 The thing is: this entails some non-trivial refactoring as well, which I deemed out of scope for here. My main issue was that we've got it mixed right now - a lot of actual beans are not explicit beans, but serve (and can only be used) as such - those are the classes I meant above, which use `AppBean.getBean`. And I suspect no "structuring"/"packaging" has been thought of explicitly whilst introducing those `AppBea.getBean` usages, so if we go the way of explicitly listing all beans in the component scan annotation (if we suppose that we undertake the refactoring of making beans explicit), we're going to end up with a rather messy incoherent list. As an example, what I've introduced here with this `runtime.builder` **sub**package doesn't make a lot of sense. 

Otherwise, I'm entirely on the same page about each bean having its own domain. It's just that those domains are not well outlined right now.

Regarding the moving of `RpcApp` class, my response here:
> Regarding your first question - I'm not sure there's any concrete design intention behind that or how up-to-date that is with the current state of the source base. I feel more like that's how it once started and it's just stuck around till now.

essentially boils down to "I'm not even sure and haven't looked into it deeply enough". So after all, this feels like a broader topic that can be picked up separately. 😄 

Feel free to resolve this discussion if you're happy with that last conclusion. 😃

src/main/java/com/limechain/runtime/builder/RuntimeBuilder.java
David-Petrov: After a brief discussion, this remains as-is for now. Think of it as a known imperfection 😄

PR #466, opened 2024-06-10
build: Add wasmer-java jar with all wasmer-jni dynamic libraries

### LINKED ISSUE #462: Remove wasmer-setup folder and make dynamic libs be loaded dynamically
The goal of the task is to decouple the user's knowledge about the dynamic libs which handle wasm files by no longer having all of the possible os-arch combinations listed in a file but rather to get them from the releases of wasmer-java as they now include the dynamic lib combos of os-arch in the release assets. They are zipped because of name collision. [Here](https://github.com/LimeChain/wasmer-java/releases/tag/1.1.0) is an example release containing them.

Zurcusa:
# Description
Forcing the user to manually add the wasmer-jni libraries to their java extensions folder is quite a cumbersome task. Also not having all possible OS and architecture combinations in a wasmer-java.jar broke tests that are dependent on the wasmer library when running on different OS from the GH runners. That is fixes in this PR.

- What does this PR do?
Abolish the manual process of importing the wasmer-jni library to the users java extensions.
For an improved user experience.
By using a wasmer-java jar that includes all wasmer-jni libs per os and arch.

Fixes #462 

## Checklist:


src/test/java/com/limechain/runtime/hostapi/MiscellaneousHostFunctionsTest.java
```diff
@@ -66,7 +66,7 @@ void printHexV1() {
     }
 
     @Test
-    @Disabled("Does not work in github actions")
+    @Disabled("Failing probably because of runtime refactor.")
```
David-Petrov: Is fixed after the merge of #464 in 4f388cb7e2cc0434189d813fc4c65c30052457a9. Merge or rebase on `dev` and try again. 😄

georg-getz (approved):

PR #468, opened 2024-06-21
feat: streamline hanging stateversions

### LINKED ISSUE #438: Streamline hanging hardcoded `StateVersion`s into parameters
> - [ ] [LocalStorage](https://github.com/LimeChain/Fruzhin/blob/e068f0a5cb2dd81e207119dc3673bb175eba85c3/src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java#L70) must be [fork-aware](https://spec.polkadot.network/chap-host-api#defn-offchain-local-storage), which ours is not. Aside from extracting this dependency, we must consider where the use site is going to be regarding fork awareness.
> Gossamer solves this by instantiating a as a new in-memory DB, but I'm not sure when / how those get persisted if at all. Not that high priority for now as it's not been a blocker. 
> 
> - [ ] Take time go properly handle `per-node` state-version... or at least streamline hanging hardcoded `StateVersion`s into parameters. [Murad]
> 
> - [ ] Migrate trie from in-memory to on-disk. [Yordan to research & work on + support from David & Murad]
> 
> - [ ] After executing a block, the `BlockTrieAccessor` simply dumps all the trie nodes in the database (i.e. to even retrieve it from the DB later, we must at least know the root hash... to identify the trie) and the information about what root we've managed to calculate is lost (i.e. we have no link saying "this is the root hash for this block"). Has to do rather with the `BlockState` and block finalisation / handling of forks; BUT I've stumbled upon this being an issue whilst trying to write the assertion for the `kusamaFirstBlock` test case...: after we've executed the block, we can't actually "see what the result of the execution was", i.e. "which is the hash that had been calculated"... so my temporary workaround is to simply recalculate it again from the last state of the in-memory trie... which is not that cool. 😄 [David & Murad to sync] 

 _Originally posted by @David-Petrov in [#355](https://github.com/LimeChain/Fruzhin/issues/355#issuecomment-2109743987)_

ablax:
# Description

- What does this PR do?
Add state version 1 support to the triestructure
To be able to execute latest blocks and load latest trie.

Fixes #438

## Checklist:


src/main/java/com/limechain/runtime/hostapi/TrieHostFunctions.java
```diff
@@ -36,6 +36,7 @@
 @Log
 public class TrieHostFunctions implements PartialHostApi {
     static final int TRIE_ROOT_HASH_BYTE_LEN = 32;
+    public static final StateVersion STATE_VERSION_V0 = StateVersion.V0;
```
Zurcusa: I'd keep the usage uniform with other classes instead of creating a constant. The enum already has a meaningful name and is created only once during class init, so I don't see the advantage of a constant.

src/main/java/com/limechain/trie/TrieStructureFactory.java
Zurcusa, line 1: We should create a shared intellij code format file so that we avoid having many diffs when auto formatting.

src/test/java/com/limechain/trie/structure/TrieStructureTest.java
Zurcusa, line 1: The holy book states:
"8 indent tabs is haram"

src/test/java/com/limechain/trie/TrieAccessorTest.java
Zurcusa, line 1: Do we have a test somewhere if the stored node values are actually correctly hashed when >= 32 bytes and state version is 1?

ablax: currently we don't have any trie root calculation tests (neither state version 0 nor 1)

src/main/java/com/limechain/trie/TrieStructureFactory.java
ablax: yes, we should

Zurcusa (approved):

src/main/java/com/limechain/trie/structure/Vacant.java
```diff
@@ -419,52 +430,57 @@ private Two(TrieStructure<T> trieStructure,
...
-                    //  I'll never get those three hours of my life back :)
-                    //  Manage your ownership, kids.
-                    ArrayUtils.clone(this.branchChildrenIndices),
-                    false,
-                    branchUserData
+                        this.branchParent,
+                        this.branchPartialKey,
+                        // NOTE: Crucial mutability moment, this has to be a copy of the array.
+                        //  I'll never get those three hours of my life back :)
+                        //  Manage your ownership, kids.
```
georg-getz: I know it's not written by you but you can delete the second and 3rd row of that comment.

georg-getz (approved):

PR #471, opened 2024-06-26
437 on disk storage

### LINKED ISSUE #437: Migrate trie from in-memory to on-disk
> - [ ] [LocalStorage](https://github.com/LimeChain/Fruzhin/blob/e068f0a5cb2dd81e207119dc3673bb175eba85c3/src/main/java/com/limechain/runtime/hostapi/OffchainHostFunctions.java#L70) must be [fork-aware](https://spec.polkadot.network/chap-host-api#defn-offchain-local-storage), which ours is not. Aside from extracting this dependency, we must consider where the use site is going to be regarding fork awareness.
> Gossamer solves this by instantiating a as a new in-memory DB, but I'm not sure when / how those get persisted if at all. Not that high priority for now as it's not been a blocker. 
> 
> - [ ] Take time go properly handle `per-node` state-version... or at least streamline hanging hardcoded `StateVersion`s into parameters. [Murad]
> 
> - [ ] Migrate trie from in-memory to on-disk. [Yordan to research & work on + support from David & Murad]
> 
> - [ ] After executing a block, the `BlockTrieAccessor` simply dumps all the trie nodes in the database (i.e. to even retrieve it from the DB later, we must at least know the root hash... to identify the trie) and the information about what root we've managed to calculate is lost (i.e. we have no link saying "this is the root hash for this block"). Has to do rather with the `BlockState` and block finalisation / handling of forks; BUT I've stumbled upon this being an issue whilst trying to write the assertion for the `kusamaFirstBlock` test case...: after we've executed the block, we can't actually "see what the result of the execution was", i.e. "which is the hash that had been calculated"... so my temporary workaround is to simply recalculate it again from the last state of the in-memory trie... which is not that cool. 😄 [David & Murad to sync] 

 _Originally posted by @David-Petrov in [#355](https://github.com/LimeChain/Fruzhin/issues/355#issuecomment-2109743987)_

- [ ] Implement a Java equivalent of `TrieDiff` and various `StorageChanges`
- [ ] Implement `TrieStructure` like interface, which has similar functionalities as the ones used in `TrieAccessor` for the in memory trie.
- [ ] Substitute all usages of `TrieAccessor` with the new addition.

Zurcusa:


src/main/java/com/limechain/rpc/methods/state/StateRPCImpl.java
```diff
@@ -66,22 +65,22 @@ public String[][] stateGetPairs(final String prefixHex, final String blockHashHe
 
         // FIXME: This seems to not traverse the entire trie, but a single level deep only
```
georg-getz, line 66: are we worried about the comment here?

src/main/java/com/limechain/rpc/methods/state/StateRPCImpl.java
```diff
@@ -255,29 +254,29 @@ public List<StorageChangeSet> stateQueryStorage(final List<String> keysHex, fina
...
-                .map(StringUtils::hexToBytes)
-                .map(key -> blockTrieAccessor
-                        .findMerkleValue(Nibbles.fromBytes(key))
-                        .orElse(null))
-                .filter(Objects::nonNull)
-                .map(StringUtils::toHexWithPrefix)
-                .toList();
+            .stream()
+            .map(StringUtils::hexToBytes)
+            // TODO change implementation logic as this is not correct
```
georg-getz, line 269: is this comment outside of the scope of the PR?

src/main/java/com/limechain/trie/DiskTrieService.java
```diff
@@ -0,0 +1,826 @@
...
+            Optional<Nibbles> result = findNextKey(prefix, nextPath, childNode);
+            if (result.isPresent()) {
+                return result;
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    public void upsertNode(byte[] merkleRoot, Nibbles key, byte[] storageValue, StateVersion stateVersion) {
```
georg-getz: What does 'upsert' mean?

src/main/java/com/limechain/trie/DiskTrieService.java
```diff
@@ -0,0 +1,826 @@
...
+            result.put(key, toPendingInsertUpdate(
+                newNodeValue,
+                key,
+                newNodeStateVersion,
+                newNodeChildren,
+                true));
+
+            return result;
+        }
+
```
georg-getz: you can segregate this private function into several, it's way too long currently.

src/main/java/com/limechain/trie/DiskTrieService.java
```diff
@@ -0,0 +1,826 @@
...
+                            node.getFullKey(), nibble, childMerkle, limit, deleted);
+                        if (recursionResult.isPresent()) {
+                            PendingTrieNodeChange change = recursionResult.get();
+                            node.getChildrenMerkleValues().set(nibble.asInt(),
+                                change instanceof PendingInsertUpdate c
+                                    ? c.newMerkleValue()
+                                    : null);
+                        }
+                    }
+                }
```
georg-getz: When you have nested logic like that it's always good to leave 1 comment shortly explaining what's going on. Might add such a comment in several places in this file.

src/main/java/com/limechain/trie/DiskTrieService.java
```diff
@@ -0,0 +1,826 @@
...
+                    PendingTrieNodeChange change = recursionResult.get();
+                    foundNodeChildrenCopy.set(nibble.asInt(),
+                        change instanceof PendingInsertUpdate c
+                            ? c.newMerkleValue()
+                            : null);
+                }
+            }
+        }
+
+
```
georg-getz: doulbe empty lines

src/main/java/com/limechain/trie/MemoryTrieAccessor.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public abstract sealed class MemoryTrieAccessor extends TrieAccessor
+    permits BlockTrieAccessor, MemoryChildTrieAccessor {
+
+    private final TrieStructure<NodeData> initialTrie;
+    private List<TrieNodeIndex> updates;
+    private DiskTrieService diskTrieService;
```
georg-getz: why not final?

src/main/java/com/limechain/trie/DiskTrieService.java
```diff
@@ -0,0 +1,826 @@
...
+import lombok.Getter;
+import lombok.extern.java.Log;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.*;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.IntStream;
+
+@Log
+public class DiskTrieService {
```
georg-getz, line 41: Should add Javadoc to this class as it is very integral.

src/main/java/com/limechain/trie/DiskTrieService.java
```diff
@@ -0,0 +1,826 @@
...
+import com.limechain.trie.dto.node.TraversedNode;
+import com.limechain.trie.structure.nibble.Nibble;
+import com.limechain.trie.structure.nibble.Nibbles;
+import com.limechain.trie.structure.node.TrieNodeData;
+import com.limechain.utils.HashUtils;
+import lombok.Getter;
+import lombok.extern.java.Log;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.*;
```
georg-getz: Remove wildcard import, change your IDE setting if not so by default.

src/main/java/com/limechain/trie/cache/node/PendingInsertUpdate.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "PendingInsertUpdate{" +
+            "newMerkleValue=" + Arrays.toString(newMerkleValue) +
+            ", partialKey=" + partialKey +
+            ", childrenMerkleValues=" + childrenMerkleValues +
+            '}';
+    }
```
georg-getz, line 39: None of these can be switched with a lombok annotation? `@EqualsAndHasCode`, `@ToString`

src/main/java/com/limechain/trie/cache/node/PendingTrieNodeChange.java
```diff
@@ -0,0 +1,4 @@
+package com.limechain.trie.cache.node;
+
+public interface PendingTrieNodeChange {
+}
```
georg-getz, line 4: Doesn't seem like a valuable interface

src/main/java/com/limechain/rpc/methods/state/StateRPCImpl.java
Zurcusa: Not for the scope of this PR.

src/main/java/com/limechain/rpc/methods/state/StateRPCImpl.java
Zurcusa: Indeed it is. I made changes so that it's not tightly coupled to the memory trie accessor.

src/main/java/com/limechain/trie/DiskTrieService.java
Zurcusa: It's usually a database operation meaning "update/insert". I find it appropriate for cases like this one where the method can do both operations depending on some underlying logic.

src/main/java/com/limechain/trie/DiskTrieService.java
Zurcusa: I'm working on a thorough documentation of this file specifically.

src/main/java/com/limechain/trie/MemoryTrieAccessor.java
Zurcusa: This is not meant for the MemoryTrieAccessor. It's leftover from testing that I've removed locally after this commit.

src/main/java/com/limechain/trie/cache/node/PendingInsertUpdate.java
Zurcusa: Lombok annotations are not usable on java records.

src/main/java/com/limechain/trie/cache/node/PendingTrieNodeChange.java
Zurcusa: Depends if we want to segregate updates from deletions in the cache. Initially I had the idea of doing so, but since we want to keep stale nodes for the prune trie we could remove the delete notion from the cache. @ablax  wdyt? Do we need to delete the stale nodes if they're no longer referenced in the main trie?

ablax: No, we should keep at least 256 blocks trie history. Everything after the 256 block limit can be pruned.

Zurcusa:
The main idea behind these changes are to move the in memory trie fully to the disk using a cache layer. That role is played by the `TrieChanges` class and put into use in `DiskTrieService`. The main flow that is executed is the same for each of the insert, update, delete operations:
- traverse cache
- traverse disk if needed
- execute the operation
- recalculate merkles for traversed nodes in reversed order starting from the result from the execution (newly inserted, deleted, etc)
- populate cache

src/main/java/com/limechain/storage/trie/TrieStorage.java
```diff
@@ -96,12 +99,13 @@ private NodeData getNodeFromDb(byte[] nodeMerkleValue, Nibbles key) {
         return getNodeFromDb(childMerkleValue, remainder.drop(1));
     }
 
-    private TrieNodeData getTrieNodeFromMerkleValue(@NotNull byte[] childMerkleValue) {
+    @Nullable
+    public TrieNodeData getTrieNodeFromMerkleValue(@NotNull byte[] childMerkleValue) {
```
georg-getz, line 103: why is this public?

src/main/java/com/limechain/trie/DiskTrieService.java
```diff
@@ -0,0 +1,1112 @@
...
+            Optional<Nibbles> result = findNextKey(prefix, nextPath, childNode);
+            if (result.isPresent()) {
+                return result;
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    public void upsertNode(Nibbles key, byte[] storageValue, StateVersion stateVersion) {
```
georg-getz, line 153: Add Javadoc to this function as well as it gets used a lot.

georg-getz (approved):

src/main/java/com/limechain/storage/trie/TrieStorage.java
Zurcusa: The method is crucial for the logic of the `DiskTrieService`. One of the performance drivers is that after traversing the cache layer we don't traverse the disk up until the last ancestor, but we directly jump to it via this method and continue traversal from there.

PR #472, opened 2024-06-26
feat: Add new DiskTrieAccessor and refactor TrieAccessor into an inte…

Zurcusa:


src/main/java/com/limechain/storage/DBRepository.java
Zurcusa, line 1: Removed from the KV repository as we don't seem to use the type of RocksDb implementation that supports transactions. Imo we never planned on implementing them in the KV repo. See [this](https://github.com/facebook/rocksdb/wiki/Transactions).

src/main/java/com/limechain/trie/MemoryTrieAccessor.java
```diff
@@ -0,0 +1,154 @@
...
+        trieStorage.updateTrieStorage(initialTrie, updates, StateVersion.V0);
+    }
+
+    /**
+     * Retrieves the Merkle root hash of the trie with the specified state version.
+     *
+     * @param version The state version.
+     * @return The Merkle root hash.
+     */
+    public byte[] getMerkleRoot(StateVersion version) {
```
Zurcusa, line 144: This method is not a part of the interface since we'll have differences in the implementation logic which would change the method signature. 
In the memory trie we create a separate trie accessor for each child trie and then get the root node from there, whereas in the disk one I plan to directly retrieve the merkle root from the node in the main trie which holds the referencing merkle value as storage.

src/main/java/com/limechain/rpc/methods/state/StateRPCImpl.java
```diff
@@ -255,29 +254,29 @@ public List<StorageChangeSet> stateQueryStorage(final List<String> keysHex, fina
...
-                .map(StringUtils::hexToBytes)
-                .map(key -> blockTrieAccessor
-                        .findMerkleValue(Nibbles.fromBytes(key))
-                        .orElse(null))
-                .filter(Objects::nonNull)
-                .map(StringUtils::toHexWithPrefix)
-                .toList();
+            .stream()
+            .map(StringUtils::hexToBytes)
+            // TODO change implementation logic as this is not correct
```
Zurcusa, line 269: @ablax

PR #479, opened 2024-07-11
feat: improve syncing strategy

### LINKED ISSUE #353: Improve syncing strategy
So far the user is required to manually select whether he wants the node to sync in full or warp mode. Given that node reputation is something we want to keep high it would be better to think of a strategy where most of the synching happens with warp sync.

ablax:
# Description

- What does this PR do?

Fixes #353

## Checklist:

### DIFF
--- a/src/main/java/com/limechain/network/protocol/state/StateMessages.java
+++ b/src/main/java/com/limechain/network/protocol/state/StateMessages.java
@@ -6,7 +6,6 @@ import com.limechain.exception.global.ThreadInterruptedException;
 import com.limechain.network.StrictProtocolBinding;
 import com.limechain.network.protocol.sync.pb.SyncMessage;
 import com.limechain.utils.StringUtils;
-import io.libp2p.core.AddressBook;
 import io.libp2p.core.Host;
 import io.libp2p.core.PeerId;
 import lombok.extern.java.Log;
@@ -22,10 +21,10 @@ public class StateMessages extends StrictProtocolBinding<StateController> {
         super(protocolId, protocol);
     }
 
-    public SyncMessage.StateResponse remoteStateRequest(Host us, AddressBook addrs, PeerId peer,
+    public SyncMessage.StateResponse remoteStateRequest(Host us, PeerId peer,
                                                         String blockHash, ByteString after) {
         try {
-            StateController controller = dialPeer(us, peer, addrs);
+            StateController controller = dialPeer(us, peer, us.getAddressBook());
 
             return controller
                     .sendStateRequest(StringUtils.remove0xPrefix(blockHash), after)
--- a/src/main/java/com/limechain/network/protocol/state/StateMessages.java
+++ b/src/main/java/com/limechain/network/protocol/state/StateMessages.java
@@ -1,5 +1,6 @@
 package com.limechain.network.protocol.state;
 
+import com.google.protobuf.ByteString;
 import com.limechain.exception.global.ExecutionFailedException;
 import com.limechain.exception.global.ThreadInterruptedException;
 import com.limechain.network.StrictProtocolBinding;
@@ -22,12 +23,12 @@ public class StateMessages extends StrictProtocolBinding<StateController> {
     }
 
     public SyncMessage.StateResponse remoteStateRequest(Host us, AddressBook addrs, PeerId peer,
-                                                        String blockHash) {
+                                                        String blockHash, ByteString after) {
         try {
             StateController controller = dialPeer(us, peer, addrs);
 
             return controller
-                    .sendStateRequest(StringUtils.remove0xPrefix(blockHash))
+                    .sendStateRequest(StringUtils.remove0xPrefix(blockHash), after)
                     .get(10, TimeUnit.SECONDS);
         } catch (ExecutionException | TimeoutException | IllegalStateException e) {
             log.log(Level.SEVERE, "Error while sending remote state: ", e);
--- a/src/main/java/com/limechain/runtime/SharedMemory.java
+++ b/src/main/java/com/limechain/runtime/SharedMemory.java
@@ -1,7 +1,7 @@
 package com.limechain.runtime;
 
-import com.limechain.runtime.hostapi.dto.RuntimePointerSize;
 import com.limechain.runtime.allocator.Allocator;
+import com.limechain.runtime.hostapi.dto.RuntimePointerSize;
 import com.limechain.runtime.memory.Memory;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
@@ -55,7 +55,7 @@ public class SharedMemory {
     public void writeData(byte[] data, RuntimePointerSize runtimePointerSize) {
         ByteBuffer memoryBuffer = memory.buffer();
         memoryBuffer.position(runtimePointerSize.pointer());
-        memoryBuffer.put(data, 0, runtimePointerSize.size());
+        memoryBuffer.put(data, 0, Math.min(data.length, runtimePointerSize.size()));
     }
 
     /**
--- a/src/main/java/com/limechain/storage/block/SyncState.java
+++ b/src/main/java/com/limechain/storage/block/SyncState.java
@@ -76,6 +76,7 @@ public class SyncState {
                 this.lastFinalizedBlockNumber = commitMessage.getVote().getBlockNumber();
             }
         } catch (HeaderNotFoundException ignored) {
+            //TODO: Ignored for now
         }
     }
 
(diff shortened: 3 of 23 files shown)


src/main/java/com/limechain/network/protocol/state/StateMessages.java
```diff
@@ -22,12 +23,12 @@ public StateMessages(String protocolId, StateProtocol protocol) {
     }
 
     public SyncMessage.StateResponse remoteStateRequest(Host us, AddressBook addrs, PeerId peer,
-                                                        String blockHash) {
+                                                        String blockHash, ByteString after) {
```
Zurcusa, line 25:
```suggestion
                                                        String blockHash,
                                                        ByteString after) {
```
Maybe this could be one of the style guidelines of the project. If we have too many method params for a single line we separate them on 1 per line. Imo it would look better than having different number of params on each line.

src/main/java/com/limechain/runtime/SharedMemory.java
```diff
@@ -55,7 +55,7 @@ public RuntimePointerSize writeData(byte[] data) {
     public void writeData(byte[] data, RuntimePointerSize runtimePointerSize) {
         ByteBuffer memoryBuffer = memory.buffer();
         memoryBuffer.position(runtimePointerSize.pointer());
-        memoryBuffer.put(data, 0, runtimePointerSize.size());
+        memoryBuffer.put(data, 0, Math.min(data.length, runtimePointerSize.size()));
```
Zurcusa, line 58: Since the SharedMemory is an important part of the host could you give a bit of info on why we're doing this?

src/main/java/com/limechain/storage/trie/TrieStorage.java
```diff
@@ -102,6 +102,11 @@ private TrieNodeData getTrieNodeFromMerkleValue(@NotNull byte[] childMerkleValue
         return (TrieNodeData) encodedChild.orElse(null);
     }
 
+
+    public boolean merkleValueExists(Hash256 lastFinelizedStateRoot) {
```
Zurcusa:
```suggestion
    public boolean merkleValueExists(Hash256 lastFinalizedStateRoot) {
```

src/main/java/com/limechain/sync/fullsync/FullSyncMachine.java
```diff
@@ -96,6 +102,32 @@ public void start() {
...
+
+        makeStateRequest(lastFinalizedBlockHash, kvps, ByteString.EMPTY);
+
+        TrieStructure<NodeData> trieStructure = TrieStructureFactory.buildFromKVPs(kvps);
+        trieStorage.insertTrieStorage(trieStructure);
+
+        kvps.clear();
+    }
+
+    private void makeStateRequest(Hash256 lastFinalizedBlockHash, Map<ByteString, ByteString> kvps, ByteString start) {
```
Zurcusa: Mutating a data structure in a method is a bit hard to read for me. Maybe have this method return a populated Map<ByteString, ByteString> for kvps.

src/main/java/com/limechain/sync/fullsync/FullSyncMachine.java
```diff
@@ -119,8 +151,8 @@ private List<Block> requestBlocks(int start, int amount) {
             List<SyncMessage.BlockData> blockDatas = response.getBlocksList();
 
             return blockDatas.stream()
-                .map(FullSyncMachine::protobufDecodeBlock)
-                .toList();
+                    .map(FullSyncMachine::protobufDecodeBlock)
```
Zurcusa, line 176: I think it's time to agree on how many indents we'll have on code reformat. We keep conflicting with each other. Can we agree to have them on 4 tabs?

src/main/java/com/limechain/sync/fullsync/FullSyncMachine.java
```diff
@@ -154,7 +186,11 @@ private void executeBlocks(List<Block> receivedBlockDatas, BlockTrieAccessor blo
         for (Block block : receivedBlockDatas) {
             log.fine("Block number to be executed is " + block.getHeader().getBlockNumber());
 
-            blockState.addBlock(block);
+            try {
+                blockState.addBlock(block);
+            } catch (BlockNodeNotFoundException ignored) {
```
Zurcusa: I don't see `addBlock` throwing `BlockNodeNotFoundException`. Are you sure it's correct?

src/main/java/com/limechain/trie/TrieAccessor.java
Zurcusa, line 1: What's the idea behind this?

Zurcusa:
What's currently stopping us from enabling the integration tests? I think it would be a good idea to have a more thorough testing suite.

src/main/java/com/limechain/sync/fullsync/FullSyncMachine.java
ablax: BlockState addBlock calls BlockTree adblock, which checks for a parent node, but because we warped, we don't have the current parent and that's why it will throw this exception.

src/main/java/com/limechain/trie/TrieAccessor.java
ablax: The newer block executions were calling the start transction method -> which throws exception and made the runtime to panic

Zurcusa (approved):

src/main/java/com/limechain/runtime/SharedMemory.java
ablax: For some reason the Runtime was panicking when trying to write the data to a given pointer-size. In the given usecase the data's length was smaller than the size of the pointer. After this change the runtime stopped panicking, but when I think about it now it might cause issues in some use cases

src/main/java/com/limechain/network/protocol/state/StateMessages.java
georg-getz: Are there any occurrences of this being called with `AddressBook` that's not gotten from the `Host` object? If not then you can delete the second parameter.

src/main/java/com/limechain/storage/block/SyncState.java
```diff
@@ -66,11 +68,15 @@ public void finalizeHeader(BlockHeader header) {
...
-            this.stateRoot = blockByHash.getHeader().getStateRoot();
+        try {
+            Block blockByHash = BlockState.getInstance().getBlockByHash(commitMessage.getVote().getBlockHash());
+            if (blockByHash != null) {
+                this.stateRoot = blockByHash.getHeader().getStateRoot();
+                this.lastFinalizedBlockHash = commitMessage.getVote().getBlockHash();
+                this.lastFinalizedBlockNumber = commitMessage.getVote().getBlockNumber();
+            }
+        } catch (HeaderNotFoundException ignored) {
+            //TODO: Ignored for now
```
georg-getz: Don't we wanna at least log something? Maybe with a lower log level.

src/main/java/com/limechain/storage/trie/TrieStorage.java
```diff
@@ -102,6 +102,11 @@ private TrieNodeData getTrieNodeFromMerkleValue(@NotNull byte[] childMerkleValue
         return (TrieNodeData) encodedChild.orElse(null);
     }
 
+
```
georg-getz: consecutive empty lines

src/main/java/com/limechain/sync/fullsync/FullSyncMachine.java
```diff
@@ -154,7 +203,11 @@ private void executeBlocks(List<Block> receivedBlockDatas, BlockTrieAccessor blo
...
+                //todo: currently ignored
```
georg-getz: You can add a log here as well.

src/main/java/com/limechain/sync/fullsync/FullSyncMachine.java
```diff
@@ -177,11 +230,15 @@ private void executeBlocks(List<Block> receivedBlockDatas, BlockTrieAccessor blo
             blockTrieAccessor.persistUpdates();
 
             BlockHeader blockHeader = block.getHeader();
-            blockState.setFinalizedHash(blockHeader.getHash(), BigInteger.ZERO, BigInteger.ZERO);
+            try {
+                blockState.setFinalizedHash(blockHeader.getHash(), BigInteger.ZERO, BigInteger.ZERO);
+            } catch (BlockNodeNotFoundException ignored) {
+                //ignored
```
georg-getz: log

georg-getz (approved):

PR #525, opened 2024-09-05
feat: handle block tree after warp sync 

### LINKED ISSUE #389: Implement block tree pruning
It’s necessary to have access to the pruned block tree for the correct execution of part of the following algorithms.

Grigorov-Georgi:
# Description

- Introduced a mechanism to handle block announcements before the finish of the full-sync. Blocks received during the block announcement events are first collected into a queue. After the full-sync process, these queued blocks are processed and added into the block tree. After full-sync is complete, new blocks from block announcements are directly inserted into the block tree, bypassing the queue to streamline the process. Block announcement events happen concurrently and that's the reason why the method that interacts with them is marked as synchronized.

- When running the node in warp-sync mode, the block tree initialization starts from the last finalized block rather than the genesis block, setting this block as the root.

- Block finalization has been enabled for Kusama and Westend networks.

Closes -> https://github.com/LimeChain/Fruzhin/issues/389

### DIFF
--- a/src/main/java/com/limechain/network/ProtocolUtils.java
+++ b/src/main/java/com/limechain/network/ProtocolUtils.java
@@ -31,6 +31,10 @@ public final class ProtocolUtils {
         return String.format("/%s/kad", chainId);
     }
 
+    public static String getGrandpaProtocol(String chainId, boolean legacyProtocol) {
+        return String.format("/%s/grandpa/1", legacyProtocol ? "paritytech" : chainId);
+    }
+
     public static String getGrandpaProtocol(String chainId) {
         return String.format("/%s/grandpa/1", grandpaProtocolChain(chainId));
     }
--- a/src/main/java/com/limechain/network/ProtocolUtils.java
+++ b/src/main/java/com/limechain/network/ProtocolUtils.java
@@ -35,10 +35,6 @@ public final class ProtocolUtils {
         return String.format("/%s/grandpa/1", legacyProtocol ? "paritytech" : chainId);
     }
 
-    public static String getGrandpaProtocol(String chainId) {
-        return String.format("/%s/grandpa/1", grandpaProtocolChain(chainId));
-    }
-
     //TODO: figure out a more elegant solution
     private static String grandpaProtocolChain(String chainId) {
         return chainId.equals("dot") ? "paritytech" : chainId;
--- a/src/main/java/com/limechain/network/ProtocolUtils.java
+++ b/src/main/java/com/limechain/network/ProtocolUtils.java
@@ -35,11 +35,6 @@ public final class ProtocolUtils {
         return String.format("/%s/grandpa/1", legacyProtocol ? "paritytech" : chainId);
     }
 
-    //TODO: figure out a more elegant solution
-    private static String grandpaProtocolChain(String chainId) {
-        return chainId.equals("dot") ? "paritytech" : chainId;
-    }
-
     public static String getTransactionsProtocol(String chainId) {
         return String.format("/%s/transactions/1", chainId);
     }
--- a/src/main/java/com/limechain/network/protocol/blockannounce/BlockAnnounceEngine.java
+++ b/src/main/java/com/limechain/network/protocol/blockannounce/BlockAnnounceEngine.java
@@ -21,18 +21,18 @@ import io.libp2p.core.Stream;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.extern.java.Log;
-import org.javatuples.Pair;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.time.Instant;
 import java.util.ArrayList;
 import java.util.logging.Level;
 
 @Log
 @AllArgsConstructor(access = AccessLevel.PROTECTED)
 public class BlockAnnounceEngine {
+
     public static final int HANDSHAKE_LENGTH = 69;
+
     protected ConnectionManager connectionManager;
     protected WarpSyncState warpSyncState;
     protected BlockAnnounceHandshakeBuilder handshakeBuilder;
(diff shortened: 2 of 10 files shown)


src/main/java/com/limechain/network/ProtocolUtils.java
```diff
@@ -31,6 +31,10 @@ public static String getKadProtocol(String chainId) {
         return String.format("/%s/kad", chainId);
     }
 
+    public static String getGrandpaProtocol(String chainId, boolean legacyProtocol) {
+        return String.format("/%s/grandpa/1", legacyProtocol ? "paritytech" : chainId);
+    }
+
     public static String getGrandpaProtocol(String chainId) {
```
Zurcusa: This one seems unused. Can we remove it?

src/main/java/com/limechain/storage/block/BlockState.java
```diff
@@ -53,6 +50,11 @@ public class BlockState {
     private Hash256 lastFinalized;
     @Getter
     private boolean initialized;
+    @Getter
```
Zurcusa: Do we need the getter here?

src/main/java/com/limechain/storage/block/BlockState.java
```diff
@@ -16,23 +16,20 @@
...
 
 import java.math.BigInteger;
 import java.time.Instant;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
```
Zurcusa: Avoid star imports. Change your intellij settings so that you don't accidentally add them.

src/main/java/com/limechain/storage/block/BlockState.java
```diff
@@ -904,4 +918,56 @@ public void handleFinalizedBlock(final Hash256 currentFinalizedHash) {
             //TODO: If currentFinalizedHash is not equal to subchain hash, delete subchain state trie
         }
     }
+
+    public synchronized void addBlockToBlockTree(BlockHeader blockHeader) {
+        if (isInitialized()) {
+            if (isFullSyncFinished()) {
+
+                processPendingBlocksFromQueue();
```
Zurcusa: Is there a reason why we try to reprocess the queue every time we add instead of processing once after full sync finish?

src/main/java/com/limechain/storage/block/BlockState.java
```diff
@@ -904,4 +918,56 @@ public void handleFinalizedBlock(final Hash256 currentFinalizedHash) {
...
+                        log.fine(String.format("[%s] %s", blockHeader.getHash().toString(), ex.getMessage()));
+                    }
+
+                    return;
+                }
+            }
+
+            addBlockToQueue(blockHeader);
+        }
+    }
```
Zurcusa: Imo it's good to avoid nested ifs when we can.

```suggestion
    public synchronized void addBlockToBlockTree(BlockHeader blockHeader) {
        if (!isInitialized()) {
            // either log and return or throw an exception
        }

        if (!isFullSyncFinished()) {
            addBlockToQueue(blockHeader);
            return;
        }
        
        processPendingBlocksFromQueue();

        if (getPendingBlocksQueue().isEmpty()) {
            try {
                addBlock(new Block(blockHeader, new BlockBody(new ArrayList<>())));
            } catch (BlockStorageGenericException ex) {
                log.fine(String.format("[%s] %s", blockHeader.getHash().toString(), ex.getMessage()));
            }
    }
```

Zurcusa:
Don't forget to check for unused imports. Good time to bring up checkstyle again.

src/main/java/com/limechain/network/ProtocolUtils.java
ablax: It will be used for grandpa substream

src/main/java/com/limechain/storage/block/BlockState.java
Grigorov-Georgi: With the current implementation, there should be only one invocation of the method, but for my previous attempts, it was necessary to process them at least twice and I missed changing that.

Grigorov-Georgi: After some investigation, I recalled the exact reason I chose to do it every time. Making processPendingBlocksFromQueue() public and calling it from FullStateMachine introduces the risk of adding new blocks from block announcements while the old ones are being processed. Since I'm using a non-thread-safe queue, this can lead to unpredictable behavior. By making only the addBlockToBlockTree public and synchronized at the same time, I can be sure that new blocks will not be added while the old ones are processed. With that in mind, there are two options: 1. To keep that solution (which imo doesn't have big overhead because if the queue is empty the processPendingBlocksFromQueue() does nothing) or 2. To use a thread-safe queue. @Zurcusa What do you think would be the better approach?

Zurcusa: You're right. Overhead is close to none. No need to waste time changing it.

src/main/java/com/limechain/network/protocol/blockannounce/BlockAnnounceEngine.java
```diff
@@ -85,20 +87,12 @@ private void handleBlockAnnounce(byte[] msg, PeerId peerId) {
...
-        if (BlockState.getInstance().isInitialized()) {
-            try {
-                BlockState.getInstance().addBlock(new Block(announce.getHeader(), new BlockBody(new ArrayList<>())));
-            } catch (BlockNodeNotFoundException ignored) {
-                // Currently we ignore this exception, because our syncing strategy as full node is not implemented yet.
-                // And thus when we receive a block announce and try to add it in the BlockState we will get this
-                // exception because the parent block of the received one is not found in the BlockState.
-            }
-        }
+        BlockState.getInstance().addBlockToBlockTree(announce.getHeader());
```
ablax: It would be good idea to add the check if the BlockState is initialized, because we do not use the BlockState when the node is started in light client mode.

src/main/java/com/limechain/storage/block/SyncState.java
```diff
@@ -76,7 +76,10 @@ public void finalizedCommitMessage(CommitMessage commitMessage) {
                 this.stateRoot = blockByHash.getHeader().getStateRoot();
                 this.lastFinalizedBlockHash = commitMessage.getVote().getBlockHash();
                 this.lastFinalizedBlockNumber = commitMessage.getVote().getBlockNumber();
+
+                BlockState.getInstance().setFinalizedHash(commitMessage.getVote().getBlockHash(), commitMessage.getRoundNumber(), commitMessage.getSetId());
```
ablax: You can add the initialization check here as well

src/main/java/com/limechain/network/ProtocolUtils.java
```diff
@@ -31,8 +31,8 @@ public static String getKadProtocol(String chainId) {
         return String.format("/%s/kad", chainId);
     }
 
-    public static String getGrandpaProtocol(String chainId) {
-        return String.format("/%s/grandpa/1", grandpaProtocolChain(chainId));
+    public static String getGrandpaProtocol(String chainId, boolean legacyProtocol) {
+        return String.format("/%s/grandpa/1", legacyProtocol ? "paritytech" : chainId);
     }
 
     //TODO: figure out a more elegant solution
```
ablax: Maybe remove the method below as well

src/main/java/com/limechain/network/protocol/blockannounce/BlockAnnounceEngine.java
Grigorov-Georgi: The check is inside the method and throws exception if the state is not initialized

Zurcusa (approved):

georg-getz (approved):

ablax (approved):

PR #562, opened 2024-10-14
Initialize BABE constant

osrib:
- What does this PR do? Added Babe Api Configuration for constant initialization
- Why are these changes needed?  https://github.com/LimeChain/Fruzhin/issues/545

## Checklist:


Grigorov-Georgi (approved):

src/main/java/com/limechain/rpc/config/CommonConfig.java
Zurcusa, line 1: I suggest we start using `Component` and `Service` annotations for beans that require no configuration whatsoever. I think it will bring some advantages:

- A common approach for all similar beans that require no specific configuration upon initialization.
- Less boilerplate code as we won't need endlessly big configuration files.

 Currently it will bring the inconvenience of having to add the beans package to the SpringAppication class (RpcApp), but we should change that at some point and move it in the main package/find another place to start the spring app. It's a good start for cleaning up the current spring situation in the project.

src/main/java/com/limechain/runtime/RuntimeEndpoint.java
Zurcusa, line 1: Unrelated to your changes, but we could start ordering enums in an alphabetical order.

src/main/java/com/limechain/runtime/babeapi/BabeApiConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.math.BigInteger;
+import java.util.List;
+
+
+/**
+ * Current configuration of the BABE consensus protocol.
+ */
+@Getter
+@Setter
```
Zurcusa:
```suggestion
@Data
```
[Data](https://projectlombok.org/features/Data) would be more appropriate here. Keeps the same getter/setter with no arg setup, but also adds toString and Equals for tests etc.

src/main/java/com/limechain/runtime/babeapi/EpochState.java
Zurcusa, line 1: I have a few remarks regarding the configuration of this class:

- Avoid `Data` and `Value` for non DTO classes. They add annotations that we do not need. For example we don't need the `ToString` and `EqualsAndHashCode`. An `AllArgsConstructor` would be enough in our case.
- Avoid setters in states/most beans. It would be a better approach to provide an initial configuration (`BabeApiConfiguration` in this case) and initialize the state in its constructor or init method.
- We don't need `Component` and `Bean` in the configuration class. One of them is enough for Spring to register class as a bean. As my previous comments states, I'd go for the `Component`.

Regarding it's members, I think we shouldn't keep the configuration field itself. We can have multiple members that would split the epoch state in separate parts that get updated in a similar fashion during block production:

- slot duration and epoch length as they are constants only updated at genesis load. See [this](https://spec.polkadot.network/chap-runtime-api#tabl-babeapi-configuration).
- structure to hold current authority list and randomness
- structure to hold future epoch auth list and randomness (this can be done later when we implement the consensus message reader and epoch update)
- structure to hold current babe constant and 2nd slot config

See [this](https://spec.polkadot.network/sect-block-production#defn-consensus-message-babe).

Zurcusa:
We should create a separate task for adding the consensus message reader and updating the `EpochState` members appropriately.

src/main/java/com/limechain/runtime/RuntimeEndpoint.java
osrib: I can change whole enum to alphabetical order

src/main/java/com/limechain/runtime/babeapi/EpochState.java
osrib: Updated Structure. Added EpochData and EpochDescriptor to EpochState

src/test/java/com/limechain/runtime/babeapi/scale/BabeApiConfigurationReaderTest.java
```diff
@@ -0,0 +1,35 @@
...
+import com.limechain.utils.scale.ScaleUtils;
+import org.junit.jupiter.api.Test;
+
+import java.math.BigInteger;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+public class BabeApiConfigurationReaderTest {
+    private static final String BABE_API_CONFIGURATION_HEX = "0x701700000000000060090000000000000100000000000000040000000000000018fa3437" +
```
Zurcusa, line 16: Is this buggy for me or we have multiple empty lines?

src/main/java/com/limechain/babe/state/EpochData.java
```diff
@@ -0,0 +1,17 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+import java.util.List;
+
+/**
+ * Represents the supplied authority set and randomness.
+ */
+@Getter
+@AllArgsConstructor
```
Zurcusa, line 13:
```suggestion
@Value
```

src/main/java/com/limechain/babe/state/EpochDescriptor.java
```diff
@@ -0,0 +1,18 @@
...
+import java.math.BigInteger;
+
+/**
+ * Represents the BABE constant and secondary slot.
+ */
+@Getter
+@AllArgsConstructor
+public class EpochDescriptor {
+    private Pair<BigInteger, BigInteger> constant;
+    private BabeEpoch.BabeAllowedSlots allowedSlots;
```
Zurcusa, line 17:
```suggestion
@Value
public class EpochDescriptor {
    Pair<BigInteger, BigInteger> constant;
    BabeEpoch.BabeAllowedSlots allowedSlots;
```

src/main/java/com/limechain/babe/state/EpochData.java
```diff
@@ -0,0 +1,17 @@
...
+public class EpochData {
+    private List<Authority> authorities;
+    private byte[] randomness;
```
Zurcusa, line 16:
```suggestion
@Value
public class EpochData {
    List<Authority> authorities;
    byte[] randomness;
```

Zurcusa (approved):

src/main/java/com/limechain/babe/state/EpochDescriptor.java
osrib: I think this values are going to be changeble in runtime so @Value would provide immutability in this case

Zurcusa: In the current setup we also have some kind of an immutability, private fields with no setters that are changeable only by creating a new instance.

osrib: Yeah I planned to add setters while would be working on another task(for state update). Do you want me to add setters now?

Zurcusa: Ah then it's ok.

src/test/java/com/limechain/runtime/babeapi/scale/BabeApiConfigurationReaderTest.java
osrib: it is real configuration response converted to string hex. I dived it into parts for better readability

PR #564, opened 2024-10-16
feat: implement rotateKeys, insertKey, hasKey, hasSessionKeys author rpc api

### LINKED ISSUE #542: Implement the author RPC api.
Create the following:

- [x] author_rotateKeys - Used to generate and insert session keys in the host key-store.
- [x] author_insertKey - Used to insert a key into the key-store when it has been manually generated.
- [x] author_hasKey - Checks the key-store for an existing key entry.
- [x] author_hasSessionKeys - Same as hasKey, but for all session keys concatenated.
- [x] author_submitExtrinsic - Submit an extrinsic (e.g., key registration) for the validator to be registered.
- [x] author_submitAndWatchExtrinsic - Same as submitExtrinsic, but a pub-sub

Grigorov-Georgi:
# Description

1. author_rotateKeys() - Used to generate and insert session keys in the host key store. Runtime method ```generate_session_keys``` is called with ```seed = null``` and hex with 5 session keys for Babe, Grandpa, Audi, Im_online and Parachain_validator is returned.

2. author_insertKey(String keyType, String suri, String publicKey) - Used to insert a key into the key store when manually generated (```subkey generate --scheme ed25519``` and ```subkey generate --scheme sr25519``` are used). For the 'suri' is used the value of the secret seed from the ```subkey generate``` response. It supports two key types: ed25519 and sr25519. The process involves generating a private key from the provided seed and deriving the corresponding public key from the private key. The derived public key is then compared with the public key provided as an argument. If the derived public key matches the provided public key, the method considers the key pair valid and inserts it into the key store.

3. author_hasKey(String publicKey, String keyType) - Checks the key store for an existing key entry.

4. author_hasSessionKeys(String sessionKeys) - Same as hasKey, but for all session keys concatenated. Runtime method ```decode_session_keys``` is called in order to match every session key with the appropriate key type. Then author_hasKey() is called for every of the decoded keys.

Partially fixes: https://github.com/LimeChain/Fruzhin/issues/542


src/main/java/com/limechain/rpc/methods/author/AuthorRPCImpl.java
```diff
@@ -0,0 +1,169 @@
...
+import org.javatuples.Pair;
+import org.springframework.stereotype.Service;
+
+import java.util.Arrays;
+import java.util.List;
+
+@Service
+public class AuthorRPCImpl {
+
+    private final BlockState blockState = BlockState.getInstance();
```
Zurcusa: I suggest we set all beans and singletons inside of class constructors. It might be a good guideline to start following. Wdyt?

src/main/java/com/limechain/rpc/methods/author/AuthorRPCImpl.java
```diff
@@ -0,0 +1,169 @@
...
+
+        if (runtime == null) {
+            throw new IllegalStateException("Runtime is null");
+        }
+
+        byte[] response;
+        try {
+            response = runtime.call(endpoint, parameter);
+        } catch (Exception e) {
+            throw new IllegalArgumentException(e.getMessage());
```
Zurcusa: Maybe use a more specific exception? `WasmRuntimeException` for example.

src/main/java/com/limechain/rpc/methods/author/dto/DecodedKey.java
```diff
@@ -0,0 +1,14 @@
+package com.limechain.rpc.methods.author.dto;
+
+import com.limechain.storage.crypto.KeyType;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@AllArgsConstructor
```
Zurcusa:
```suggestion
@Value
```
Seems like this only uses the all arg ctor and could have final for all fields as they never change.

Zurcusa (approved):

PR #568, opened 2024-10-21
Implemented consensus message epoch state update

osrib:
- What does this PR do? Implementa consensus message epoch state update during block announcement
- Why are these changes needed? https://github.com/LimeChain/Fruzhin/issues/563

## Checklist:

### DIFF
--- a/src/main/java/com/limechain/babe/consesus/BabeConsensusMessage.java
+++ b/src/main/java/com/limechain/babe/consesus/BabeConsensusMessage.java
@@ -7,7 +7,7 @@ import lombok.Data;
 @Data
 public class BabeConsensusMessage {
     private EpochData nextEpochData;
-    private long   disabledAuthority;
+    private long disabledAuthority;
     private EpochDescriptor nextEpochDescriptor;
     private BabeConsensusMessageFormat format;
 }
--- a/src/main/java/com/limechain/babe/consesus/BabeConsensusMessage.java
+++ b/src/main/java/com/limechain/babe/consesus/BabeConsensusMessage.java
@@ -1,0 +1,13 @@
+package com.limechain.babe.consesus;
+
+import com.limechain.babe.state.EpochData;
+import com.limechain.babe.state.EpochDescriptor;
+import lombok.Data;
+
+@Data
+public class BabeConsensusMessage {
+    private EpochData nextEpochData;
+    private long   disabledAuthority;
+    private EpochDescriptor nextEpochDescriptor;
+    private BabeConsensusMessageFormat format;
+}
--- a/src/main/java/com/limechain/babe/consesus/BabeConsensusMessageFormat.java
+++ b/src/main/java/com/limechain/babe/consesus/BabeConsensusMessageFormat.java
@@ -1,0 +1,23 @@
+package com.limechain.babe.consesus;
+
+import lombok.Getter;
+
+@Getter
+public enum BabeConsensusMessageFormat {
+    ONE((byte) 1), TWO((byte) 2), TREE((byte) 3);
+
+    private final byte format;
+
+    BabeConsensusMessageFormat(byte format) {
+        this.format = format;
+    }
+
+    public static BabeConsensusMessageFormat fromFormat(byte format) {
+        for (BabeConsensusMessageFormat messageFormat : values()) {
+            if (messageFormat.getFormat() == format) {
+                return messageFormat;
+            }
+        }
+        throw new IllegalArgumentException("Unknown babe consensus message format: " + format);
+    }
+}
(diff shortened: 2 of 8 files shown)


src/main/java/com/limechain/network/protocol/blockannounce/BlockAnnounceEngine.java
```diff
@@ -91,6 +98,19 @@ private void handleBlockAnnounce(byte[] msg, PeerId peerId) {
         if (BlockState.getInstance().isInitialized()) {
             BlockState.getInstance().addBlockToBlockTree(announce.getHeader());
         }
+        updateEpochStateIfBabeConsensusMessageExists(announce);
```
osrib, line 101: So basically I think in future we need some **BabeSync class (or BabeHelper)** that could invoke this code, meaning that BlockAnnounceEngine would be more single responsible. Also we can put this logic inside of **warpSyncState.syncBlockAnnounce** . Or it could be some **DigestHelper**  that would have all logic needed for all types message handling. For now I put logic for consensus message into method below. What do you think?

src/main/java/com/limechain/babe/consesus/BabeConsensusMessage.java
```diff
@@ -0,0 +1,13 @@
+package com.limechain.babe.consesus;
+
+import com.limechain.babe.state.EpochData;
+import com.limechain.babe.state.EpochDescriptor;
+import lombok.Data;
+
+@Data
+public class BabeConsensusMessage {
+    private EpochData nextEpochData;
+    private long   disabledAuthority;
```
Zurcusa:
```suggestion
    private long disabledAuthority;
```

src/main/java/com/limechain/babe/consesus/BabeConsensusMessageFormat.java
Zurcusa, line 1: You can remove the byte casting here and have the field as int. When you use readByte from the scale reader you can then directly use the byte as an int.

src/main/java/com/limechain/babe/consesus/BabeConsensusMessageFormat.java
```diff
@@ -0,0 +1,23 @@
+package com.limechain.babe.consesus;
+
+import lombok.Getter;
+
+@Getter
+public enum BabeConsensusMessageFormat {
+    ONE((byte) 1), TWO((byte) 2), TREE((byte) 3);
```
Zurcusa: I would use more informative names. Eg. NEXT_EPOCH_DATA instead of 1, DISABLED_AUTHORITY instead of 2 etc.

src/main/java/com/limechain/babe/consesus/scale/BabeConsensusMessageReader.java
Zurcusa, line 1: Couldn't the consensus message have multiple data types inside? For example could a single message have both type 1 and type 3 or would we receive them as separate consensus messages?

src/main/java/com/limechain/network/protocol/blockannounce/BlockAnnounceEngine.java
Zurcusa: We're not yet sure where some digest types will be used to update the host states. Imo we could have something like the `DigestHelper` that you described. It can have a method that takes the `HeaderDigest[]` and based on each type a factory returns a specific type handler, which takes care of updating the states. For now it should be fine.

Zurcusa: Maybe add a // TODO so we don't forget about this.

src/main/java/com/limechain/network/protocol/blockannounce/BlockAnnounceEngine.java
```diff
@@ -91,6 +98,19 @@ private void handleBlockAnnounce(byte[] msg, PeerId peerId) {
...
+    }
+
+    public void updateEpochStateIfBabeConsensusMessageExists(BlockAnnounceMessage announce) {
+        HeaderDigest[] headerDigests = announce.getHeader().getDigest();
+        if (headerDigests != null && headerDigests.length > 0) {
+            Arrays.stream(headerDigests)
+                    .filter(headerDigest -> DigestType.CONSENSUS_MESSAGE.equals(headerDigest.getType()) &&
+                            ConsensusEngine.BABE.equals(headerDigest.getId()))
+                    .findFirst()
```
Zurcusa, line 111: I see a possible issue here. In the `BabeConsensusMessageReader` the logic assumes that we have only one type of babe consensus message data per message. Here we only retrieve a single consensus message using the `findFirst`. What happens if in a single block header we have multiple babe consensus messages / a single consensus message with more than one type?

src/main/java/com/limechain/babe/consesus/scale/BabeConsensusMessageReader.java
osrib: I didn't find any cases in documentation that it could have multiple values. Also in Go implemantation they also using switch. Honestly it's hard to answer tot his question fully

Zurcusa (approved):

PR #578, opened 2024-10-24
fix: Fix failing "generate_session_keys" runtime call on newer runtimes.

Zurcusa:
A key type was missing, which resulted in runtime failure on newer runtimes.


Zurcusa:
In polkadot spec it says that the "BEEFY" keys are of type `secp256k1`, but in our crypto host functions and also in [kagome](https://github.com/qdrvm/kagome/blob/2433e357341bed6fc2207f544395cc80e7c9550d/core/crypto/key_store/session_keys.cpp#L78) I see that it is appropriate to use ECDSA as well.

osrib (approved):

Grigorov-Georgi (approved):

PR #589, opened 2024-10-29
548 Implement Secondary slots

### LINKED ISSUE #548: Implement Secondary slots

Grigorov-Georgi:
# Description

- Add runtime call [BabeApi_current_epoch] to retrieve current `epochIndex` and `epochStartSlotNumber`, which is used to derive the `epochEndSlotNumber`. All of these epoch details are added to `EpochState` and are updated constantly when there is `BlockAnnounceMessage` containing `HeaderDigest`
- Implement a `claimSlot` method that verifies if the primary slot is claimable. If not, it should attempt to claim a secondary slot, selecting between a Plain or VRF slot
- Add a method for handling secondary slot claims, with logic that checks if the network supports Plain secondary slots

Fixes https://github.com/LimeChain/Fruzhin/issues/548


src/main/java/com/limechain/babe/Authorship.java
```diff
@@ -10,54 +15,194 @@
...
+                slotNumber,
+                epochIndex,
+                authorities,
+                c,
+                indexKeyPairMap
+        );
+
+        if (primarySlot != null) return primarySlot;
+
+        boolean authorSecondaryVrfSlot =
```
Zurcusa: Something is weird with the empty spaces here.

src/main/java/com/limechain/babe/Authorship.java
Zurcusa, line 1: I see both claim methods could return null. How are we going to handle this if it happens? If I understand correctly we only get a pre-digest if we have won a slot in the lottery. Will we use the slot number inside of the digest when deciding when to build a block afterwards ?

src/main/java/com/limechain/babe/Authorship.java
```diff
@@ -10,54 +15,194 @@
...
         return null;
     }
 
     // threshold = 2^128 * (1 - (1 - c) ^ (authority_weight / sum(authorities_weights)))
-    public static BigInteger calculatePrimaryThreshold(
+    private static BigInteger calculatePrimaryThreshold(
             @NotNull final Pair<BigInteger, BigInteger> constant,
             @NotNull final List<Authority> authorities,
-            final int authorityIndex) {
+            final Integer authorityIndex) {
```
Zurcusa: Is this nullable now?

src/main/java/com/limechain/network/protocol/warp/DigestHelper.java
```diff
@@ -37,6 +40,14 @@ private void updateEpochStateIfBabeConsensusMessageExists(HeaderDigest[] headerD
                 .findFirst()
                 .map(HeaderDigest::getMessage)
                 .ifPresent(message -> epochState.updateNextEpochBlockConfig(message));
+
+        if (BlockState.getInstance().isFullSyncFinished()) {
```
Zurcusa: This field is removed on my branch. Could you give me more info on why you need this here so I know when I am refactoring?

src/main/java/com/limechain/babe/Authorship.java
Grigorov-Georgi: My idea was to use Integer in order to use Integer.longValue() when adding the index to the PreDigest, but I agree is better to stay with the primitive int type.

src/main/java/com/limechain/babe/Authorship.java
Grigorov-Georgi: When comparing two values, I always try to keep them on one line to make the comparison more intuitive. But I see that I didn't manage to do that properly here.

src/main/java/com/limechain/network/protocol/warp/DigestHelper.java
Grigorov-Georgi: This runtime call isn't accessible in the earlier runtime state. The best solution I've found is to wait until the full sync finishes, at this point, the runtime can be called with BabeApi_current_epoch. I had two options for implementing the logic related to updating the current epoch fields: DigestHelper or BlockState. From a data encapsulation perspective, placing the logic in BlockState would be preferable since it keeps fullSyncFinished without needing a getter. However, the functionality seems more relevant to DigestHelper than BlockState, so I decided to add a getter and place the code within DigestHelper instead.

src/main/java/com/limechain/babe/Authorship.java
Grigorov-Georgi: When executing the `claimSlot` method at the start of each epoch for all of the slots, we should add the created predigests to an array, if the result is null, we can skip it. Then, when we begin building blocks, we can use the slot number and authority index to determine the appropriate slot to build and the correct keypair to use.

src/main/java/com/limechain/babe/Authorship.java
```diff
@@ -10,54 +15,194 @@
...
+                authorSecondaryVrfSlot
+        );
+    }
+
+    private static BabePreDigest claimPrimarySlot(final byte[] randomness,
+                                                  final BigInteger slotNumber,
+                                                  final BigInteger epochIndex,
+                                                  final List<Authority> authorities,
+                                                  final Pair<BigInteger, BigInteger> c,
+                                                  final Map<Integer, Schnorrkel.KeyPair> indexKeyPairMap) {
```
georg-getz: no need to put final in function parameters

src/main/java/com/limechain/babe/Authorship.java
```diff
@@ -82,16 +227,27 @@ public static BigInteger calculatePrimaryThreshold(
...
+                indexKeyPairMap.put(authorities.indexOf(authority), keyPair);
+            }
         }
 
-        if (authorityIndex >= authorities.size() || authorityIndex < 0) {
-            throw new IllegalArgumentException("Invalid denominator provided");
+        return indexKeyPairMap;
+    }
+
+    private static double getBabeConstant(@NotNull final Pair<BigInteger, BigInteger> constant) {
```
georg-getz: Can you remove the `@NotNull`

src/main/java/com/limechain/network/protocol/warp/DigestHelper.java
```diff
@@ -1,45 +1,47 @@
...
+//        if (headerDigests == null || headerDigests.length == 0) {
+//            return null;
+//        }
+//
+//        Map<DigestType, Digest> digests = new HashedMap<>();
+//        getBabeConsensusMessage(headerDigests).ifPresent(d -> digests.put(DigestType.CONSENSUS_MESSAGE, d));
+//        getBabePreRuntimeDigest(headerDigests).ifPresent(d -> digests.put(DigestType.PRE_RUNTIME, d));
+//
+//        return digests;
+//    }
```
georg-getz: Is this forgotten?

Grigorov-Georgi: No, I just committed a temporal state of the branch and didn't find how to put the PR on hold. It has been removed now, but there are new TODOs that will be fixed at https://github.com/LimeChain/Fruzhin/issues/593.

Zurcusa (approved):

PR #591, opened 2024-11-01
feat: Keep up with the head of the chain.

### LINKED ISSUE #573: Execute blocks from block announcement protocol.
Currently we only execute blocks up until full sync finishes. Meaning we stop at a specific finalized block and never continue to the tip of the chain. This causes many issues, one of which being inability to validate new transactions. For this to happen Fruzhin needs to be fork aware when executing blocks.

### LINKED ISSUE #574: Make state storage fork aware.
When we are executing non finalized blocks forks can happen. These forks need to only finalize their storage state (update the disk storage) when they are finalized. If a fork is pruned its changes are discarded and any existing transactions in the pruned blocks must be returned to the TQ or TP.

Zurcusa:
# Description

- What does this PR do?
Our node needs to keep up with the head of the chain. In other words, all announced blocks must be executed, and their states must be kept in memory before being finalized. This PR implements the functionality to do so. It also exposes a runtime API interface. This makes it more readable and easily extensible. A few other changes were made along the way that improved performance (async calls) and structured the project in a better way.

We need the latest state of the chain to validate transactions.

Fixes #574 and #573 

## Checklist:
- [X] I have read the [contributing guidelines](https://github.com/LimeChain/Fruzhin/blob/dev/CONTRIBUTING.md).
- [X] My PR title matches the [Conventional Commits spec](https://www.conventionalcommits.org/).
- [X] All new and existing tests passed.


src/main/java/com/limechain/rpc/methods/state/StateRPCImpl.java
```diff
@@ -264,19 +263,19 @@ public Map<String, Object> stateGetReadProof(final List<String> keyHexList, fina
...
-            // TODO change implementation logic as this is not correct
-            .map(key -> trieStorage.getByKeyFromMerkle(blockStateRoot, Nibbles.fromBytes(key))
-                .orElse(null))
-            .filter(Objects::nonNull)
-            .map(NodeData::getMerkleValue)
-            .map(StringUtils::toHexWithPrefix)
-            .toList();
+                .stream()
+                .map(StringUtils::hexToBytes)
+                // TODO change implementation logic as this is not correct
```
georg-getz: I know you didn't leave the comment but can you update it to point at the incorrect part

georg-getz (approved):

Grigorov-Georgi (approved):

PR #596, opened 2024-11-06
Implement ExecuteEpochLottery (EpochAuthorship)

### LINKED ISSUE #594: Implement EpochAuthorship method
The ```EpochAuthorship``` method will iterate from ```epochStartSlotNumber``` to ```epochEndSlotNumber``` to determine authorship status for each slot within the epoch. For each slot, it will check if a primary or secondary slot claim has been made, or if no claim exists.

Grigorov-Georgi:
# Description

- Created the BabeService class to handle block production.
- Implemented the ExecuteEpochLottery method, which iterates from the current epoch's start slot number to the current epoch's end slot number. It checks if a BabePreDigest can be created; if so, the slot and PreDigest are added to a map, which will later be used to create blocks for claimed slots.
- Only the current epoch’s start slot is fetched from the epoch state, while the end slot is calculated on the fly to avoid fetching the start of one epoch and the end of the next (probably this will never happen)
- Refactored EpochState: the epoch index field was removed, as it can be easily derived from the genesis slot. Now, the epoch index, epoch start slot number, and epoch end slot number are all derived from the genesis slot. Note that the genesis slot is not initialized at the moment, but this will be implemented soon.
- EpochState is now related only to the current epoch and is unsuitable for handling previous epochs. This is because calculations for the epoch index and epoch start slot number are based on the current slot number, and getter methods will provide the current epoch index and current epoch start slot number.

Fixes: https://github.com/LimeChain/Fruzhin/issues/594

### DIFF
--- a/src/main/java/com/limechain/babe/state/EpochState.java
+++ b/src/main/java/com/limechain/babe/state/EpochState.java
@@ -11,8 +11,9 @@ import java.math.BigInteger;
 import java.time.Instant;
 
 /**
- * Represents the state information for an epoch in the system.
- * This class encapsulates all the necessary configuration and parameters related to a specific epoch.
+ * Represents the state information for the current/next epoch that is needed
+ * for block production with BABE.
+ * Note: Intended for use only when the host is configured as an Authoring Node.
  */
 @Getter
 @Component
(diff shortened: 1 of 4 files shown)


src/main/java/com/limechain/babe/state/EpochState.java
```diff
@@ -46,4 +47,12 @@ public void updateNextEpochBlockConfig(byte[] message) {
     public BigInteger getCurrentSlotNumber() {
         return BigInteger.valueOf(Instant.now().toEpochMilli()).divide(slotDuration);
     }
+
+    public BigInteger getCurrentEpochStartSlotNumer() {
+        return getCurrentEpochIndex().multiply(epochLength).add(genesisSlotNumber);
+    }
+
+    public BigInteger getCurrentEpochIndex() {
+        return getCurrentSlotNumber().subtract(genesisSlotNumber).divide(epochLength);
```
Zurcusa, line 63: Maybe add some kind of reference/comment regarding the formula used. This could also be done for methods above.

georg-getz (approved):

Zurcusa (approved):

PR #600, opened 2024-11-07
Implement Genesis Slot Number Retrieval

### LINKED ISSUE #593: Epoch Index Calculation and Genesis Slot Number Retrieval
This task involves calculating the epoch index using the formula ```(currentSlotNumber - genesisSlotNumber) / epochLength```. To enable this, we need to implement a mechanism to retrieve the genesisSlotNumber from runtime storage using the genesis slot hash.

Grigorov-Georgi:
# Description

- Add a method to call Runtime Storage and retrieve the genesis slot number.
- Add a method to set the genesis slot number in EpochState. The provided value may be null, indicating that the chain has no executed blocks, in which case the current slot number is used as the genesis slot number.
- Add an isInitialized flag to EpochState, which is set to TRUE only when the goal is to run an authoring node.

Fixes: https://github.com/LimeChain/Fruzhin/issues/593

### DIFF
--- a/src/main/java/com/limechain/runtime/RuntimeStorageKey.java
+++ b/src/main/java/com/limechain/runtime/RuntimeStorageKey.java
@@ -4,6 +4,7 @@ import com.limechain.trie.structure.nibble.Nibbles;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 
+// all available keys can be found here: https://www.shawntabrizi.com/substrate-known-keys/
 @Getter
 @AllArgsConstructor
 public enum RuntimeStorageKey {
--- a/src/main/java/com/limechain/runtime/RuntimeStorageKey.java
+++ b/src/main/java/com/limechain/runtime/RuntimeStorageKey.java
@@ -4,11 +4,16 @@ import com.limechain.trie.structure.nibble.Nibbles;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/limechain/runtime/RuntimeStorageKey.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.limechain.trie.structure.nibble.Nibbles;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+// all available keys can be found here: https://www.shawntabrizi.com/substrate-known-keys/
+@Getter
+@AllArgsConstructor
+public enum RuntimeStorageKey {
+    GENESIS_SLOT(Nibbles.fromHexString("0x1cb6f36e027abb2091cfb5110ab5087f678711d15ebbceba5cd0cea158e6675a"));
```
Zurcusa:
```suggestion
    GENESIS_SLOT("babe genesisSlot");
```

We can have the keys in their string form and have a method hex and return nibbles.

Grigorov-Georgi: Keeping the keys in human-readable form means that we should then encode them with the xxHash algorithm, which we don't have as a dependency at the moment, so I decided to use the encoded form directly.

Zurcusa: ok

Zurcusa (approved):

PR #601, opened 2024-11-08
Implement EpochCoordinator

### LINKED ISSUE #599: Epoch coordinator

Grigorov-Georgi:
# Description

- Renamed to SlotCoordinator because is triggering events based on slot change.
- Add ```SlotCoordinator``` class that should check for slot changes and if the current slot is the last slot in the epoch
- ```executeEpochLottery``` in BabeService is invoked at the last slot of the current epoch -> when the isLastSlotInCurrentEpoch flag in the SlotChange event is ```true```
- Add methods for calculating the epoch start slot number for specified epoch index and also a method for calculating the last slot number of current epoch

## TODO
- Additional implementation should be added in ```BabeService``` that handles block creation on slot change

Fixes https://github.com/LimeChain/Fruzhin/issues/599


src/main/java/com/limechain/babe/coordinator/SlotChangeEvent.java
```diff
@@ -0,0 +1,32 @@
...
+public class SlotChangeEvent extends EventObject {
+
+    private final BigInteger slotNumber;
+    private final BigInteger epochIndex;
+    private final boolean isLastSlotFromCurrentEpoch;
+
+    /**
+     * Constructs a prototypical Event.
+     *
+     * @param source the object on which the Event initially occurred
```
Zurcusa, line 18: What's the point of source? It seems to always be the coordinator, which is a single instance. Can we skip it somehow?

src/main/java/com/limechain/sync/fullsync/FullSyncMachine.java
```diff
@@ -120,6 +124,7 @@ public void start() {
     private void initializeEpochState() {
         epochState.initialize(runtime.callBabeApiConfiguration());
         epochState.setGenesisSlotNumber(runtime.getGenesisSlotNumber());
+        slotCoordinator.start(List.of(babeService));
```
Zurcusa: I'd put `AppBean.getBean(BabeService.class)` as a local var here because it wouldn't be used anywhere else in the fullsync machine.

src/main/java/com/limechain/babe/coordinator/SlotCoordinator.java
```diff
@@ -0,0 +1,86 @@
...
+                this,
+                currentSlotNumber,
+                currentEpochIndex,
+                isLastSlot
+        );
+
+        notifySlotChangeListeners(event);
+    }
+
+    private void updateLocalSlotState(BigInteger currentSlotNumber) {
```
Zurcusa: Could you add a comment explaining that `localSlotState` is actually the state of the coordinator itself. The name is a bit misleading.

Zurcusa (approved):

src/main/java/com/limechain/babe/coordinator/SlotChangeEvent.java
Grigorov-Georgi: The java.util.EventObject requires that field. If we want to remove it, we should not extend the EventObject class and EventListener interface. I decided to follow the standard Java event-handling approach, but it would also be easy to switch to a custom implementation without EventObject and EventListener.

Zurcusa: Not worth it. Let's keep it.

PR #602, opened 2024-11-10
feat: Revert state storage updates for runtime calls...

Zurcusa:
# Description
Runtime now reverts all state storage changes originating from a runtime call different from `Core_execute_block`. This is needed when validating transactions. The runtime updates state storage when checking if a provided transaction is valid.

### DIFF
--- a/src/main/java/com/limechain/runtime/RuntimeImpl.java
+++ b/src/main/java/com/limechain/runtime/RuntimeImpl.java
@@ -48,7 +48,7 @@ public class RuntimeImpl implements Runtime {
     @Override
     public List<DecodedKey> decodeSessionKeys(String sessionKeys) {
         byte[] encodedRequest = ScaleUtils.Encode.encode(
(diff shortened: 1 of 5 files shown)


src/main/java/com/limechain/runtime/RuntimeImpl.java
Zurcusa, line 1: The null checks are needed since some initial runtime instances are created with an empty context. The trieAccessor there is null. That is not the case for any block aside from the genesis. Maybe that can be changed, but it's not a priority rn.

src/main/java/com/limechain/trie/TrieAccessor.java
```diff
@@ -26,11 +26,17 @@ public abstract sealed class TrieAccessor permits MemoryTrieAccessor, DiskTrieAc
     protected byte[] mainTrieRoot;
     @Setter
     protected StateVersion currentStateVersion;
+    // True if
```
Grigorov-Georgi: Consider removing or expanding this comment

Grigorov-Georgi (approved):

PR #608, opened 2024-11-12
Implement submitExtrinsic and submitAndWatchExtrinsic Author RPC methods

### LINKED ISSUE #542: Implement the author RPC api.

Grigorov-Georgi:
# Description

- Added the `author_submitExtrinsic` RPC method to submit an extrinsic to the transaction pool.
- Added the `author_submitAndWatchExtrinsic` WebSocket RPC method to submit an extrinsic and subscribe to its state updates.
- Added the `author_unwatchExtrinsic` method to end the session.
- Resolved the issue in the `TransactionPool` related to operating with byte[] keys in the transaction map. The problem is that the `.equals()` method for byte[] doesn't compare the content of the arrays but instead checks if they reference the same array. Since the HashMap implementation uses .equals() for nearly all of its operations, this behavior caused the `TransactionPool` to act strangely.

Fixes https://github.com/LimeChain/Fruzhin/issues/542

### DIFF
--- a/src/main/java/com/limechain/rpc/methods/author/AuthorRPCImpl.java
+++ b/src/main/java/com/limechain/rpc/methods/author/AuthorRPCImpl.java
@@ -117,7 +117,8 @@ public class AuthorRPCImpl {
         // TODO: Gossip this transaction to other peers
     }
 
-    public String authorSubmitAndWatchExtrinsic(String extrinsics) {
+    //TODO: This method might be removed from here - not sure at the moment of writing
+    public String authorSubmitAndWatchExtrinsic(String extrinsic) {
         return "";
     }
 
--- a/src/main/java/com/limechain/rpc/methods/author/AuthorRPCImpl.java
+++ b/src/main/java/com/limechain/rpc/methods/author/AuthorRPCImpl.java
@@ -117,11 +117,6 @@ public class AuthorRPCImpl {
         // TODO: Gossip this transaction to other peers
     }
 
-    //TODO: This method might be removed from here - not sure at the moment of writing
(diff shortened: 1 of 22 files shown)


georg-getz (approved):

src/main/java/com/limechain/rpc/methods/author/AuthorRPCImpl.java
```diff
@@ -82,12 +96,25 @@ public Boolean authorHasSessionKeys(String sessionKeys) {
...
+        try {
+            validTransaction = transactionValidator.validateExternalTransaction(decodedExtrinsic);
+        } catch (TransactionValidationException e) {
+            throw new ExecutionFailedException("Failed to executed submit_extrinsic call: " + e.getMessage());
+        }
+
+        return StringUtils.toHexWithPrefix(
+                transactionState.addToPool(validTransaction)
+        );
+        // TODO: Gossip this transaction to other peers
```
Zurcusa, line 117: Similar to block announce and the `BlockHandler`. I believe the announcement logic should belong in the `TransactionValidator`.

PR #609, opened 2024-11-13
387 Build block

Zurcusa:
# Description

Implements the [build block](https://spec.polkadot.network/sect-block-production#algo-build-block) and [invoke block authoring](https://spec.polkadot.network/sect-block-production#algo-block-production). It uses the babe pre digests created during the block lottery. If one is present for the current slot block production is invoked.


src/main/java/com/limechain/transaction/dto/ApplyExtrinsicResult.java
```diff
@@ -0,0 +1,14 @@
+package com.limechain.transaction.dto;
+
+import lombok.Data;
+import org.jetbrains.annotations.Nullable;
+
+@Data
+public class ApplyExtrinsicResult {
+
+    @Nullable
+    DispatchOutcome outcome;
```
Grigorov-Georgi: Do we need to keep the fields with package-private access modifier here and in ```DispatchOutcome``` class

src/main/java/com/limechain/babe/dto/EpochSlot.java
```diff
@@ -0,0 +1,18 @@
...
+
+import lombok.Value;
+
+import java.io.Serializable;
+import java.math.BigInteger;
+import java.time.Duration;
+import java.time.Instant;
+
+@Value
+public class EpochSlot implements Serializable {
```
Grigorov-Georgi: Can you consider choosing another name for this class because it's misleading whether it provides information for an epoch or a slot? I see the purpose of this class is to give information about both, but the combination of the class + getters/setters (like epochSlot.getStart or epochSlot.getNumber) isn't intuitive because they both (slot and epoch) have start, duration, and number.

Grigorov-Georgi (approved):

src/main/java/com/limechain/babe/dto/EpochSlot.java
Zurcusa: The idea behind this structure is to represent information only related to a babe/epoch slot. If we take into account that a slot is only related to BABE and nothing else we might as well just call it `Slot`.

src/main/java/com/limechain/transaction/dto/ApplyExtrinsicResult.java
Zurcusa: Initially I wanted to use lombok's `@Value`. It seems I forgot to add back the private modifiers. The `DispatchOutcome` is alright since we want to have its fields `private final`, hence the use of `@Value`.

PR #610, opened 2024-11-15
Announce produced block via block announce

### LINKED ISSUE #597: Announce produced block via block announce

osrib:
Announce produced block via block announce

## Checklist:

### DIFF
--- a/src/main/java/com/limechain/babe/BabeService.java
+++ b/src/main/java/com/limechain/babe/BabeService.java
@@ -4,19 +4,31 @@ import com.limechain.babe.coordinator.SlotChangeEvent;
 import com.limechain.babe.coordinator.SlotChangeListener;
 import com.limechain.babe.predigest.BabePreDigest;
 import com.limechain.babe.state.EpochState;
+import com.limechain.exception.scale.ScaleEncodingException;
+import com.limechain.network.Network;
+import com.limechain.network.protocol.blockannounce.messages.BlockAnnounceMessage;
+import com.limechain.network.protocol.blockannounce.scale.BlockAnnounceMessageScaleWriter;
+import com.limechain.network.protocol.warp.dto.BlockHeader;
+import com.limechain.rpc.server.AppBean;
+import com.limechain.storage.block.BlockState;
 import com.limechain.storage.crypto.KeyStore;
+import io.emeraldpay.polkaj.scale.ScaleCodecWriter;
 import org.apache.commons.collections4.map.HashedMap;
 import org.springframework.stereotype.Component;
 
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.math.BigInteger;
 import java.util.Map;
 
 @Component
 public class BabeService implements SlotChangeListener {
 
+    private final BlockState blockState = BlockState.getInstance();
     private final EpochState epochState;
     private final KeyStore keyStore;
     private final Map<BigInteger, BabePreDigest> slotToPreRuntimeDigest = new HashedMap<>();
+    final Network network = AppBean.getBean(Network.class);
 
     public BabeService(EpochState epochState, KeyStore keyStore) {
         this.epochState = epochState;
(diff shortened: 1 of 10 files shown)


src/main/java/com/limechain/network/protocol/blockannounce/BlockAnnounceService.java
```diff
@@ -19,4 +24,20 @@ public void sendHandshake(Host us, PeerId peer) {
...
+     * @param us     our host object
+     * @param peerId message receiver
+     */
+    public void sendBlockAnnounceMessage(Host us, PeerId peerId, byte[] encodedBlockAnnounceMessage) {
+        Optional.ofNullable(connectionManager.getPeerInfo(peerId))
+                .map(p -> p.getBlockAnnounceStreams().getInitiator())
+                .ifPresentOrElse(
+                        stream -> new BlockAnnounceController(stream).sendBlockAnnounceMessage(encodedBlockAnnounceMessage),
+                        () -> sendHandshake(us, peerId)
+                );
```
georg-getz, line 41: Don't we want the `.map` to happen in the `.isPresent` clause?

src/main/java/com/limechain/network/protocol/blockannounce/BlockAnnounceService.java
```diff
@@ -1,12 +1,17 @@
...
 import com.limechain.network.protocol.NetworkService;
 import io.libp2p.core.Host;
 import io.libp2p.core.PeerId;
 import lombok.extern.java.Log;
 
+import java.util.Optional;
+
 @Log
 public class BlockAnnounceService extends NetworkService<BlockAnnounce> {
+    ConnectionManager connectionManager = ConnectionManager.getInstance();
```
georg-getz, line 13: Could be private

src/main/java/com/limechain/babe/BabeService.java
```diff
@@ -4,19 +4,31 @@
...
 import java.util.Map;
 
 @Component
 public class BabeService implements SlotChangeListener {
 
+    private final BlockState blockState = BlockState.getInstance();
     private final EpochState epochState;
     private final KeyStore keyStore;
     private final Map<BigInteger, BabePreDigest> slotToPreRuntimeDigest = new HashedMap<>();
+    final Network network = AppBean.getBean(Network.class);
```
georg-getz: Should be private

georg-getz:
You should also fix the 3 new issues from sonar cloud.

src/main/java/com/limechain/network/protocol/blockannounce/BlockAnnounceService.java
osrib: There is no need to check isPresent() before .map(), because .map() is only invoked if the Optional is present.

georg-getz (approved):

src/main/java/com/limechain/babe/BabeService.java
```diff
@@ -4,19 +4,31 @@
...
 import java.util.Map;
 
 @Component
 public class BabeService implements SlotChangeListener {
 
+    private final BlockState blockState = BlockState.getInstance();
     private final EpochState epochState;
     private final KeyStore keyStore;
     private final Map<BigInteger, BabePreDigest> slotToPreRuntimeDigest = new HashedMap<>();
+    private final Network network = AppBean.getBean(Network.class);
```
Zurcusa, line 31: We should keep all the bean and singleton inits inside of the constructor. Regarding empty collection initialization... I don't have a strict opinion.
`Network` class is a bean as far as I remember. We don't need to use `AppBean`.

src/main/java/com/limechain/babe/BabeService.java
```diff
@@ -43,4 +55,21 @@ public void slotChanged(SlotChangeEvent event) {
             executeEpochLottery(nextEpochIndex);
         }
     }
+
+    public byte[] createEncodedBlockAnnounceMessage(BlockHeader blockHeader) {
```
Zurcusa, line 59: Imo this logic and the logic to "announce" the block itself belong in the BlockHandler.

Zurcusa, line 59: Imo this logic and the logic to "announce" the block itself belong in the BlockHandler.

src/main/java/com/limechain/network/Network.java
```diff
@@ -313,4 +313,11 @@ public void sendMessagesToPeers() {
     public void sendNeighbourMessage(PeerId peerId) {
         grandpaService.sendNeighbourMessage(this.host, peerId);
     }
+
+    public void sendBlockAnnounceMessage(byte[] encodedBlockAnnounceMessage) {
```
Zurcusa, line 317: The business logic here is incorrect. `kademliaService.getBootNodePeerIds()` returns the boot nodes, whereas we need the current connected peers.

src/main/java/com/limechain/network/Network.java
```diff
@@ -313,4 +313,11 @@ public void sendMessagesToPeers() {
...
+        kademliaService.getBootNodePeerIds()
+                .stream()
+                .distinct()
+                .forEach(p -> new Thread(() -> blockAnnounceService.sendBlockAnnounceMessage(this.host, p, encodedBlockAnnounceMessage)).start());
```
Zurcusa, line 321: For asynchronous logic, I'd use the `AsyncExecutor`.

PR #617, opened 2024-11-20
Maintain Transaction Pool

### LINKED ISSUE #615: Maintain Transaction Pool
https://spec.polkadot.network/chap-state#algo-maintain-transaction-pool

Grigorov-Georgi:
# Description

- Rename ```TransactionValidator``` to ```TransactionProcessor``` because it does more than just to validate.
- Refactor duplicated code for handling transactions by extracting it into methods in the ```TransactionProcessor``` class.
- Add ```maintainTransactionPool``` method, which iterates through all transactions in the pool, validates each one individually, and attempts to add them to the queue. If a transaction fails validation, that means it's invalid -> it should be dropped. If validation succeeds, the transaction is added to the queue and removed from the pool. In the both cases we remove the transaction from the pool.

Fixes https://github.com/LimeChain/Fruzhin/issues/615


genesis/ksmcc3.json
Zurcusa, line 1: Why are the genesis files updated in this PR?

src/main/java/com/limechain/transaction/TransactionProcessor.java
```diff
@@ -0,0 +1,168 @@
...
+
+            try {
+                processTransaction(current, peerId);
+            } catch (TransactionValidationException e) {
+                log.fine("Error when validating transaction " + current.toString()
+                        + " from protocol: " + e.getMessage());
+                continue;
+            }
+
+            maintainTransactionPool();
```
Zurcusa: We miss out on possible “ready for queue transactions” that are submitted via the RPC api. Imagine that we don’t receive any transactions via a protocol message for a few blocks. At the same time we have a transaction submitted via the RPC api. This transaction could be ready for the queue from the start, but it would be stuck in the pool until we receive a protocol message. I suggest we invoke the `maintainTransactionPool` after each newly received block from the block announce. In that way the runtime we have a constant cleanup in the pool that also covers transactions already executed in the newly received block.

src/main/java/com/limechain/rpc/subscriptions/author/AuthorRpcImpl.java
```diff
@@ -35,7 +41,19 @@ public AuthorRpcImpl(String forwardNodeAddress, AuthorRPC authorRPC) {
...
+                )
+        );
+
+        try {
+            transactionProcessor.handlePoolOnlyExternalTransaction(decodedExtrinsic);
+        } catch (TransactionValidationException e) {
+            throw new ExecutionFailedException("Failed to executed submit_extrinsic call: " + e.getMessage());
+        }
+
         rpcClient.send(SubscriptionName.AUTHOR_SUBMIT_AND_WATCH_EXTRINSIC.getValue(),
```
Zurcusa, line 57: It seems that our implementation doesn't cover past the "submit" part of the RPC call. Currently we receive an extrinsic and create a WS subscription on the `author_extrinsicUpdate` topic. Afterwards no following updates are sent. As [per spec](https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-6.md#189-author_submitandwatchextrinsic-pubsub)  we should notify the user on the different states of the extrinsic lifetime. See the usage of `Topic.CHAIN_ALL_HEAD` for example.

Grigorov-Georgi: You're correct, but currently, we are not supporting all of the states provided in the provided link, so I decided to return the responses from the smaldot. So if you don't mind, I would suggest sticking with the current solution and changing it at a later stage.

src/main/java/com/limechain/transaction/TransactionProcessor.java
```diff
@@ -0,0 +1,179 @@
...
+
+                // If no validity error object is present, the extrinsic is removed from the pool and added to the queue.
+                var validTransaction = new ValidTransaction(extrinsic, response.getValidity());
+                if (transactionState.shouldAddToQueue(validTransaction)) {
+                    transactionState.removeExtrinsicFromPool(extrinsic);
+                    transactionState.pushTransaction(validTransaction);
+                }
+
+            } catch (TransactionValidationException e) {
+                // In an exception is thrown, no action is taken because maintainTransactionPool is invoked frequently,
```
georg-getz: "If" instead of "In"

Zurcusa (approved):

PR #626, opened 2024-11-26
625 epoch switch

Zurcusa:
# Description
Added functionality to switch epochs on the first slot of the epoch. This is taken care of by the `SlotCoordinator`. 
Consensus messages are read and `EpochState` is updated if a consensus digest is present in an imported block inside of the `BlockHandler`


src/main/java/com/limechain/storage/block/BlockHandler.java
```diff
@@ -34,10 +37,9 @@ public BlockHandler(ProtocolRequester requester, RuntimeBuilder builder) {
         asyncExecutor = AsyncExecutor.withPoolSize(10);
     }
 
-    public synchronized void handleBlock(Instant arrivalTime, BlockHeader header) {
+    public synchronized void handleBlockNoBody(Instant arrivalTime, BlockHeader header) {
```
georg-getz: This function name could be `handleBlockHeader` and the other one could become `handleBlock`

georg-getz (approved):

Grigorov-Georgi (approved):

PR #630, opened 2024-11-27
624 propagate block

### LINKED ISSUE #624: Propagate created block to peers.
When our node produces a block we need to propagate it to all our connected peers. That happens after [import and validate block](https://spec.polkadot.network/chap-sync#algo-import-and-validate-block) in the case of a produced and received block.

Zurcusa:
# Description

This PR extracts the functionalities of sending peer messages and handshakes in a separate class. This further isolates the `Network` class's responsibilities to solely manage the network and connected peers. The PR aims to implement broadcasting functionalities for newly produced or received block announcement protocol blocks.


src/main/java/com/limechain/network/protocol/message/ProtocolMessageBuilder.java
```diff
@@ -0,0 +1,34 @@
...
+    public NeighbourMessage buildNeighbourMessage() {
+        return new NeighbourMessage(
+                NEIGHBOUR_MESSAGE_VERSION,
+                syncState.getLatestRound(),
+                syncState.getSetId(),
+                syncState.getLastFinalizedBlockNumber()
+        );
+    }
+
+    public static BlockAnnounceMessage buildBlockAnnounceMessage(BlockHeader blockHeader, boolean isBestBlock) {
```
Grigorov-Georgi: Can we make this method non-static to be consistent with the method above?

Grigorov-Georgi (approved):

Grigorov-Georgi:
Can you also check the Gradle Build process because it's failing?

src/main/java/com/limechain/network/protocol/message/ProtocolMessageBuilder.java
Zurcusa: I made the whole class static.

Zurcusa:
> Can you also check the Gradle Build process because it's failing?

It was due to a timeout when downloading dependencies.

PR #634, opened 2024-12-02
Broadcast Transactions

### LINKED ISSUE #616: Broadcast[Propagate] transactions to other peers
https://spec.polkadot.network/chap-state#algo-validate-transactions

Grigorov-Georgi:
# Description

- When a new transaction is received from another peer or from a RPC request with the propagate flag TRUE, it is broadcasted to other connected peers.
- Transaction handshakes are sent only when running as an authoring node.
- The ignoredPeers list for transactions is updated when the same transaction is received from multiple peers.

Fixes https://github.com/LimeChain/Fruzhin/issues/616


src/main/java/com/limechain/network/PeerMessageCoordinator.java
```diff
@@ -55,6 +61,20 @@ public void sendBlockAnnounceMessageExcludingPeer(BlockAnnounceMessage message,
...
+
+        sendMessageToActivePeers(p -> {
+            if (excludingSet.contains(p)) {
+                return;
+            }
+            asyncExecutor.executeAndForget(() -> network.getTransactionsService().sendTransactionsMessage(
+                    network.getHost(), p, scaleMessage
+            ));
+        });
+    }
```
georg-getz: Can you add a short javadoc type of description to the function like in the `ConnectionManager` class

Grigorov-Georgi: Added

src/main/java/com/limechain/transaction/TransactionState.java
```diff
@@ -125,4 +126,16 @@ public void removeExtrinsicFromPool(Extrinsic extrinsic) {
     public byte[] addToPool(ValidTransaction validTransaction) {
         return transactionPool.insert(validTransaction);
     }
+
+    public void addAndReturnIgnoredPeers(Extrinsic extrinsic, PeerId peerId) {
+        if (existsInQueue(extrinsic)) {
+            for (ValidTransaction validTransaction : transactionQueue) {
+                if (validTransaction.getExtrinsic().equals(extrinsic)) {
+                    validTransaction.getIgnore().add(peerId);
```
georg-getz: Can you change the name of `ignore` in the `ValidTransaction` class to something a bit more informative like `peersToIgnore` or smth similar

Grigorov-Georgi: I created getter [getIgnoredPeers] and setter [addPeerToIgnore] for ```ignore``` field, because it is used by the runtime and renaming it results in in INALID_TRANSACTION.

georg-getz (approved):

src/main/java/com/limechain/transaction/TransactionProcessor.java
```diff
@@ -118,12 +121,22 @@ private byte[] processTransaction(Extrinsic extrinsic, PeerId peerId) {
...
 
-        return transactionState.shouldAddToQueue(validTransaction)
+        var result = transactionState.shouldAddToQueue(validTransaction)
                 ? transactionState.pushTransaction(validTransaction)
                 : transactionState.addToPool(validTransaction);
+
+        if (response.getValidity() != null && response.getValidity().getPropagate()) {
+            messageCoordinator.sendTransactionMessageExcludingPeer(
+                    validTransaction.getExtrinsic(),
+                    validTransaction.getPeersToIgnore()
```
Zurcusa, line 131: Nothing to change for this PR. I just want to leave a comment regarding `peersToIgnore`. Imo not the best solution to use. It would be better to keep track of all incoming messages inside of each peer in some way instead. In this way we can have a centralized place to keep track of all messages received by peers (block announce, transaction, etc).

src/main/java/com/limechain/transaction/TransactionProcessor.java
```diff
@@ -118,12 +121,22 @@ private byte[] processTransaction(Extrinsic extrinsic, PeerId peerId) {
             return transactionState.addToPool(validTransaction);
         }
 
-        return transactionState.shouldAddToQueue(validTransaction)
+        var result = transactionState.shouldAddToQueue(validTransaction)
```
Zurcusa: `var result` is a bit confusing. Does the method return the extrinsic bytes that were added to the pool/queue or the hex value that we return in the rpc request?

Zurcusa (approved):

src/main/java/com/limechain/transaction/TransactionProcessor.java
Grigorov-Georgi: TransactionState methods for adding a transaction to pool/queue return blake2b hash of the extrinsic.

PR #638, opened 2024-12-02
631 state transactions

### LINKED ISSUE #631: Implement trie transactions.
Some changes to the state trie need to be executed in a transaction so that they can be either committed or rolled back afterwards. We currently do not support that.

Zurcusa:
# Description

With newer runtimes block execution is certain to use transaction. We were missing the implementation for those. This PR implements start, commit and roll-back for state transaction.

### DIFF
--- a/src/main/java/com/limechain/storage/block/BlockHandler.java
+++ b/src/main/java/com/limechain/storage/block/BlockHandler.java
@@ -53,9 +53,17 @@ public class BlockHandler {
 
     public synchronized void handleBlockHeader(Instant arrivalTime, BlockHeader header, PeerId excluding) {
         try {
+            if (epochState.isInitialized() && !verifier.verifyAuthorship(header,
+                    epochState.getCurrentEpochData(),
+                    epochState.getCurrentEpochDescriptor(),
+                    epochState.getCurrentEpochIndex(),
+                    epochState.getCurrentSlotNumber())) {
+                log.fine("Block No: " + header.getBlockNumber() + " with hash: " + header.getHash()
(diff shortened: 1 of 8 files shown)


src/main/java/com/limechain/storage/block/BlockHandler.java
```diff
@@ -53,9 +53,17 @@ public BlockHandler(EpochState epochState,
 
     public synchronized void handleBlockHeader(Instant arrivalTime, BlockHeader header, PeerId excluding) {
         try {
+            if (epochState.isInitialized() && !verifier.verifyAuthorship(header,
```
Grigorov-Georgi, line 56: We initialize EpochState only when running an authoring node. Does that mean this method should not be called when running a full node, because of this epochState.isInitialized() check?

Grigorov-Georgi (approved):

georg-getz (approved):

src/main/java/com/limechain/storage/block/BlockHandler.java
Zurcusa: As [per spec](https://spec.polkadot.network/sect-block-production#algo-verify-authorship-right) The verification happens during `Import-and-Validate-Block`. That usually happens when we produce a block (applying the extrinsics, and finalizing the block) or when we handle announced blocks (executing the block). In the case of production we already suppose that we are allowed to build a block in that slot so no need to do it. However, during block announce we need to validate the right of authorship of every single one. This means that it should run in a full node as well. For that to happen we'd need some refactoring. However, my opinion is that we're good enough for now.

PR #640, opened 2024-12-04
Implemented Block equivocation without unit tests

### LINKED ISSUE #632: Implement Block equivocation

osrib:
Implemented Block equivocation via creating map per slot to keep information about produced blocks buy authorities in given slot.
## Checklist:

### DIFF
--- a/src/main/java/com/limechain/babe/BlockProductionVerifier.java
+++ b/src/main/java/com/limechain/babe/BlockProductionVerifier.java
@@ -18,7 +18,6 @@ import com.limechain.runtime.hostapi.dto.Key;
 import com.limechain.runtime.hostapi.dto.VerifySignature;
 import com.limechain.utils.LittleEndianUtils;
 import com.limechain.utils.Sr25519Utils;
-import com.limechain.utils.StringUtils;
 import io.emeraldpay.polkaj.merlin.TranscriptData;
 import io.emeraldpay.polkaj.schnorrkel.Schnorrkel;
 import io.emeraldpay.polkaj.schnorrkel.VrfOutputAndProof;
@@ -69,11 +68,6 @@ public class BlockProductionVerifier implements SlotChangeListener {
         // TODO Key should be available before the start of the method so that we avoid duplicate code.
         byte[] authorityPublicKey = getAuthority(authorities, (int) preDigest.getAuthorityIndex())
                 .getPublicKey();
-
-        if (isBlockEquivocationExist(authorityPublicKey, blockHeader, runtime, preDigest.getSlotNumber())) {
-            return false;
-        }
-
         byte[] signatureData = sealDigest.getMessage();
         VerifySignature signature = new VerifySignature(
                 signatureData, blockHeader.getBlake2bHash(false), authorityPublicKey, Key.SR25519);
@@ -82,6 +76,10 @@ public class BlockProductionVerifier implements SlotChangeListener {
             log.warning(String.format("Signature of block No: %s with hash %s is not valid",
                     blockHeader.getBlockNumber(),
                     blockHeader.getHash()));
+            return false;
+        }
+
+        if (isBlockEquivocationExist(authorityPublicKey, blockHeader, runtime, preDigest.getSlotNumber())) {
             return false;
         }
 
@@ -196,17 +194,20 @@ public class BlockProductionVerifier implements SlotChangeListener {
         String hexPublicKey = HexUtils.toHexString(authorityPublicKey);
         if (currentSlotAuthorBlockMap.containsKey(hexPublicKey)) {
             BlockHeader firstBlockHeader = currentSlotAuthorBlockMap.get(hexPublicKey);
-            if (!firstBlockHeader.equals(blockHeader)) {
+            if (!firstBlockHeader.getHash().equals(blockHeader.getHash())) {
                 BlockEquivocationProof blockEquivocationProof = new BlockEquivocationProof();
                 blockEquivocationProof.setPublicKey(authorityPublicKey);
                 blockEquivocationProof.setSlotNumber(currentSlotNumber);
                 blockEquivocationProof.setFirstBlockHeader(firstBlockHeader);
                 blockEquivocationProof.setSecondBlockHeader(blockHeader);
 
-                Optional<OpaqueKeyOwnershipProof> opaqueKeyOwnershipProof = runtime.generateKeyOwnershipProof(currentSlotNumber, authorityPublicKey);
+                Optional<OpaqueKeyOwnershipProof> opaqueKeyOwnershipProof = runtime.generateKeyOwnershipProof(
+                        currentSlotNumber, authorityPublicKey);
                 opaqueKeyOwnershipProof.ifPresentOrElse(
                         key -> runtime.submitReportEquivocationUnsignedExtrinsic(blockEquivocationProof, key.getProof()),
-                        () -> log.warning(String.format("Unable to generate Opaque Key Ownership Proof for authority: %s", hexPublicKey)));
+                        () -> log.warning(String.format(
+                                "Failure to report equivocation for authority: %s. Authorship verification marked as failure.",
+                                hexPublicKey)));
                 return true;
             }
         }
--- a/src/main/java/com/limechain/babe/BlockProductionVerifier.java
+++ b/src/main/java/com/limechain/babe/BlockProductionVerifier.java
@@ -1,5 +1,9 @@
 package com.limechain.babe;
 
+import com.limechain.babe.api.BlockEquivocationProof;
+import com.limechain.babe.api.OpaqueKeyOwnershipProof;
+import com.limechain.babe.coordinator.SlotChangeEvent;
+import com.limechain.babe.coordinator.SlotChangeListener;
 import com.limechain.babe.predigest.BabePreDigest;
 import com.limechain.babe.state.EpochData;
 import com.limechain.babe.state.EpochDescriptor;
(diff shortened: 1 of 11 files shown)


georg-getz (approved):

src/main/java/com/limechain/babe/BlockProductionVerifier.java
```diff
@@ -39,19 +55,45 @@ public boolean verifyAuthorship(BlockHeader blockHeader, EpochData currentEpochD
...
+        }
+
+        return isVerified;
+    }
+
+    private boolean checkBlockEquivocation(Authority authority,
+                                           BlockHeader blockHeader,
+                                           Runtime runtime,
+                                           BigInteger currentSlotNumber) {
+        String hexPublicKey = HexUtils.toHexString(authority.getPublicKey());
```
Grigorov-Georgi: You are using HexUtils.toHexString() and StringUtils.toHexWithPrefix() for creating the keys of the map. The difference is the prefix before the hex, please choose one of them and use it consistently. 
Screenshot 2024-12-04 at 23 39 05

src/main/java/com/limechain/babe/BlockProductionVerifier.java
```diff
@@ -39,19 +55,45 @@ public boolean verifyAuthorship(BlockHeader blockHeader, EpochData currentEpochD
...
+                verifySlotWinner(authorityIndex, authorities, currentEpochIndex, randomness, descriptor, currentSlotNumber, vrfOutputAndProof);
+
+        if (isVerified) {
+            currentSlotAuthorBlockMap.put(StringUtils.toHexWithPrefix(verifyingAuthority.getPublicKey()),blockHeader);
+        }
+
+        return isVerified;
+    }
+
+    private boolean checkBlockEquivocation(Authority authority,
```
Grigorov-Georgi: That method name and its return values are a bit confusing to me. If I’m checking for equivocation and it exists, I feel it’s more appropriate to return true, not false. IMPO The word "check" suggests that the method will perform some action if the check fails (throwing an exception) and not return a boolean value. A name like detectBlockEquivocation or checkBlockEquivocationExists seems better, as they imply that the method will return true if equivocation is detected, and false otherwise. This would also make the check in the verifyAuthorship() method more intuitive -> if (detectBlockEquivocation(...)) { return false }

src/main/java/com/limechain/babe/BlockProductionVerifier.java
```diff
@@ -39,19 +55,45 @@ public boolean verifyAuthorship(BlockHeader blockHeader, EpochData currentEpochD
...
+                                           BlockHeader blockHeader,
+                                           Runtime runtime,
+                                           BigInteger currentSlotNumber) {
+        String hexPublicKey = HexUtils.toHexString(authority.getPublicKey());
+        if (currentSlotAuthorBlockMap.containsKey(hexPublicKey)) {
+            BlockHeader firstBlockHeader = currentSlotAuthorBlockMap.get(hexPublicKey);
+            if (!firstBlockHeader.equals(blockHeader)) {
```
Grigorov-Georgi: Please, override the equals() method for BlockHeader, because currently it uses Object equals() method, which return true if we are comparing one object instance to itself.
Screenshot 2024-12-05 at 0 02 48

Grigorov-Georgi (changes requested):

src/main/java/com/limechain/babe/BlockProductionVerifier.java
Zurcusa: s

src/main/java/com/limechain/babe/BlockProductionVerifier.java
```diff
@@ -39,19 +55,45 @@ public boolean verifyAuthorship(BlockHeader blockHeader, EpochData currentEpochD
...
+                BlockEquivocationProof blockEquivocationProof = new BlockEquivocationProof();
+                blockEquivocationProof.setPublicKey(authority.getPublicKey());
+                blockEquivocationProof.setSlotNumber(currentSlotNumber);
+                blockEquivocationProof.setFirstBlockHeader(firstBlockHeader);
+                blockEquivocationProof.setSecondBlockHeader(blockHeader);
+
+                Optional<OpaqueKeyOwnershipProof> opaqueKeyOwnershipProof = runtime.generateKeyOwnershipProof(currentSlotNumber, authority.getPublicKey());
```
Zurcusa: We should try and keep our code lines shorter than the suggested length limit (vertical line in intellij). This will help improve code readability on different monitors and get us closer to a uniform code styling.

Zurcusa:
Resolve conflicts from `dev` branch.

src/main/java/com/limechain/babe/BlockProductionVerifier.java
Grigorov-Georgi: I just realized that the ```firstBlockHeader.equals(blockHeader)``` check might not be necessary after all. You could directly use ```firstBlock.getHash().equals(blockHeader.getHash())``` instead. Please feel free to choose the approach you think works best.

Grigorov-Georgi (approved):

src/main/java/com/limechain/babe/BlockProductionVerifier.java
```diff
@@ -58,6 +69,11 @@ public boolean isAuthorshipValid(BlockHeader blockHeader,
         // TODO Key should be available before the start of the method so that we avoid duplicate code.
         byte[] authorityPublicKey = getAuthority(authorities, (int) preDigest.getAuthorityIndex())
                 .getPublicKey();
+
+        if (isBlockEquivocationExist(authorityPublicKey, blockHeader, runtime, preDigest.getSlotNumber())) {
```
Zurcusa: I have a feeling this should be done last. If the author doesn't have the right to create a block at a given slot it's a waste of time checking for equivocation. Especially considering that it involves calls to runtime and additional scale encoding/decoding.

src/main/java/com/limechain/babe/BlockProductionVerifier.java
```diff
@@ -171,4 +188,33 @@ private Authority getAuthority(List<Authority> authorities, int index) {
...
     }
+
+    private boolean isBlockEquivocationExist(byte[] authorityPublicKey,
+                                           BlockHeader blockHeader,
+                                           Runtime runtime,
+                                           BigInteger currentSlotNumber) {
+        String hexPublicKey = HexUtils.toHexString(authorityPublicKey);
+        if (currentSlotAuthorBlockMap.containsKey(hexPublicKey)) {
+            BlockHeader firstBlockHeader = currentSlotAuthorBlockMap.get(hexPublicKey);
+            if (!firstBlockHeader.equals(blockHeader)) {
```
Zurcusa: Imo we should use the block hash as an equals check. No need for additional complexity.

src/main/java/com/limechain/babe/api/BlockEquivocationProof.java
Zurcusa, line 1: Fields can be private here.

src/main/java/com/limechain/babe/api/OpaqueKeyOwnershipProof.java
Zurcusa, line 1: Here as well.

src/main/java/com/limechain/runtime/RuntimeImpl.java
```diff
@@ -57,6 +62,29 @@ public BabeApiConfiguration getBabeApiConfiguration() {
...
+        byte[] encodedProof = ArrayUtils.addAll(ScaleUtils.Encode.encode(
+                new UInt64Writer(), slotNumber), authorityPublicKey);
+        byte[] encodedResponse = call(RuntimeEndpoint.BABE_API_GENERATE_KEY_OWNERSHIP_PROOF, encodedProof);
+        return new ScaleCodecReader(encodedResponse).readOptional(new OpaqueKeyOwnershipProofReader());
+    }
+
+    @Override
+    public void submitReportEquivocationUnsignedExtrinsic(BlockEquivocationProof blockEquivocationProof,
+                                                          byte[] keyOwnershipProof) {
+        // Key ownership needs to be encoded as a list (have its length encoded) as it is of varying size.
```
Zurcusa:
```suggestion
```

src/main/java/com/limechain/babe/BlockProductionVerifier.java
```diff
@@ -171,4 +188,33 @@ private Authority getAuthority(List<Authority> authorities, int index) {
...
+                BlockEquivocationProof blockEquivocationProof = new BlockEquivocationProof();
+                blockEquivocationProof.setPublicKey(authorityPublicKey);
+                blockEquivocationProof.setSlotNumber(currentSlotNumber);
+                blockEquivocationProof.setFirstBlockHeader(firstBlockHeader);
+                blockEquivocationProof.setSecondBlockHeader(blockHeader);
+
+                Optional<OpaqueKeyOwnershipProof> opaqueKeyOwnershipProof = runtime.generateKeyOwnershipProof(currentSlotNumber, authorityPublicKey);
+                opaqueKeyOwnershipProof.ifPresentOrElse(
+                        key -> runtime.submitReportEquivocationUnsignedExtrinsic(blockEquivocationProof, key.getProof()),
+                        () -> log.warning(String.format("Unable to generate Opaque Key Ownership Proof for authority: %s", hexPublicKey)));
```
Zurcusa:
```suggestion
                        () -> log.warning(String.format("Failure to report equivocation for authority: %s. Authorship verification marked as failure.", hexPublicKey)));
```

I feel like we should note that authorship verification fails even if the equivocation doesn't get reported.

src/main/java/com/limechain/runtime/RuntimeImpl.java
```diff
@@ -57,6 +62,29 @@ public BabeApiConfiguration getBabeApiConfiguration() {
...
+        try (ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+             ScaleCodecWriter scaleCodecWriter = new ScaleCodecWriter(buffer)) {
```
Zurcusa, line 78: I have a feeling it would be a good idea to have some kind of an encoding util method that can concatenate encoded bytes to one another. In this way we can have chains of writers for a single result. This is just and idea nothing to change here.

Zurcusa (approved):

PR #641, opened 2024-12-05
fix: Fix block authorship verification

### LINKED ISSUE #639: Fix block authorship verification.
Currently the [verification](https://spec.polkadot.network/sect-block-production#algo-verify-authorship-right) does not work. As it is a crucial part of the [block import](https://spec.polkadot.network/chap-sync#sect-block-validation) and is executed on each imported block we need it working.

Zurcusa:
# Description

The current implementation only supported verifying a primary block. This PR implements verification for primary, secondary VRF and secondary plain blocks.

### DIFF
--- a/src/main/java/com/limechain/exception/misc/AuthorshipVerificationException.java
+++ b/src/main/java/com/limechain/exception/misc/AuthorshipVerificationException.java
@@ -1,0 +1,7 @@
+package com.limechain.exception.misc;
+
+public class AuthorshipVerificationException extends RuntimeException {
+    public AuthorshipVerificationException(String message) {
+        super(message);
+    }
+}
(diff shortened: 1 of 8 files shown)


Grigorov-Georgi (approved):

georg-getz (approved):

src/main/java/com/limechain/exception/misc/AuthorshipVerificationException.java
```diff
@@ -0,0 +1,7 @@
+package com.limechain.exception.misc;
+
+public class AuthorshipVerificationException extends RuntimeException {
```
osrib, line 3: Where do we handle this type of exception?

Zurcusa: It's handled as a general `Exception` in `BlockHandler`. Since multiple exceptions can be thrown from methods in `handleBlockHeader()` I treat them equally as we currently handle a failure the same way from any kind of exception in that method.

PR #663, opened 2024-12-19
Implemented functionality for reading grandpa consensus message

### LINKED ISSUE #652: Implement functionality for reading grandpa consensus message

osrib:
Implemented functionality for reading grandpa consensus message

## Checklist:

### DIFF
--- a/src/main/java/com/limechain/network/protocol/grandpa/messages/consensus/GrandpaConsensusMessage.java
+++ b/src/main/java/com/limechain/network/protocol/grandpa/messages/consensus/GrandpaConsensusMessage.java
@@ -1,0 +1,16 @@
+package com.limechain.network.protocol.grandpa.messages.consensus;
+
+import com.limechain.chain.lightsyncstate.Authority;
+import lombok.Data;
+
+import java.math.BigInteger;
+import java.util.List;
+
+@Data
+public class GrandpaConsensusMessage {
+    private long bestFinalizedBlock;
+    private List<Authority> authorities;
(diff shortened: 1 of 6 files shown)


src/main/java/com/limechain/network/protocol/grandpa/messages/consensus/GrandpaConsensusMessage.java
```diff
@@ -0,0 +1,16 @@
...
+
+import com.limechain.chain.lightsyncstate.Authority;
+import lombok.Data;
+
+import java.math.BigInteger;
+import java.util.List;
+
+@Data
+public class GrandpaConsensusMessage {
+    private long bestFinalizedBlock;
```
Grigorov-Georgi: It would be more appropriate to rename the current ```lastFinalizedBlock``` field to ```forcedChangeBlockNumber ``` or ```governanceBlockNumber```. This suggestion is based on the polkadot spec definition of the ```m``` field in the gradnpa consensus message. Additionally, it would be better to change its type from long to BigInteger to ensure consistency with the type used for other block number fields.

Zurcusa: I also think we should rename this as it is explained differently in the [spec](https://spec.polkadot.network/sect-finality#sect-finality-forced-changes). Maybe something like `delayStartBlockNumber`. Using BigInteger would also be good.

Zurcusa (approved):

PR #679, opened 2025-01-09
Implemented Best PreVote Candidate algorithm

### LINKED ISSUE #397: Best PreVote Candidate
Implement the algorithm.

osrib:
Implemented Best PreVote Candidate algorithm.
https://spec.polkadot.network/sect-finality#algo-best-prevote-candidate

## Checklist:

### DIFF
--- a/src/main/java/com/limechain/grandpa/GrandpaService.java
+++ b/src/main/java/com/limechain/grandpa/GrandpaService.java
@@ -2,7 +2,9 @@ package com.limechain.grandpa;
 
 import com.limechain.exception.grandpa.GhostExecutionException;
 import com.limechain.exception.storage.BlockStorageGenericException;
-import com.limechain.grandpa.state.RoundState;
+import com.limechain.grandpa.state.GrandpaRound;
+import com.limechain.grandpa.state.GrandpaSetState;
+import com.limechain.network.protocol.grandpa.messages.catchup.res.SignedVote;
 import com.limechain.network.protocol.grandpa.messages.commit.Vote;
 import com.limechain.network.protocol.grandpa.messages.vote.SignedMessage;
 import com.limechain.network.protocol.grandpa.messages.vote.Subround;
@@ -16,19 +18,21 @@ import org.springframework.stereotype.Component;
 
 import java.math.BigInteger;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 @Log
 @Component
 public class GrandpaService {
 
-    private final RoundState roundState;
+    private final GrandpaSetState grandpaSetState;
     private final BlockState blockState;
 
-    public GrandpaService(RoundState roundState, BlockState blockState) {
-        this.roundState = roundState;
+    public GrandpaService(GrandpaSetState grandpaSetState, BlockState blockState) {
+        this.grandpaSetState = grandpaSetState;
         this.blockState = blockState;
     }
 
@@ -40,16 +44,16 @@ public class GrandpaService {
      *
      * @return the best final candidate block
      */
-    public Vote getBestFinalCandidate() {
+    public Vote getBestFinalCandidate(GrandpaRound grandpaRound) {
 
-        Vote prevoteCandidate = getGrandpaGhost();
+        Vote prevoteCandidate = getGrandpaGhost(grandpaRound);
 
-        if (roundState.getRoundNumber().equals(BigInteger.ZERO)) {
+        if (grandpaRound.getRoundNumber().equals(BigInteger.ZERO)) {
             return prevoteCandidate;
         }
 
-        var threshold = roundState.getThreshold();
-        Map<Hash256, BigInteger> possibleSelectedBlocks = getPossibleSelectedBlocks(threshold, Subround.PRECOMMIT);
+        var threshold = grandpaSetState.getThreshold();
+        Map<Hash256, BigInteger> possibleSelectedBlocks = getPossibleSelectedBlocks(grandpaRound, threshold, Subround.PRECOMMIT);
 
         if (possibleSelectedBlocks.isEmpty()) {
             return prevoteCandidate;
@@ -97,14 +101,14 @@ public class GrandpaService {
      *
      * @return GRANDPA GHOST block as a vote
      */
-    public Vote getGrandpaGhost() {
-        var threshold = roundState.getThreshold();
+    public Vote getGrandpaGhost(GrandpaRound grandpaRound) {
+        var threshold = grandpaSetState.getThreshold();
 
-        if (roundState.getRoundNumber().equals(BigInteger.ZERO)) {
+        if (grandpaRound.getRoundNumber().equals(BigInteger.ZERO)) {
             return getLastFinalizedBlockAsVote();
         }
 
-        Map<Hash256, BigInteger> blocks = getPossibleSelectedBlocks(threshold, Subround.PREVOTE);
+        Map<Hash256, BigInteger> blocks = getPossibleSelectedBlocks(grandpaRound, threshold, Subround.PREVOTE);
 
         if (blocks.isEmpty() || threshold.equals(BigInteger.ZERO)) {
             throw new GhostExecutionException("GHOST not found");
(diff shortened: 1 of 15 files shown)


georg-getz (approved):

src/main/java/com/limechain/grandpa/state/RoundState.java
```diff
@@ -34,6 +35,8 @@ public class RoundState {
     private Map<PubKey, Vote> prevotes = new ConcurrentHashMap<>();
     private Map<PubKey, SignedVote> pvEquivocations = new ConcurrentHashMap<>();
     private Map<PubKey, SignedVote> pcEquivocations = new ConcurrentHashMap<>();
+    //TODO: Figure out if we need to store vote messages in round state
+    private VoteMessage voteMessage;
```
Zurcusa: This field serves the purpose of a primary proposal message from the network or from ourselves if we are the primary authority/voter. However it currently gets updated on every handled vote message in the `GrandpaEngine`.  The correct flow of this would be as follows:

`GrandpaEngine` receives a vote message -> if the message of of type `PREVOTE` or `PRIMARY_PROPOSAL` it gets added to the appropriate collection inside of `RoundState`, same goes for `PRECOMMIT`.

Regarding the TODO. The state currently holds data that belongs to different aspects of grandpa:
- Data related to the set: voters, setId and roundNumber
- Data related to the round: votes and equivocation votes

Imo we should separate this. We can rename this class to `GrandpaSetState` and keep the data related to sets. Create an object that holds round data eg `GrandpaRound` that holds the data related to rounds. The state would hold the current and previous rounds as those are the only ones we need per spec. In this way we can apply a more structured update of sets and rounds:

`GrandpaSetState` will update its `voters`, `setId` and set `roundNumber` to 0 every time we receive an appropriate consensus message from the runtime (similar to BABE).
The current and previous `GrandpaRound` fields will be updated every time we finalize a round in the grandpa play round algo. This way we'd have an appropriate place  to store the votes, equivocation votes and the primary proposal.

src/main/java/com/limechain/grandpa/GrandpaService.java
```diff
@@ -111,6 +113,27 @@ public Vote getGrandpaGhost() {
         return selectBlockWithMostVotes(blocks);
     }
 
+    /**
+     * Determines what block is our pre-voted block for the current round
```
Zurcusa, line 116: This comment explains the pseudo code of the algo from the spec, but does not reflect the code in the method.

src/main/java/com/limechain/grandpa/GrandpaService.java
```diff
@@ -111,6 +113,27 @@ public Vote getGrandpaGhost() {
...
+     * if we receive a vote message from the network with a
+     * block that's greater than or equal to the current pre-voted block
+     * and greater than the best final candidate from the last round, we choose that.
+     * otherwise, we simply choose the head of our chain.
+     *
+     * @return the best pre-voted block
+     */
+    public Vote getBestPreVoteCandidate() {
+        Vote currentVote = getGrandpaGhost();
+        VoteMessage voteMessage = roundState.getVoteMessage();
```
Zurcusa: The getVoteMessage() would almost never return a correct value here. Check my comment under `RoundState`

src/main/java/com/limechain/grandpa/GrandpaService.java
```diff
@@ -111,6 +113,27 @@ public Vote getGrandpaGhost() {
...
+        SignedMessage signedMessage = voteMessage.getMessage();
+
+        if (signedMessage != null && signedMessage.getBlockNumber().compareTo(currentVote.getBlockNumber()) > 0) {
```
Zurcusa: This doesn't correctly check if we have a primary proposal vote for the current round. What's happening now is we invoke GHOST and get the current pre-vote block with most votes (B r pv in spec) and a wrong value for primary proposed block (B in spec).
`signedMessage.getBlockNumber().compareTo(currentVote.getBlockNumber()) > 0`
This would translate to pseudo code as following (considering we have correct B r pv and B prim):
`B > B r pv`.

Zurcusa:
The comments that I left may be outside the scope of this PR. What i consider most important for this PR is the following:
- Separate set data from round data in the `RoundState` 
- Create the `GrandpaRound` objects and place them in the state.
- Change the logic of the Best-PreVote-Candidate method so that it conforms to the spec. For this to happen the primary proposal vote inside of the current `GrandpaRound` object should be used.

What should be done outside this PR:
- Proper vote message handling (based on round type)
- Add logic regarding updating sets, rounds etc.

src/main/java/com/limechain/grandpa/GrandpaService.java
Grigorov-Georgi: To fully align with the specification, you should also add the result of Best_Final_Candidate(r - 1) in the comparison.  
Screenshot 2025-01-09 at 17 43 20
I added collection for holding previous result of BEST_FINAL_CANDIDATE to the RoundState in order to have access to the previous rounds data. This collection should probably also be removed form the RoundState and added to the GrandpaSetState.
```
   //TODO: Refactor if this map is accessed/modified concurrently
   private Map<BigInteger, Vote> bestFinalCandidateArchive = new HashMap<>();
    ...
    public void addBestFinalCandidateToArchive(BigInteger roundNumber, Vote vote) {
        this.bestFinalCandidateArchive.put(roundNumber, vote);
    }

    public Vote getBestFinalCandidateForRound(BigInteger roundNumber) {
        return this.bestFinalCandidateArchive.get(roundNumber);
    }
```

src/main/java/com/limechain/grandpa/state/RoundCache.java
```diff
@@ -0,0 +1,86 @@
...
+// We should add new GrandpaRound objects to this structure if
+// A) We receive new vote for round which is greater with 1 than the newest round that we have in this queue
+// B) When we start new Play-Grandpa
+
+// When we finalize a round we should delete all previous GrandpaRounds
+@Component
+public class RoundCache {
+
+    private static final int CACHE_SET_CAPACITY = 4;
+    private final Map<BigInteger, List<GrandpaRound>> roundMap = new LinkedHashMap<>(
```
Grigorov-Georgi: Have we talked about making RoundCache thread-safe, or there is no need for that?

Grigorov-Georgi (approved):

Grigorov-Georgi:
Also, I made a branch with refactored GrandpaService methods, where I use RoundCache -> 398-refactoring-grandpa-methods. You don't need to refactor them; I can merge them to your branch when you are ready.

src/main/java/com/limechain/grandpa/state/GrandpaRound.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private GrandpaRound previous;
+    private BigInteger roundNumber;
+
+    private Vote preVotedBlock;
+    private Vote bestFinalCandidate;
+
+    private Map<Hash256, SignedVote> preVotes = new ConcurrentHashMap<>();
+    private Map<Hash256, SignedVote> preCommits = new ConcurrentHashMap<>();
+    private Map<Hash256, SignedVote> primaryProposals = new ConcurrentHashMap<>();
```
Zurcusa: The primary proposal can be only one per round. I suggest we either treat it as a prevote as it is counted the same way other prevotes are or we have a single `SignedVote` field.

src/main/java/com/limechain/grandpa/state/RoundCache.java
```diff
@@ -0,0 +1,86 @@
...
+            CACHE_SET_CAPACITY, 0.75f, true) {
```
Zurcusa: What do the 2nd and 3rd parameters do here?

src/main/java/com/limechain/grandpa/state/RoundCache.java
Zurcusa: From the comment above the class I have a feeling a thread save structure would be needed.

Zurcusa (approved):

Zurcusa:
There are some new sonar issues. Please have a look.

src/main/java/com/limechain/grandpa/state/GrandpaRound.java
Zurcusa: Come to think of it I'd include it in the prevote collection so that we don't have to refactor methods that rely on counting prevotes.

PR #686, opened 2025-01-14
Scheduled and Forced Authority Set Change Handling

### LINKED ISSUE #401: Forced authority set changes detection and handling
Implement logic behind the handling and detection of forced authority set changes.

Grigorov-Georgi:
# Description

- Change the place of the AuthoritySetChanges PriorityQueue and its methods for adding and handling from WarpSyncState to RoundState. I assume that the RoundState would be renamed to GrandpaSetState or something similar.
- Merge incrementSetId and resetRound methods in RoundState into startNewSet(List<Authority>)
- Create FocedAuthoritySetChange and ScheduledAuthoritySetChange classes which extend AuthoritySetChange
- Add a comparator to AuthoritySetChange to return a ForceAuthoritySetChange object with priority over ScheduledAuthoritySetChange objects
- Change the delay type in the GrandpaConsensusMessage from long to BigInteger
- Handle the additional offset 'm' in ForcedAuthoritySetChange

Fixes https://github.com/LimeChain/Fruzhin/issues/401


src/main/java/com/limechain/sync/warpsync/dto/AuthoritySetChange.java
```diff
@@ -1,18 +1,40 @@
...
+    private Long delay;
+    private BigInteger applicationBlock;
+
+    public AuthoritySetChange(List<Authority> authorities, Long delay, BigInteger announceBlock) {
+        this.authorities = authorities;
+        this.delay = delay;
+        this.applicationBlock = announceBlock.add(BigInteger.valueOf(delay));
+    }
+
+    // ForcedAuthoritySetChange has priority over ScheduledAuthoritySetChanges
```
georg-getz: nit: either both in singular or both in plural

src/main/java/com/limechain/grandpa/state/RoundState.java
```diff
@@ -130,17 +140,74 @@ public void persistState() {
...
+
+        return updated;
+    }
+
+    /**
+     * Handles grandpa consensus message
+     *
+     * @param consensusMessage grandpa consensus message provided by any block header digest
+     */
+    public void handleGrandpaConsensusMessage(GrandpaConsensusMessage consensusMessage, BigInteger currentBlockNumber) {
```
georg-getz: you can delete this javadoc

georg-getz:
We had a [discussion with the team about numeric types about 2 years ago](https://github.com/LimeChain/Fruzhin/issues/85). I think it was mostly because of Java's lack of unsigned values which forced us to save u32 into a `long` and u64 into a `BigInteger`. We then agreed that having all the values as BigInteger, although more memory consuming, is the direction we wanna go in to reduce the need for parsing when comparing and to increase readability. Writing this because you have switched the type of the offset in GRANDPA consensus message.

If you feel like this decision needs revisiting feel free.

src/main/java/com/limechain/sync/warpsync/WarpSyncState.java
```diff
@@ -281,8 +270,13 @@ public void syncNeighbourMessage(NeighbourMessage neighbourMessage, PeerId peerI
     }
 
     private void updateSetData(BigInteger setChangeBlock) {
+
         List<BlockData> response = requester.requestBlockData(
-                BlockRequestField.ALL, setChangeBlock.intValue(), 1).join();
+                BlockRequestField.ALL,
+                setChangeBlock.intValue(),
```
georg-getz: I know you haven't written this but NEVER use `intValue()` over `intValueExact()` - it doesn't check for overflows and always returns a result. If you see it elsewhere in the code change it there as well.

src/main/java/com/limechain/sync/warpsync/dto/AuthoritySetChange.java
Grigorov-Georgi: based

georg-getz (approved):

Zurcusa (approved):

PR #688, opened 2025-01-15
refactor: Add NodeService and ServiceState interfaces

### LINKED ISSUE #684: Add interfaces for services and states and refactor related code.

Zurcusa:
# Description
This PR aims to improve the overall structure of Fruzhin in 2 ways:
- Add a shared interface for services (sync and network, babe and grandpa are to follow) `NodeService`. This enables services to share a class signature and be started/stopped in a similar fashion. It also helps separate what services are required and started for each type of node.
- Add a shared interface for states (block, sync, epoch, grandpa and transaction) `ServiceState`. Reasoning is similar to the service one, but it also aims to centralize the initialization of the states as they were not concise. It also prepares Fruzhin to follow the Runtime to Consensus engine communication. (initializing states at genesis and using consensus message to update)


src/main/java/com/limechain/transaction/TransactionState.java
```diff
@@ -39,7 +36,17 @@ public TransactionState() {
...
+    }
+
+    @Override
+    public void initializeFromDatabase() {
+        //Nothing happens here as we don't store any historical transaction data in DB.
+    }
+
+    @Override
+    public void persistState() {
+        //There's no need to store historical transaction data in DB.
```
Grigorov-Georgi: Not the biggest improvement but you can add default empty implementation to the AbstractState class in order not to add this overrides here.
Screenshot 2025-01-15 at 14 22 47

src/main/java/com/limechain/client/AbstractNode.java
```diff
@@ -1,89 +1,13 @@
 package com.limechain.client;
 
-import com.limechain.cli.CliArguments;
```
Grigorov-Georgi, line 3: With this refactoring, we don't need this class anymore.

src/main/java/com/limechain/config/SystemInfo.java
```diff
@@ -27,7 +27,7 @@ public class SystemInfo {
     private String hostVersion;
     private final BigInteger highestBlock;
 
-    public SystemInfo(HostConfig hostConfig, Network network, SyncState syncState) {
+    public SystemInfo(HostConfig hostConfig, NetworkService network, SyncState syncState) {
```
georg-getz, line 30: Rename to `networkService`

georg-getz: actually you can leave it as is, didn't see that Network class is getting removed

src/main/java/com/limechain/network/NetworkService.java
```diff
@@ -47,12 +51,11 @@
  */
 @Log
 @Getter
-@Component
-public class Network {
+@Service
+public class NetworkService implements NodeService {
```
georg-getz: Curious that you decided to change the `@Component` annotation into a `@Service` one. I think all of our services are annotated with `@Component` so there's probably no reason to break that trend. There is no functional difference between the annotations.

src/main/java/com/limechain/network/NetworkService.java
```diff
@@ -187,13 +190,32 @@ private Ed25519PrivateKey loadPrivateKeyFromDB(KVRepository<String, Object> repo
...
+
+            log.log(Level.INFO, "Waiting for peer connection...");
+            Thread.sleep(10000);
+        }
+
+        log.log(Level.INFO, "Node successfully connected to a peer! Sync can start!");
     }
 
+    @Override
+    @PreDestroy
```
georg-getz, line 218: Explain this `@PreDestroy`. AFAIK it's kind of a destructor to be called right before an object is destroyed which here implies that the `NetworkService` could potentially be destroyed before the `stop` method gets called. Am I missing something?

Grigorov-Georgi (approved):

src/main/java/com/limechain/client/AbstractNode.java
Grigorov-Georgi: I mean AbstractNode, not the one in the import.

src/main/java/com/limechain/sync/SyncMode.java
```diff
@@ -2,5 +2,6 @@
 
 public enum SyncMode {
     WARP,
-    FULL
+    FULL,
+    HEAD
```
georg-getz, line 21: head?

src/main/java/com/limechain/sync/SyncService.java
```diff
@@ -0,0 +1,52 @@
...
+                    default -> throw new IllegalStateException("Unexpected value: " + arguments.syncMode());
+                }
+            }
+            case null -> throw new IllegalStateException("Node role should not be null.");
+            default -> throw new IllegalStateException("Unexpected node role: " + arguments.nodeRole());
+        }
+    }
+
+    @Override
+    @PreDestroy
```
georg-getz, line 48: Same concern as the previous `@PreDestroy`

src/main/java/com/limechain/sync/state/SyncState.java
```diff
@@ -1,44 +1,65 @@
...
-    private final BigInteger startingBlock;
-    private final Hash256 genesisBlockHash;
+    private final BlockState blockState;
+
     private Hash256 lastFinalizedBlockHash;
     private Hash256 stateRoot;
+    private BigInteger lastFinalizedBlockNumber;
+    private BigInteger startingBlock;
+
+    private Hash256 genesisBlockHash;
```
georg-getz, line 36: Why the empty space before `genesisBlockHash`?

src/main/java/com/limechain/sync/state/SyncState.java
```diff
@@ -76,10 +98,9 @@ public void finalizedCommitMessage(CommitMessage commitMessage) {
...
 
-    private static boolean updateBlockState(CommitMessage commitMessage, Block blockByHash) {
+    private boolean updateBlockState(CommitMessage commitMessage, BlockHeader blockHeader) {
         try {
-            BlockState.getInstance().setFinalizedHash(
-                    blockByHash.getHeader(), commitMessage.getRoundNumber(), commitMessage.getSetId());
+            blockState.setFinalizedHash(blockHeader, commitMessage.getRoundNumber(), commitMessage.getSetId());
         } catch (Exception e) {
             log.fine(e.getMessage());
             return false;
```
georg-getz: What exception are we catching here?

src/main/java/com/limechain/sync/warpsync/WarpSyncMachine.java
```diff
@@ -110,16 +112,13 @@ private void finishWarpSync() {
...
 
         log.info("Warp sync finished.");
         this.onFinishCallbacks.forEach(executor::submit);
     }
 
-    public void onFinish(Runnable function) {
-        onFinishCallbacks.add(function);
+    public void onFinish(Runnable... function) {
+        onFinishCallbacks.addAll(List.of(function));
     }
```
georg-getz, line 123: When are we giving this more than 1 runnable?

georg-getz: Also should be `functions`

src/main/java/com/limechain/network/NetworkService.java
Zurcusa: The annotation is an interceptor that tells spring to run the annotated method before the bean is removed from the spring context. In our case inside of `RpcApp` we have this line `this.springCtx.close();`. After testing it it seems that the SIGINT signal does not get handled properly, thus the method inside of `RpcApp` is not invoked. I suggest we leave the annotation and get back to the sigint issue when it becomes a priority.

src/main/java/com/limechain/sync/SyncMode.java
Zurcusa: Head of chain. Other option I thought was TIP as in tip of the chain.

src/main/java/com/limechain/sync/state/SyncState.java
Zurcusa: All other fields are related to the current run of the node, whereas genesis is always the same.

src/main/java/com/limechain/sync/warpsync/WarpSyncMachine.java
Zurcusa: In `SyncService`

`warpSyncMachine.onFinish(() -> AbstractState.setSyncMode(SyncMode.HEAD),
                        messageCoordinator::handshakeBootNodes);`
`warpSyncMachine.onFinish(() -> AbstractState.setSyncMode(SyncMode.FULL),
                                fullSyncMachine::start);`

src/main/java/com/limechain/sync/state/SyncState.java
Zurcusa: Various errors from `BlockState`. Such as `BlockNodeNotFoundException`, `LowerThanRootException`

src/main/java/com/limechain/sync/SyncMode.java
georg-getz: Aren't we always synching to the head of the chain?

src/main/java/com/limechain/sync/state/SyncState.java
georg-getz: Change `Exception` to `RuntimeException` so that it's clear that there are no checked exceptons.

src/main/java/com/limechain/sync/SyncMode.java
Zurcusa: Some logic will depend on the state of the syncing. For example when we receive block announcements we'll have to put them in a list of pending blocks before we finish with syncing. This is in order to avoid a missing parent in the block tree. There are other such example, but this is the first that comes to my head.

georg-getz (approved):

PR #690, opened 2025-01-16
Disable Authoriry

### LINKED ISSUE #405: Disable Validators
Faulty validators should be reported.

Grigorov-Georgi:
# Description

- The responsibility for handling disabled authorities and restricting them from block production or voting, lies within the runtime, not in the host implementation.
- Disabling authorities should not affect threshold calculations or any other computations related to authorities and their weights, as the disabled authorities remain part of the validator set.
- In the past, some blocks were produced by disabled authorities before runtime handling was properly implemented. However, this is no longer the case, as the runtime now handles this scenario.
- Information source: https://app.element.io/#/room/#polkadot-implementers:web3.foundation/$5bXTpPOld-h93QVOZhBVsVzdGcQR5ZyAWu3Di0QsJVU

- Conclusion: The runtime ensures that disabled authorities cannot participate in consensus processes.
- Our responsibility is only to receive and store ```disabled authority``` value

Fixes: https://github.com/LimeChain/Fruzhin/issues/405

### DIFF
--- a/src/main/java/com/limechain/babe/consensus/BabeConsensusMessage.java
+++ b/src/main/java/com/limechain/babe/consensus/BabeConsensusMessage.java
@@ -4,10 +4,12 @@ import com.limechain.babe.state.EpochData;
 import com.limechain.babe.state.EpochDescriptor;
 import lombok.Data;
 
+import java.math.BigInteger;
+
 @Data
 public class BabeConsensusMessage {
     private EpochData nextEpochData;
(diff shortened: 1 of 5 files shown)


Zurcusa:
If the runtime takes care of disabled authorities why do we even need to store them?

georg-getz (approved):

Zurcusa (approved):

PR #691, opened 2025-01-16
687 populate grandpa and babe states from runtime

### LINKED ISSUE #687: Add functionality to populate grandpa and babe states from runtime.
There are 2 types of states:
- Ones that don't communicate with the runtime. (sync and block state)
- Ones that get populated via a runtime call and continue to get updates via consensus messages. (currently babe and grandpa and beefy in the future)

Currently they all share an interface, `ServiceState` that provide a class signature to initialize states. However an interface for the ones that need data from the runtime is missing. In order to properly populate data to the runtime states we should do the following:
1. Add a new interface that would be shared for states that need data from the runtime.
2. Add the functionality to call the runtime method that retrieves **GENESIS** grandpa authorities. (check `getBabeApiConfiguration`)
3. Find an appropriate place to invoke the new interface's method so that both babe and grandpa states get populated from the runtime. (take into account that grandpa method only returns genesis authorities).

hMitov:
# Description

feat: Add ServiceConsensusState interface with populateFromRuntime() method

- Introduced `ServiceConsensusState` interface to define the initializeFromRuntime() method for RoundState and EpochState as requiring runtime data.
- Implemented populateFromRuntime()
- Integrated these state initializations within FullSyncMachine to ensure proper synchronization during node startup.
Fixes #687


src/main/java/com/limechain/ServiceConsensusState.java
```diff
@@ -0,0 +1,14 @@
+package com.limechain;
+
+import com.limechain.runtime.Runtime;
+
+/**
+ * The {ServiceConsensusState} interface defines states that are initialized by runtime data and updated
+ * via consensus messages (e.g., BABE, GRANDPA, Beefy).
```
georg-getz: You can replace "The {ServiceConsensusState}" with "This". BEEFY should be capitalized. You have a coma before 'BABE'.

src/main/java/com/limechain/client/HostNode.java
```diff
@@ -46,6 +46,8 @@ protected void initializeStates() {
         // if: database has some persisted storage
         if (db.find(BlockStateHelper.headerHashKey(BigInteger.ZERO)).isPresent()) {
             states.forEach(ServiceState::initializeFromDatabase);
+            //TODO: should implement the method
+            //    consensusStates.forEach(ServiceState::initializeFromDatabase);
```
georg-getz: Is this blocked or?

hMitov: We arranged that it will not be part of this task, so i removed the comment from the code.

src/main/java/com/limechain/runtime/RuntimeBuilder.java
```diff
@@ -63,7 +63,7 @@ public Runtime copyRuntime(Runtime original) {
      * @param trieAccessor provides access to the trie storage for a given block
      * @return a ready to execute `Runtime` instance
      */
-    public Runtime buildRuntime(byte[] code, @Nullable TrieAccessor trieAccessor) {
+    private Runtime buildRuntime(byte[] code, @Nullable TrieAccessor trieAccessor) {
```
georg-getz, line 66: That `@Nullable` seems misleading. Is there really a case in which trieAccessor is null?

hMitov: Yes, there is. In WarpSyncState, we build runtime from WASM bytecode. The method, responsible for this, calls the upper method, mentioned in the comment, with null value for trieAccessor parameter.

src/main/java/com/limechain/runtime/RuntimeBuilder.java
```diff
@@ -88,4 +88,11 @@ public Runtime buildRuntime(byte[] code, @Nullable TrieAccessor trieAccessor) {
 
         return RuntimeFactory.buildRuntime(code, cfg);
     }
+
+    public Runtime buildRuntimeFromState(TrieAccessor trieAccessor) {
+        return trieAccessor
+                .findStorageValue(Nibbles.fromBytes(":code".getBytes()))
+                .map(wasm -> buildRuntime(wasm, trieAccessor))
+                .orElseThrow(() -> new RuntimeException("Runtime code not found in the trie"));
```
georg-getz, line 96: Change this to one of our custom exceptions or create 1 if there isn't a good match. We want to generally avoid using generic exceptions such as this one.

src/main/java/com/limechain/runtime/Runtime.java
```diff
@@ -58,4 +59,7 @@ public interface Runtime {
     void persistsChanges();
 
     void close();
+
+    List<Authority> getGrandpaApiGrandpaAuthorities();
```
Zurcusa, line 63: I know this is the name of the API method, but I'd go for something like `getGrandpaApiAuthorities` or `getGrandpaAuthorities`. Preferably the first one so it's similar to the babe API one.

src/main/java/com/limechain/runtime/RuntimeImpl.java
```diff
@@ -268,5 +275,11 @@ private byte[] callInner(RuntimeEndpoint function, RuntimePointerSize parameterP
     private Optional<byte[]> findStorageValue(Nibbles key) {
         return this.context.trieAccessor.findStorageValue(key);
     }
+
+    @Override
+    public List<Authority> getGrandpaApiGrandpaAuthorities() {
+        return ScaleUtils.Decode.decode(call(RuntimeEndpoint.GRANDPA_API_GRANDPA_AUTHORITIES), new ListReader<>(new AuthorityReader()));
```
Zurcusa, line 281: Split this into two lines as it is longer than the suggested by Intellij.

src/main/java/com/limechain/grandpa/state/RoundState.java
```diff
@@ -56,9 +58,10 @@ public class RoundState extends AbstractState {
     private Map<PubKey, SignedVote> pcEquivocations = new ConcurrentHashMap<>();
 
     @Override
-    public void initialize() {
-        initialized = true;
-        //TODO: Find a way to initiate a runtime instance during node startup to populate from genesis.
+    public void populateDataFromRuntime(Runtime runtime) {
+        this.authorities = runtime.getGrandpaApiGrandpaAuthorities();
+        this.setId = BigInteger.ZERO;
+        this.roundNumber = BigInteger.ZERO;
```
Zurcusa, line 64: As [per spec](https://spec.polkadot.network/sect-finality#algo-initiate-grandpa) the round at initialization should be 1.

src/main/java/com/limechain/sync/fullsync/FullSyncMachine.java
Zurcusa, line 1: This is not strictly related to this PR, but we need something like a base state to hold all the states. This way, we can have a centralized place to keep them and avoid injecting them as beans all over the place. The constructors are getting ridiculously large.

src/main/java/com/limechain/sync/fullsync/FullSyncMachine.java
```diff
@@ -67,6 +68,8 @@ public class FullSyncMachine {
     private final TrieStorage trieStorage = AppBean.getBean(TrieStorage.class);
     private final RuntimeBuilder runtimeBuilder = AppBean.getBean(RuntimeBuilder.class);
     private final EpochState epochState = AppBean.getBean(EpochState.class);
+    private final RoundState roundState = AppBean.getBean(RoundState.class);
+    private final GenesisBlockHash genesisBlockHash = AppBean.getBean(GenesisBlockHash.class);
```
Zurcusa, line 72: We don't need the `GenesisBlockHash` here. We have `SyncState` which holds the last finalized block number. We can compare it with 0 (genesis block number) and avoid this bean.

PR #702, opened 2025-01-20
Fix grandpa initialization issues

osrib:
Fix grandpa initialization issues

## Checklist:

### DIFF
--- a/src/main/java/com/limechain/grandpa/state/GrandpaSetState.java
+++ b/src/main/java/com/limechain/grandpa/state/GrandpaSetState.java
@@ -9,7 +9,6 @@ import com.limechain.network.protocol.grandpa.messages.consensus.GrandpaConsensu
 import com.limechain.network.protocol.grandpa.messages.vote.SignedMessage;
 import com.limechain.network.protocol.grandpa.messages.vote.Subround;
 import com.limechain.network.protocol.grandpa.messages.vote.VoteMessage;
-import com.limechain.rpc.server.AppBean;
 import com.limechain.storage.DBConstants;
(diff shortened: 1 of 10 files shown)


src/main/java/com/limechain/grandpa/state/GrandpaSetState.java
```diff
@@ -48,13 +48,13 @@ public class GrandpaSetState {
     private List<Authority> authorities;
     private BigInteger disabledAuthority;
     private BigInteger setId;
-    private RoundCache roundCache;
+    private final RoundCache roundCache;
 
     private final PriorityQueue<AuthoritySetChange> authoritySetChanges = new PriorityQueue<>(AuthoritySetChange.getComparator());
 
+     @PostConstruct
```
Zurcusa: I don't think we want this to execute for every node start. We should remove this annotation.

src/main/java/com/limechain/grandpa/state/GrandpaSetState.java
```diff
@@ -48,13 +48,13 @@ public class GrandpaSetState {
     private List<Authority> authorities;
     private BigInteger disabledAuthority;
     private BigInteger setId;
-    private RoundCache roundCache;
+    private final RoundCache roundCache;
```
Zurcusa: Imo it's better to group the injected beans together and leave the remaining fields as they are.

Zurcusa:
Node still doesn't start. Last log we get is `log.log(Level.INFO, "Finished with warp sync!");`.

Zurcusa (approved):

PR #703, opened 2025-01-21
399 Attempt to finalize at round

### LINKED ISSUE #399: Attempt to Finalize At Round
Implement the algorithm as well as any underlying functions.

Grigorov-Georgi:
# Description

- Implement Attempt-To-Finalize-At-Round by the polkadot spec definition -> https://spec.polkadot.network/sect-finality#algo-attempt-to%E2%80%93finalize
- Rename ```getBestFinalCandidate```, ```getGrandpaGhost``` and ```getBestPreVoteCandidate``` to ```findBestFinalCandidate```, ```findGrandpaGhost``` and ```findBestPreVoteCandidate``` in the ```GrandpaService``` in order not to have different names from the ```GrandpaRound``` getters
- The ```GrandpaGhost``` and ```BestFinalCandidate``` fields of the ```GrandpaRound``` object are now being set directly in the methods ```findGrandpaGhost()``` and ```findBestFinalCandidate()```, rather than in ```isFinalizable()``` 
- Add ```commitMessagesArchive```, a map of ```SetId``` as a key, and a list of commit messages as value. On every new set start the commit messages for (setId - 2) are deleted because they will probably never be used again.
- Updating the last finalized block from commit messages is now executed only when we (as a node) aren't part of the voters set.
- Make all methods in GrandpaService private

Fixes: https://github.com/LimeChain/Fruzhin/issues/399

### DIFF
--- a/src/main/java/com/limechain/grandpa/GrandpaService.java
+++ b/src/main/java/com/limechain/grandpa/GrandpaService.java
@@ -54,7 +54,7 @@ public class GrandpaService {
             BlockHeader header = blockState.getHeader(bestFinalCandidate.getBlockHash());
             blockState.setFinalizedHash(header, grandpaRound.getRoundNumber(), grandpaSetState.getSetId());
 
-            if (grandpaSetState.isCommitMessageInArchive(bestFinalCandidate)) {
+            if (!grandpaSetState.isCommitMessageInArchive(bestFinalCandidate)) {
                 //TODO: broadcast
             }
         }
--- a/src/main/java/com/limechain/grandpa/GrandpaService.java
+++ b/src/main/java/com/limechain/grandpa/GrandpaService.java
@@ -48,10 +48,7 @@ public class GrandpaService {
         long totalVoters = grandpaSetState.getAuthorities().size();
         long threshold = (2 * totalVoters) / 3;
 
-        // L - last finalized block
-        // E - best final candidate
-        // The spec defines E >= L, but here E > L ensures E is new and not already received.
-        if (bestFinalCandidate.getBlockNumber().compareTo(lastFinalizedBlockNumber) > 0 &&
+        if (bestFinalCandidate.getBlockNumber().compareTo(lastFinalizedBlockNumber) >= 0 &&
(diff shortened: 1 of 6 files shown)


src/main/java/com/limechain/grandpa/GrandpaService.java
```diff
@@ -31,6 +31,35 @@ public GrandpaService(GrandpaSetState grandpaSetState, BlockState blockState) {
...
+        }
+
+        var bestFinalCandidateVotesCount = getObservedVotesForBlock(
+                grandpaRound,
+                bestFinalCandidate.getBlockHash(),
+                Subround.PRECOMMIT
+        );
+
+        long totalVoters = grandpaSetState.getAuthorities().size();
+        long threshold = (2 * totalVoters) / 3;
```
georg-getz: Can you put a comment with a link showing how these numbers were chosen or explain them in he comment, because they look like magic numbers right now.

src/main/java/com/limechain/grandpa/state/GrandpaSetState.java
```diff
@@ -230,4 +241,23 @@ public void handleVoteMessage(VoteMessage voteMessage) {
...
+    }
+
+    public void addCommitMessageToArchive(CommitMessage message) {
+        commitMessagesArchive.putIfAbsent(setId, new ArrayList<>());
+        List<CommitMessage> commitMessages = commitMessagesArchive.get(setId);
+        commitMessages.add(message);
+    }
+
+    public void cleanCommitMessagesArchive() {
+        commitMessagesArchive.remove(setId.subtract(BigInteger.TWO));
```
georg-getz: Why are we subtracting 2 here?

Grigorov-Georgi: The only reason is that we will only need the commit messages for the current set and in the edge case also the commit messages from the previous set. By subtracting 2 I delete commit messages for older sets. I will add a comment about the reason behind that action because this is not defined in the spec.

Zurcusa (approved):

PR #710, opened 2025-01-23
Refactor startNewSet method and persist Round/Set state

### LINKED ISSUE #712: Refactor Initiate Granpda
1. Initiate-grandpa
This should be invoked every time a new set is started. What currently happens after our flow is:
SetId is incremented by 1 - This is okay.
Authorities are updated - This is okay.
We create a round with number 0 with no preset values for preVotedBlock and bestFinalCandidate - This is wrong. In reality we need to create a round 0, set the fields based on the current finalized block in our BlockState. It acts as a dummy round that has preset values. Then we create round with number 1. Its previous round will be round 0. preVotedBlock and bestFinalCandidate will be null at start and will be set at some point. 

2. We don't persist round and sets when they get removed from the cache.
When the round cache goes above the threshold of 4 we need to persist the removed set and all it's rounds to the db.
When we remove all older rounds than the one finalized we need to store them as well.

Grigorov-Georgi:
# Description

- The ```startNewSet``` method initializes a round with number 0, where ```preVotedBlock``` and ```bestFinalCandidate``` are set to the ```lastFinalizedBlock```. This round is directly persisted, followed by the creation of round 1, which acts as functional start of the GRANDPA set.
- Each other round (except the round with the number 0) is persisted when the ```attemptToFinalizeAt``` method passes all of the checks for finalizing a block.

Fixes: https://github.com/LimeChain/Fruzhin/issues/712

### DIFF
--- a/src/main/java/com/limechain/storage/block/BlockState.java
+++ b/src/main/java/com/limechain/storage/block/BlockState.java
@@ -7,6 +7,7 @@ import com.limechain.exception.storage.BlockStorageGenericException;
 import com.limechain.exception.storage.HeaderNotFoundException;
 import com.limechain.exception.storage.LowerThanRootException;
 import com.limechain.exception.storage.RoundAndSetIdNotFoundException;
+import com.limechain.network.protocol.grandpa.messages.commit.Vote;
 import com.limechain.network.protocol.warp.dto.Block;
 import com.limechain.network.protocol.warp.dto.BlockBody;
 import com.limechain.network.protocol.warp.dto.BlockHeader;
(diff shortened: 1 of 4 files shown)


src/main/java/com/limechain/storage/block/BlockState.java
```diff
@@ -893,6 +894,15 @@ public Pair<BigInteger, BigInteger> getHighestRoundAndSetID() {
         return helper.bytesToRoundAndSetId(data);
     }
 
+    public Vote getLastFinalizedBlockAsVote() {
```
Zurcusa, line 897: I have a feeling this would be more appropriate in the grandpa package.

Zurcusa (approved):

PR #711, opened 2025-01-24
402 catchup request sending

### LINKED ISSUE #402: Catchup request sending
Implement logic for sending catchup requests.

hMitov:
# Description
Implemented logic for sending catchup requests.

**When sending Catch-Up Requests?**

When a neighbor message is received, the following conditions must be checked to determine whether to issue a catch-up request:
**1. Peer Node setID Validation**
       Ensure the setID of the peer node matches the setID of the current round in our node. If they do not match, ignore the 
       peer.

**2. Round Number Validation**
        **Per spec:** Check if the roundNumber of the neighbor is greater than the current roundNumber of our node.

Fixes #402 >

### DIFF
--- a/src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
+++ b/src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
@@ -3,11 +3,13 @@ package com.limechain.network.protocol.grandpa;
 import com.limechain.exception.scale.ScaleEncodingException;
 import com.limechain.grandpa.GrandpaService;
 import com.limechain.grandpa.state.GrandpaSetState;
+import com.limechain.grandpa.state.RoundCache;
 import com.limechain.network.ConnectionManager;
 import com.limechain.network.protocol.blockannounce.messages.BlockAnnounceHandshakeBuilder;
 import com.limechain.network.protocol.grandpa.messages.GrandpaMessageType;
 import com.limechain.network.protocol.grandpa.messages.catchup.req.CatchUpReqMessage;
 import com.limechain.network.protocol.grandpa.messages.catchup.req.CatchUpReqMessageScaleReader;
+import com.limechain.network.protocol.grandpa.messages.catchup.req.CatchUpReqMessageScaleWriter;
 import com.limechain.network.protocol.grandpa.messages.catchup.res.CatchUpMessage;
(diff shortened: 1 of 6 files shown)


src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
```diff
@@ -134,11 +137,27 @@ private void handleHandshake(byte[] message, PeerId peerId, Stream stream) {
...
         log.log(Level.FINE, "Received neighbour message from Peer " + peerId + "\n" + neighbourMessage);
         new Thread(() -> warpSyncState.syncNeighbourMessage(neighbourMessage, peerId)).start();
+
+        // If peer has the same voter set id
+        if (neighbourMessage.getSetId().equals(grandpaSetState.getSetId())) {
+            sendCatchUpRequest(neighbourMessage, peerId, stream);
+        }
+    }
+
+    private void sendCatchUpRequest(NeighbourMessage neighbourMessage, PeerId peerId, Stream stream) {
```
Grigorov-Georgi: IMO You should move the ```if (neighbourMessage.getSetId().equals(grandpaSetState.getSetId()))``` check into the ```sendCatchUpRequest``` method and rename it to ```handleCatchUpBasedOnNeighbourMessage``` or similar, as the method is more about checking if we should send catch-up message than just sending a message, which ```writeCatchUpRequest``` does.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
```diff
@@ -139,6 +141,10 @@ private void handleNeighbourMessage(byte[] message, PeerId peerId) {
         NeighbourMessage neighbourMessage = reader.read(NeighbourMessageScaleReader.getInstance());
         log.log(Level.FINE, "Received neighbour message from Peer " + peerId + "\n" + neighbourMessage);
         new Thread(() -> warpSyncState.syncNeighbourMessage(neighbourMessage, peerId)).start();
+
+        if (NodeRole.AUTHORING.getValue().equals(connectionManager.getPeerInfo(peerId).getNodeRole())) {
```
Zurcusa, line 145: We can add a //TODO here for the check if we're a voter once implemented.

Zurcusa (approved):

PR #714, opened 2025-01-24
Add isActiveAuthority and GrandpaKeyPair fields to AbstractState

### LINKED ISSUE #713: Check and store if we are a grandpa authority for sets.
Currently, the `nodeRole` in `HostConfig` is a static field that stores the type of node we are starting. We need a way to distinguish if we're chosen for the current grandpa authority set as a voter. Alongside that, we need to store the key that we are selected to participate with so that we don't need to look it up in the keystone every time.

Grigorov-Georgi:
# Description

- Refactor keyStore.getKey() to return Pair<publicKey, privateKey>, where the keys are of type ```byte[]``` in order to work for both SR25519(Babe) and ED25519(Grandpa) key pairs.
- Add a method to convert from Pair<byte[], byte[]> to Schnorrkel.KeyPair object
- Add isActiveAuthority and grandpaKeyPair static fields to AbstractState and methods for setting and unsetting them.

Fixes https://github.com/LimeChain/Fruzhin/issues/713


src/main/java/com/limechain/state/AbstractState.java
```diff
@@ -3,13 +3,19 @@
...
+    @Getter
+    protected boolean initialized;
+
     @Getter
     private static SyncMode syncMode;
     @Getter
-    protected boolean initialized;
+    private static boolean isActiveAuthority = false;
+    @Getter
+    private static Pair<byte[], byte[]> grandpaKeyPair = null;
```
georg-getz, line 18: If every single field needs a getter than you should put the annotation on the class

src/main/java/com/limechain/state/AbstractState.java
```diff
@@ -19,6 +25,16 @@ public static void setSyncMode(SyncMode mode) {
...
 
+    public static void setAuthorityStatus(Pair<byte[], byte[]> keyPair) {
+        isActiveAuthority = true;
+        grandpaKeyPair = keyPair;
+    }
+
+    public static void unsetAuthorityStatus() {
+        isActiveAuthority = false;
+        grandpaKeyPair = null;
+    }
```
georg-getz: I like "clear" better than "unset"

src/main/java/com/limechain/state/AbstractState.java
Grigorov-Georgi: "This annotation can also be applied to a class, in which case it'll be as if all non-static fields that don't already have a Getter annotation have the annotation." taken from lombok documentation for Getter. I tried applying it only to the class, but it doesn't generate getters for the static fields.

Zurcusa (approved):

georg-getz (approved):

PR #722, opened 2025-01-28
403 process catchup request

### LINKED ISSUE #403: Process Catchup request
Implement the algorithm as well as any underlying functions.

hMitov:
# Description

- Implemented handling of catch up requests and introduced last finalized round reference to GrandpaSetState as per spec (It will be used in Play Grandpa Round and Process Catch-up Responses, too)
- Added catch up threshold when sending catch up request as per spec
- Extracted handling and sending of catch up requests to GrandpaSetState.java

Fixes #403 >


src/main/java/com/limechain/grandpa/state/GrandpaSetState.java
```diff
@@ -52,13 +58,15 @@
...
     private static final BigInteger THRESHOLD_DENOMINATOR = BigInteger.valueOf(3);
+    private static final BigInteger CATCH_UP_THRESHOLD = BigInteger.TWO;
 
     private List<Authority> authorities;
     private BigInteger disabledAuthority;
     private BigInteger setId;
 
     private final BlockState blockState;
     private final RoundCache roundCache;
+    private GrandpaRound lastFinalizedRound;
```
Grigorov-Georgi: Wouldn't it be easier to store BigInteger lastFinalizedRoundNumber instead of GrandpaRound lastFinalizedRound and then call lastFinalizedRound.getRoundNumber()?

Grigorov-Georgi: The lastFinalizedRound field should also be updated; I suggest setting it right after attemptToFinalizeAt() returns true or directly in the method itself.

src/main/java/com/limechain/sync/warpsync/WarpSyncState.java
Zurcusa, line 1: I don't think these are needed here. We can fully move the logic to the `GrandpaSetState`. The methods here do not bring any extra logic in the process.

src/main/java/com/limechain/network/protocol/grandpa/messages/catchup/res/CatchUpResMessage.java
```diff
@@ -10,7 +11,8 @@
 @Data
 @AllArgsConstructor
 @NoArgsConstructor
-public class CatchUpMessage {
+@Builder(toBuilder = true)
```
Zurcusa: Why do we need the `toBuilder` ?

src/main/java/com/limechain/network/protocol/grandpa/messages/catchup/req/CatchUpReqMessage.java
```diff
@@ -9,6 +10,7 @@
 @Data
 @AllArgsConstructor
 @NoArgsConstructor
+@Builder(toBuilder = true)
```
Zurcusa: Here as well.

Zurcusa (approved):

src/main/java/com/limechain/grandpa/state/GrandpaSetState.java
hMitov: We arranged that lastFinalizedRound will be taken from db

PR #727, opened 2025-01-28
Update justification verification

### LINKED ISSUE #648: Implement functionality for finalizing justification

### LINKED ISSUE #655: Implement functionality for getting the last finalized block

Grigorov-Georgi:
# Description

- JustificationVerifier.verify() method now has Justification as an input parameter and also executes block ancestry validation
- JustificationVerifier.verify() method checks if there are more than 3 votes cast by single authority. The vote with the smaller block number is considered as the valid vote, other two are counted as equivocations. 
- Replace PreCommit with SignedVote in all places and delete PreCommit object

Fixes: https://github.com/LimeChain/Fruzhin/issues/655 & https://github.com/LimeChain/Fruzhin/issues/648


src/main/java/com/limechain/sync/JustificationVerifier.java
```diff
@@ -26,14 +30,23 @@
 @Log
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
```
Zurcusa:
```suggestion
@UtilityClass
```

And then we can remove the static from flag from the methods.

src/main/java/com/limechain/sync/JustificationVerifier.java
```diff
@@ -44,39 +57,57 @@ public static boolean verify(PreCommit[] preCommits, BigInteger round) {
...
                 return false;
             }
 
-            if (seenPublicKeys.contains(preCommit.getAuthorityPublicKey())) {
+            if (seenPublicKeys.contains(signedVote.getAuthorityPublicKey())) {
                 log.log(Level.WARNING, "Duplicated signature");
                 return false;
             }
-            seenPublicKeys.add(preCommit.getAuthorityPublicKey());
+            seenPublicKeys.add(signedVote.getAuthorityPublicKey());
```
Zurcusa: As per [Spec](https://spec.polkadot.network/sect-finality#defn-grandpa-justification):

> A valid justification must only contain up to one valid vote from each voter and must not contain more than two equivocatory votes from each voter.

We can group them up beforehand using https://www.baeldung.com/java-groupingby-collector. Then check if any of the keys have > 3 votes and return false directly instead of checking a seen list.

Zurcusa (approved):

Zurcusa (approved):

PR #742, opened 2025-02-03
feat: Add initial skeleton for grandpa round stages.

### LINKED ISSUE #741: Add a skeleton for the stage workflow of a round.

Zurcusa:
# Description
This PR adds a frame to work on when implementing the different stages of the Grandpa play algorithm. It follows an approach similar to the state design pattern. 

Fixes #741


Grigorov-Georgi (approved):

Zurcusa:
The amount of responsibility inside of `GranpaRound` is growing more than I'd like. It would be a good idea if we move some of it out. The idea behind the class was to be a DTO/entity.

PR #744, opened 2025-02-05
fix: Fix failing light sync snapshot workflow.

Zurcusa:
# Description

Update deprecated `actions/upload-artifact` and  `actions/download-artifact` v3 actions.

### DIFF
--- a/.github/workflows/update-light-snapshot.yml
+++ b/.github/workflows/update-light-snapshot.yml
@@ -39,7 +39,7 @@ jobs:
       - id: get-chain-id  # Reads the `id` field in the newly-downloaded chain spec
         run: echo "id=`jq -r .id ./chain_spec.json`" >> $GITHUB_OUTPUT
       - if: ${{ steps.get-chain-id.outputs.id == '' }}
-        uses: actions/upload-artifact@V4
+        uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-light-snapshot.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - id: get-chain-id  # Reads the `id` field in the newly-downloaded chain spec
         run: echo "id=`jq -r .id ./chain_spec.json`" >> $GITHUB_OUTPUT
       - if: ${{ steps.get-chain-id.outputs.id == '' }}
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
```
georg-getz, line 42: why is this necessary?

Grigorov-Georgi (approved):

georg-getz (approved):

.github/workflows/update-light-snapshot.yml
Zurcusa: After extensive research (multiple hours) I reached the source of truth regarding this problem. 

https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

Then I carefully reviewed what was written and had to discuss it with our teammate @Grigorov-Georgi for a second opinion. We both concluded that this was the best approach.

PR #745, opened 2025-02-05
733 catch up response receiving

### LINKED ISSUE #733: CatchUp response receiving

hMitov:
# Description

- Implemented handling of catch-up responses
- Made changes to the previous logic for collecting preVotes and preCommits on initiating catch-up response
- Placed "Todo" comments in handling catch-up response, describing what should be added as logic, to help those who implement play-grandpa-round and finalizing preVote stage.

Fixes #733 >

### DIFF
--- a/src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
+++ b/src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
@@ -2,6 +2,7 @@ package com.limechain.network.protocol.grandpa;
 
 import com.limechain.chain.lightsyncstate.Authority;
 import com.limechain.exception.grandpa.GrandpaGenericException;
+import com.limechain.exception.sync.JustificationVerificationException;
 import com.limechain.grandpa.GrandpaService;
 import com.limechain.grandpa.round.GrandpaRound;
 import com.limechain.grandpa.round.RoundCache;
@@ -45,11 +46,18 @@ import org.springframework.stereotype.Component;
 
 import java.math.BigInteger;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.BiConsumer;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.logging.Level;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 @Log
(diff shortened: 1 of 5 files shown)


src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
```diff
@@ -313,4 +328,168 @@ private Optional<GrandpaRound> selectRound(BigInteger peerRoundNumber, BigIntege
...
+        List<SignedVote> result = new ArrayList<>();
+
+        Stream<SignedVote> allPreCommits = Stream.concat(
+                requestedRound.getPcEquivocations().values().stream().flatMap(Set::stream),
+                requestedRound.getPreCommits().values().stream()
+        );
+
+        for (SignedVote vote : allPreCommits.toList()) {
+            if (totalWeight.compareTo(threshold) >= 0)
+                break;
```
Grigorov-Georgi: If you are going to omit the brackets, please keep the logic to one line.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
```diff
@@ -313,4 +328,168 @@ private Optional<GrandpaRound> selectRound(BigInteger peerRoundNumber, BigIntege
...
+            totalWeight = validateAndProcessVote(vote, finalizedBlock, totalWeight, result);
+        }
+
+        return result.toArray(SignedVote[]::new);
+    }
+
+    private BigInteger validateAndProcessVote(SignedVote vote,
```
Grigorov-Georgi: Is there a reason to return TotalWeight from this method?

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
```diff
@@ -313,4 +328,168 @@ private Optional<GrandpaRound> selectRound(BigInteger peerRoundNumber, BigIntege
...
+    /**
+     * Handles a catch-up response from a peer, validating pre-votes and pre-commits.
+     * Initializes a new round, determines the Grandpa Ghost and finalization estimate,
+     * checks round compatibility, and executes the Play-Grandpa-Round if valid.
+     *
+     * @param peerId            peer responding with catch-up message
+     * @param catchUpResMessage received catch-up response message
+     * @param peerIds           set of connected peer ids
+     */
+    public void handleCatchUpResponse(PeerId peerId,
```
Grigorov-Georgi, line 270: I prefer structuring classes with public methods first, followed by all other private helper methods at the bottom. We haven't discussed this before, but if we all agree, we can adopt this practice.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
hMitov: ValidateAndProcessVote() calls processVote(), which adds the weight of valid votes to totalWeight.
Without the assignment, the weight would be lost in each iteration.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
hMitov: It is a good idea.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
Grigorov-Georgi: Okay, that makes sense, but IMO it might be better to rename the method to recalculateTotalWeight.

hMitov: Method validateAndProcessVote() does validation, then adds the validated vote and increases the weight. I renamed validateAndProcessVote() to validateAndAccumulateVote() and processVote() to addVoteAndIncreaseWeight().

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
```diff
@@ -294,6 +260,152 @@ public void initiateAndSendCatchUpResponse(PeerId peerId,
...
+        }
+
+        if (!catchUpResMessage.getSetId().equals(grandpaSetState.getSetId())) {
+            throw new GrandpaGenericException("Catch up response has a different setId.");
+        }
+
+        RoundCache roundCache = grandpaSetState.getRoundCache();
+        GrandpaRound latestRound = roundCache.getLatestRound(grandpaSetState.getSetId());
+        if (catchUpResMessage.getRound().compareTo(latestRound.getRoundNumber()) <= 0) {
+            throw new GrandpaGenericException("Catching up into a round in the future.");
```
Zurcusa: I don't think this is correct. If we receive round 5 from catchup and our current is 7 this statement is true, but we're not actually catching up into a round in the future. It's a round in the past.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
```diff
@@ -294,6 +260,152 @@ public void initiateAndSendCatchUpResponse(PeerId peerId,
...
+                                          SignedVote[] votes,
+                                          BiConsumer<GrandpaRound, Map<Hash256, SignedVote>> setUniqueVotes,
+                                          BiConsumer<GrandpaRound, Map<Hash256, Set<SignedVote>>> setEquivocations) {
+
+        // Group votes by AuthorityPublicKey
+        Map<Hash256, List<SignedVote>> voteCount = Arrays.stream(votes)
+                .collect(Collectors.groupingBy(SignedVote::getAuthorityPublicKey));
+
+        Map<Hash256, SignedVote> uniqueVotes = new ConcurrentHashMap<>();
+        Map<Hash256, Set<SignedVote>> equivocations = new ConcurrentHashMap<>();
```
Zurcusa: `SignedVote` has an overriden equals method. Set doesn't allow for repeating entries. I am not familiar how the signature works, but just in case I'd have this as a list here to avoid unexpected behaviour.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
```diff
@@ -294,6 +260,152 @@ public void initiateAndSendCatchUpResponse(PeerId peerId,
...
+        setVotesAndEquivocations(grandpaRound, votes, GrandpaRound::setPreVotes, GrandpaRound::setPvEquivocations);
+    }
+
+    private void setPreCommitsAndPcEquivocations(GrandpaRound grandpaRound, SignedVote[] votes) {
+        setVotesAndEquivocations(grandpaRound, votes, GrandpaRound::setPreCommits, GrandpaRound::setPcEquivocations);
+    }
+
+    private void setVotesAndEquivocations(GrandpaRound grandpaRound,
+                                          SignedVote[] votes,
+                                          BiConsumer<GrandpaRound, Map<Hash256, SignedVote>> setUniqueVotes,
```
Zurcusa, line 333: What's the point of the BiConsumers? Can't we just call `grandpaRound.setPreCommits` and `grandpaRound. setPcEquivocations`?

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
```diff
@@ -294,6 +260,152 @@ public void initiateAndSendCatchUpResponse(PeerId peerId,
...
+        if (commitMessage.getVote().getBlockNumber().compareTo(lastFinalizedBlockNumber) <= 0) {
+            return;
+        }
+        syncState.finalizedCommitMessage(commitMessage);
+
+        new Thread(() -> warpSyncState.updateRuntime(lastFinalizedBlockNumber)).start();
+    }
+
+    private void processNeighbourUpdates(SyncMessage.BlockData block) {
+        BlockHeader header = BlockHeaderReader.getInstance().read(new ScaleCodecReader(block.getHeader().toByteArray()));
```
Zurcusa: `ScaleUtils.Decode.decode(...)`

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
```diff
@@ -294,6 +260,152 @@ public void initiateAndSendCatchUpResponse(PeerId peerId,
...
+
+        stateManager.getSyncState().finalizeHeader(header);
```
Zurcusa, line 393: I am not quire sure this logic is correct. Imo we shouldn't finalize anything based on neighbor messages. Not related to PR, but we can keep this in mind.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
```diff
@@ -313,4 +425,65 @@ private Optional<GrandpaRound> selectRound(BigInteger peerRoundNumber, BigIntege
...
+
+    private SignedVote[] getPreCommitJustification(GrandpaRound requestedRound) {
+        BlockHeader finalizedBlock = requestedRound.getFinalizedBlock();
+        BigInteger totalWeight = BigInteger.ZERO;
+        BigInteger threshold = stateManager.getGrandpaSetState().getThreshold();
+
+        List<SignedVote> result = new ArrayList<>();
+
+        Stream<SignedVote> allPreCommits = Stream.concat(
+                requestedRound.getPcEquivocations().values().stream().flatMap(Set::stream),
```
Zurcusa: Same comment about the set here.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
```diff
@@ -313,4 +425,65 @@ private Optional<GrandpaRound> selectRound(BigInteger peerRoundNumber, BigIntege
...
+        BlockHeader estimate = requestedRound.getBestFinalCandidate();
+        BlockState blockState = stateManager.getBlockState();
+        Hash256 estimateHash = estimate.getHash();
+
+        Predicate<SignedVote> isDescendant = vote ->
+                blockState.isDescendantOf(estimateHash, vote.getVote().getBlockHash());
+
+        return Stream.concat(
+                        requestedRound.getPreVotes().values().stream(),
+                        requestedRound.getPvEquivocations().values().stream().flatMap(Set::stream)
```
Zurcusa: Same comment about set here.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
hMitov: We discussed that we will changed it everywhere to Map<Hash256, List<SignedVote>> as to prevent possible omitting of equivocations.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
hMitov: We arranged that in order to prevent code duplications it will remain with BiConsumer.

Zurcusa (approved):

PR #746, opened 2025-02-05
feat: Implement starting grandpa round.

### LINKED ISSUE #725: Implement initial (starting) phase of voting round.
Line 1-8 in [spec](https://spec.polkadot.network/sect-finality#id-voting-process-in-round-r).

This phase is pretty straightforward:

- Starts when grandpa round starts playing.
- Derive if we're the primary voter.
- Emit commit message.
- Emit primary proposal. (if needed)

Zurcusa:
# Description
Implement the starting grandpa round. It is responsible for sending a primary proposal to other peers if we are chosen as primary voter. Also refactor some code.

Fixes #725


Grigorov-Georgi:
With these changes, it looks to me that RoundCache is no longer useful. Here is my brainstorm on how we can remove it. 

- keeping an object GrandpaRound currentRound in GrandpaSetState
- adding a new round
	- The previous of the previous of the current round will be set to null (currentRound.getPrevious().setPrevious() = null) to make it available for garbage collection
	- create a new round with currentRound.setRoundNumber(currentRound.getRoundNumber() + 1) and set its previous to the currentRound of the GrandpaSetState and then set the currentRound to this newly created one
- getting a Round -> getRound(BigInteger roundNumber)
	- start from the current round and check if its round number is equal to the requested one if not traverse through the available previous rounds. If a result is not found try to get it from the database.
- getLatestRound() -> will be just getCurrentRound()
- getLatestRoundNumber() -> will be currentRound().getRoundNumber()

Zurcusa:
I suggest we keep this refactor for when we implement round handling inside of `GrandpaService`.

Grigorov-Georgi (approved):

Grigorov-Georgi (approved):

PR #748, opened 2025-02-06
feat: add grandpa duraction constant

Grigorov-Georgi:
# Description

Add Grandpa duration constant

### DIFF
--- a/src/main/java/com/limechain/grandpa/round/GrandpaRound.java
+++ b/src/main/java/com/limechain/grandpa/round/GrandpaRound.java
@@ -45,6 +45,9 @@ import java.util.concurrent.ScheduledExecutorService;
 @RequiredArgsConstructor
 public class GrandpaRound {
 
+    // Based on https://github.com/paritytech/polkadot/pull/6217
+    public static final long DURATION = 1000;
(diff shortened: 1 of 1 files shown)


src/main/java/com/limechain/grandpa/round/GrandpaRound.java
```diff
@@ -45,6 +45,9 @@
 @RequiredArgsConstructor
 public class GrandpaRound {
 
+    // Based on https://github.com/paritytech/polkadot/pull/6217
+    public static final long DURATION = 1000;
```
Zurcusa, line 49: Why is this necessary?

Zurcusa (approved):

src/main/java/com/limechain/grandpa/round/GrandpaRound.java
Grigorov-Georgi: We need that constant in play grandpa algorithm for the pre-votes and pre-commit timers.

Zurcusa: Insightful.

PR #749, opened 2025-02-07
728 Implement PreCommit Stage

### LINKED ISSUE #728: Implement pre commit stage.
Line  13 & 14 in [spec](https://spec.polkadot.network/sect-finality#id-voting-process-in-round-r).

Start pre commit:
- Check if round is completable
- Start a timer for Time >= t + 4 * T
- End sub stage if completable and timer passes.

Wait for votes:
- Receive pre votes.
- On each pre vote update GHOST and some other updates try an end the sub round.

End prevote:
- If GHOST has a value do line 14.
- Start next stage.

Grigorov-Georgi:
# Description

- If ```round.isCompletable() = true``` or timer ```Time >= t + 4 * T``` is expires -> end pre-commit stage
- end() method waits ```round.getGrandpaGhost()``` to have a value before proceeding to the next stage
- Once GrandpaGhost is available, a vote message is broadcast
- ```isCompleted``` field is added to ```GrandpaRound``` that is updated, when ```findGrandpaGhost``` is executed and the previous method for checking if the round is completable is renamed to ```checkIfCompletable```

Fixes https://github.com/LimeChain/Fruzhin/issues/728

### DIFF
--- a/src/main/java/com/limechain/grandpa/round/GrandpaRound.java
+++ b/src/main/java/com/limechain/grandpa/round/GrandpaRound.java
@@ -254,7 +254,7 @@ public class GrandpaRound {
      *
      * @return if the current round is completable
      */
-    public boolean checkIfRoundIsCompletable() {
+    public boolean checkIfCompletable() {
 
         Map<Vote, Long> votes = getDirectVotes(SubRound.PRE_COMMIT);
         long votesCount = votes.values().stream()
@@ -370,7 +370,7 @@ public class GrandpaRound {
 
         BlockHeader result = selectBlockWithMostVotes(blocks, getPrevBestFinalCandidate());
         this.grandpaGhost = result;
-        this.isCompletable = checkIfRoundIsCompletable();
+        this.isCompletable = checkIfCompletable();
 
         return result;
     }
--- a/src/main/java/com/limechain/grandpa/round/GrandpaRound.java
+++ b/src/main/java/com/limechain/grandpa/round/GrandpaRound.java
@@ -182,7 +182,7 @@ public class GrandpaRound {
         return Optional.ofNullable(finalizeEstimate).orElse(lastFinalizedBlock);
     }
 
-    private void attemptToFinalizeAt() {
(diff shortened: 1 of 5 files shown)


src/main/java/com/limechain/grandpa/round/GrandpaRound.java
```diff
@@ -362,10 +368,11 @@ private BlockHeader findGrandpaGhost() {
             throw new GhostExecutionException("GHOST not found");
         }
 
-        BlockHeader grandpaGhost = selectBlockWithMostVotes(blocks, getPrevBestFinalCandidate());
-        this.grandpaGhost = grandpaGhost;
+        BlockHeader result = selectBlockWithMostVotes(blocks, getPrevBestFinalCandidate());
+        this.grandpaGhost = result;
+        this.isCompletable = checkIfCompletable();
```
Zurcusa, line 373: I'm not sure this is the appropriate place to update this. We do not always recalculate grandpa ghost. For example if a pre commit comes in we only need to update the estimate (best final candidate). Let's keep this here, but it will probably be a subject of change.

src/main/java/com/limechain/grandpa/round/GrandpaRound.java
```diff
@@ -650,4 +657,11 @@ public void broadcastCommitMessage() {
 
         peerMessageCoordinator.sendCommitMessageToPeers(commitMessage);
     }
+
+    public void clearOnStageTimerHandler() {
```
Zurcusa, line 661: Check other stages in case there is repeating code.

Zurcusa (approved):

src/main/java/com/limechain/grandpa/round/GrandpaRound.java
Grigorov-Georgi: Setting the `isCompletable` field only makes sense when a new Ghost is computed. If there is no new ghost, it would mean that the previous `isCompletable` field is up to date.

PR #750, opened 2025-02-07
feat: Implement Finalize stage in Play-Grandpa-Round.

### LINKED ISSUE #729: Implement finalization stage.
Line 15-26 in [spec](https://spec.polkadot.network/sect-finality#id-voting-process-in-round-r).

Start waiting stage:

- Check for the conditions in line 18.
- Start next sub stage.

Wait for votes:

- Receive pre commits.
- On each pre commit update BEST-FINAL-CANDIDATE and attempt to finalize.

End round:

- Do whatever need to be done as an ending logic.

hMitov:
# Description

- Implemented Finalize stage in Play-Grandpa-Round
- Added callback to GrandpaRound, that should be included in the attemptToFinalize method implementation

Fixes #729

### DIFF
--- a/src/main/java/com/limechain/grandpa/round/FinalizeStage.java
+++ b/src/main/java/com/limechain/grandpa/round/FinalizeStage.java
@@ -11,13 +11,13 @@ public class FinalizeStage implements StageState {
 
         log.fine(String.format("Round %d entered the Finalize stage.", round.getRoundNumber()));
 
-        if (isRoundReadyToEnd(round)) {
+        if (isRoundReadyToFinalized(round)) {
             end(round);
             return;
         }
 
         Runnable onFinalizeHandler = () -> {
-            if (isRoundReadyToEnd(round)) {
+            if (isRoundReadyToFinalized(round)) {
                 end(round);
             }
         };
@@ -34,7 +34,7 @@ public class FinalizeStage implements StageState {
         round.end();
     }
 
-    private boolean isRoundReadyToEnd(GrandpaRound round) {
+    private boolean isRoundReadyToFinalized(GrandpaRound round) {
         BlockHeader finalized = round.getFinalizedBlock();
         BlockHeader prevBestFinalCandidate = round.getPrevBestFinalCandidate();
 
--- a/src/main/java/com/limechain/grandpa/round/FinalizeStage.java
+++ b/src/main/java/com/limechain/grandpa/round/FinalizeStage.java
@@ -17,12 +17,11 @@ public class FinalizeStage implements StageState {
             return;
         }
 
-        Runnable onFinalizeHandler = () -> {
+        round.setOnFinalizeHandler(() -> {
             if (isRoundReadyToBeFinalized(round)) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/limechain/grandpa/round/FinalizeStage.java
```diff
@@ -1,13 +1,46 @@
...
+        round.setOnFinalizeHandler(null);
+        round.attemptToFinalizeAt();
+        log.fine(String.format("Round %d exits Finalize stage.", round.getRoundNumber()));
+
+        round.end();
     }
-}
 
+    private boolean isRoundReadyToEnd(GrandpaRound round) {
+        BlockHeader finalized = round.getFinalizedBlock();
```
Grigorov-Georgi: You should use try-catch instead of null check for this one.
```
    public BlockHeader getFinalizedBlock() {
        if (finalizedBlock == null) throw new GrandpaGenericException("Finalized block has not been set.");
        return finalizedBlock;
    }
```

hMitov: The idea here is that if finalized is null, it means that round is not ready to be finalized. The method should return false(we are in waiting for preCommits as not eligible to finalize) and then the callback should be passed to the GrandpaRound handler. Then on each preCommit message, we attemptToFinalize the round and execute the logic from the callback.

Zurcusa: Return false in the catch if getter throws an error.

Grigorov-Georgi: I probably didn't describe well what I meant, but I guess after @Zurcusa's clarification you get the point.

Grigorov-Georgi (approved):

src/main/java/com/limechain/grandpa/round/FinalizeStage.java
```diff
@@ -1,13 +1,51 @@
...
     public void start(GrandpaRound round) {
+
+        log.fine(String.format("Round %d entered the Finalize stage.", round.getRoundNumber()));
+
+        if (isRoundReadyToBeFinalized(round)) {
+            end(round);
+            return;
+        }
+
+        Runnable onFinalizeHandler = () -> {
```
Zurcusa: I think we have this inline when we invoke `setOnFinalizeHandler`

src/main/java/com/limechain/grandpa/round/FinalizeStage.java
```diff
@@ -1,13 +1,51 @@
...
+        round.end();
     }
-}
 
+    private boolean isRoundReadyToBeFinalized(GrandpaRound round) {
+        BlockHeader finalized;
+        try {
+            finalized = round.getFinalizedBlock();
+        } catch (GrandpaGenericException e) {
+            return false;
```
Zurcusa, line 43: Maybe add a warning log here.

Zurcusa (approved):

PR #760, opened 2025-02-13
feat: Updated vote message handling with the newly introduced logic f…

### LINKED ISSUE #732: Refactor vote message handling to use the new update functionalities.

hMitov:
# Description

- Updated vote message handling with the newly introduced logic for updating ghost and estimate candidates. 
- Removed equivocation searching when subround is primary proposal.
- Added handleVoteMessage calling on broadcasting preVotes and preCommits.
- Refactored handleVoteMessage method.

Fixes #732


src/main/java/com/limechain/grandpa/vote/SubRound.java
Zurcusa, line 1: Did we check if the usages of `SubRound` aren't expecting null somewhere across the project?

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
Zurcusa, line 1: I see we have this piece of code in every message handle (aside from handshake). 
`        if (AbstractState.isActiveAuthority() && connectionManager.checkIfPeerIsAuthorNode(peerId)) {
            grandpaMessageHandler.handleVoteMessage(voteMessage);
        }`

Why don't we extract the logic for all messages (except handshake) + authority check in a method and use it inside of `handleResponderStreamMessage`?

src/main/java/com/limechain/network/protocol/grandpa/GrandpaMessageHandler.java
Zurcusa, line 1: After we merge https://github.com/LimeChain/Fruzhin/pull/758 this has to be reworked a bit so that it uses the `setId` from the round itself.

Grigorov-Georgi: It's merged.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
Zurcusa, line 1: Not related to the PR, but I think we should start adding `@Nullable` annotations in methods that actually do null validations. In that way it will be easier to know when we can expect a null param and also we will not duplicate null checks.

Zurcusa (approved):

Zurcusa (approved):

PR #786, opened 2025-02-20
BEEFY consensus message scale classes

### LINKED ISSUE #771: Scale classes

hMitov:
# Description

- Added BeefyConsensusMessage and BeefyConsensusMessageFormat(enum class) classes
- Implemented scale reading. 
- Introduced beefy consensus message processing in addBlockToTree() method (without implementation of the handler)

Fixes #771

### DIFF
--- a/src/main/java/com/limechain/network/protocol/beefy/messages/consensus/BeefyConsensusMessageReader.java
+++ b/src/main/java/com/limechain/network/protocol/beefy/messages/consensus/BeefyConsensusMessageReader.java
@@ -1,0 +1,45 @@
+package com.limechain.network.protocol.beefy.messages.consensus;
+
+import io.emeraldpay.polkaj.scale.ScaleCodecReader;
+import io.emeraldpay.polkaj.scale.ScaleReader;
+import io.emeraldpay.polkaj.scale.reader.ListReader;
+import io.emeraldpay.polkaj.scale.reader.UInt64Reader;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+import java.util.List;
+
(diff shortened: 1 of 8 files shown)


Grigorov-Georgi (approved):

src/main/java/com/limechain/network/protocol/beefy/messages/consensus/BeefyConsensusMessageReader.java
```diff
@@ -0,0 +1,45 @@
...
+    public BeefyConsensusMessage read(ScaleCodecReader reader) {
+
+        BeefyConsensusMessage beefyConsensusMessage = new BeefyConsensusMessage();
+        BeefyConsensusMessageFormat format = BeefyConsensusMessageFormat.fromFormat(reader.readByte());
+        beefyConsensusMessage.setFormat(format);
+
+        switch (format) {
+            case BEEFY_CHANGED_AUTHORITIES -> {
+                List<byte[]> authorityPublicKeys = new ListReader<>(
+                        rdr -> rdr.readByteArray(33)).read(reader);
```
georg-getz: can you create a constant for that 33, it feels like a magic number right now.

src/main/java/com/limechain/network/protocol/warp/DigestHelper.java
Zurcusa, line 1: Each digest is added separately in the header. This means we can have more than one digest per type. With the current implementation we use `findFirst()` and that would not work. I suggest we rework the `getBabeConsensusMessage`,  `getGrandpaConsensusMessage`, `getBeefyConsensusMessage` so that they return a list of scale decoded messages instead of a single option.

Zurcusa (approved):

PR #804, opened 2025-02-26
feat: Implemented BeefyState

### LINKED ISSUE #777: Beefy state

hMitov:
# Description
Implemented BeefyState class responsible for managing BEEFY finality rounds.
**Added essential fields for tracking:**
- Validator sets per session.
- Mandatory and non-mandatory block numbers for efficient finality.
- BEEFY genesis, finalization state, and next digest tracking.

**Introduced BeefySession class to store:**
- Authority sets for each session.
- A synchronized set of non-mandatory blocks that can be finalized for lower latency.

**P.S.**
Sessions are created by a validator set change in a block - we create a key value from the blockNumber and a session.

Non-mandatory blocks are added when votes occur.
We find the first session whose mandatory block number is greater than voted block and store that block in the previous to that session as a non-mandatory block.

Fixes #777 >

### DIFF
--- a/src/main/java/com/limechain/beefy/dto/SignedCommitment.java
+++ b/src/main/java/com/limechain/beefy/dto/SignedCommitment.java
@@ -1,5 +1,6 @@
 package com.limechain.beefy.dto;
 
+import io.libp2p.core.crypto.PubKey;
 import lombok.Data;
 
 import java.util.List;
@@ -8,5 +9,5 @@ import java.util.Optional;
 @Data
 public class SignedCommitment {
     private Commitment commitment;
-    private List<Optional<byte[]>> signatures;
+    private List<Optional<PubKey>> signatures;
 }
--- a/src/main/java/com/limechain/beefy/dto/SignedCommitment.java
+++ b/src/main/java/com/limechain/beefy/dto/SignedCommitment.java
@@ -1,6 +1,5 @@
 package com.limechain.beefy.dto;
 
-import io.libp2p.core.crypto.PubKey;
 import lombok.Data;
 
 import java.util.List;
@@ -9,5 +8,5 @@ import java.util.Optional;
 @Data
 public class SignedCommitment {
     private Commitment commitment;
-    private List<Optional<PubKey>> signatures;
+    private List<Optional<byte[]>> signatures;
 }
--- a/src/main/java/com/limechain/beefy/dto/SignedCommitment.java
+++ b/src/main/java/com/limechain/beefy/dto/SignedCommitment.java
@@ -1,6 +1,6 @@
 package com.limechain.beefy.dto;
 
-import com.limechain.runtime.hostapi.dto.Signature;
+import io.libp2p.core.crypto.PubKey;
 import lombok.Data;
 
 import java.util.List;
@@ -9,5 +9,5 @@ import java.util.Optional;
 @Data
 public class SignedCommitment {
     private Commitment commitment;
-    private List<Optional<Signature>> signature;
+    private List<Optional<PubKey>> signatures;
 }
--- a/src/main/java/com/limechain/beefy/dto/ValidatorSet.java
+++ b/src/main/java/com/limechain/beefy/dto/ValidatorSet.java
@@ -1,12 +1,16 @@
 package com.limechain.beefy.dto;
 
-import lombok.Value;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
 
 import java.math.BigInteger;
 import java.util.List;
 
-@Value
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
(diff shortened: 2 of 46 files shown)


src/main/java/com/limechain/beefy/dto/ValidatorSet.java
```diff
@@ -1,12 +1,16 @@
...
+import lombok.AllArgsConstructor;
 import lombok.Data;
+import lombok.NoArgsConstructor;
 
+import java.math.BigInteger;
 import java.util.List;
 
 @Data
+@AllArgsConstructor
+@NoArgsConstructor
```
Zurcusa:
```suggestion
@Value
```

Value enforces private final fields. As we receive the validator set from a digest and do not change any data afterwards it is a suitable candidate to enforce this annotation.

src/main/java/com/limechain/beefy/dto/SignedCommitment.java
```diff
@@ -9,5 +9,5 @@
 @Data
 public class SignedCommitment {
     private Commitment commitment;
-    private List<Optional<Signature>> signature;
+    private List<Optional<PubKey>> signatures;
```
Zurcusa:
```suggestion
    private List<Optional<byte[]>> signatures;
```

src/main/java/com/limechain/beefy/dto/VoteMessage.java
```diff
@@ -1,13 +1,11 @@
...
 
-import java.math.BigInteger;
-
 @Data
 public class VoteMessage {
     private Commitment commitment;
-    private BigInteger authorityId;
-    private Signature signature;
+    private byte[] authorityId;
+    private PubKey signature;
```
Zurcusa:
```suggestion
    private byte[] signature;
```

src/main/java/com/limechain/beefy/dto/SignedCommitment.java
Zurcusa: The signature is an array of bytes generated when we sign some data using a private key. The output is usually an array of bytes.

src/main/java/com/limechain/beefy/dto/ValidatorSet.java
```diff
@@ -1,12 +1,16 @@
...
 public class ValidatorSet {
-    private List<PubKey> validators;
-    private long id;
+    private List<byte[]> validators;
+    private BigInteger setId;
```
Zurcusa:
```suggestion
    List<byte[]> validators;
    BigInteger setId;
```

src/main/java/com/limechain/beefy/state/BeefySession.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.math.BigInteger;
+import java.util.Collections;
+import java.util.Set;
+import java.util.TreeSet;
+
+@Data
+public class BeefySession {
+    private ValidatorSet validatorSet;
+    private Set<BigInteger> nonMandatoryBlockNumbers = Collections.synchronizedSortedSet(new TreeSet<>());
```
Zurcusa:
```suggestion
@Value
public class BeefySession {
    ValidatorSet validatorSet;
    Set<BigInteger> nonMandatoryBlockNumbers = Collections.synchronizedSortedSet(new TreeSet<>());
```

src/main/java/com/limechain/beefy/state/BeefyState.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    private List<byte[]> fetchValidators(BigInteger setId) {
+        return repository.find(
+                StateUtil.generateAuthorityKey(DBConstants.AUTHORITY_SET, setId),
+                Collections.emptyList()
+        );
+    }
+
+    private void saveBeefyFinalized() {
```
Zurcusa:
```suggestion
    private void persistBeefyFinalized() {
```

Zurcusa: And for all other `save{}` methods.

src/main/java/com/limechain/storage/DBConstants.java
```diff
@@ -40,4 +40,9 @@ public class DBConstants {
     public static final String SET_ID = "gs::setId";
     public static final String GRANDPA_PREVOTES = "gs:grandpaPreVotes";
     public static final String GRANDPA_PRECOMMITS = "gs:grandpaPreCommits";
+
+    //BeefyState keys
+    public static final String BEEFY_FINALIZED = "bs:beefyFinalized";
+    public static final String BEEFY_ROUND = "bs:beefyRound";
+    public static final String BEEFY_JUSTIFICATION = "bs:justification";
```
Zurcusa, line 49: We should add a separate key for grandpa and beefy authorities.

Zurcusa (approved):

PR #805, opened 2025-02-28
fix: Fix issue with connectivity to new peers.

### LINKED ISSUE #354: Improve peer finding
- [x] Update Nabu dependency to latest version
- [x] Ensure we can keep a two digit number of peers connected and synching at all times.

Zurcusa:
# Description

Our current implementation only relies on the initial connections to boot nodes when communicating with the chain. That is not a scalable solution and does not conform to the spec. This PR aims to look for new peers and send them handshakes in a scheduled manner so that we can establish connections with them. Kademlia's `bootstrap(Host us)` method finds peers close to us and establishes a connection with them. However, we must send handshakes to the newly opened connection, or they remain idle most of the time.

Fixes #354.


Grigorov-Georgi (approved):

Grigorov-Georgi:
I have one general question -> How do you determine the appropriate thread pool size?

Zurcusa:
> I have one general question -> How do you determine the appropriate thread pool size?

I monitored with VisualVM and saw that most of the threads are in a WAITING/PARKED state meaning that they're not being actively used. So I just lowered the number of threads in the pools to avoid spawning too many threads. I am not 100% sure, but maybe we will need to increase the threads that take of sending messages to peers so that we make sure to send as many messages simultaneously as possible.

PR #812, opened 2025-03-04
Adding target vote number calculation

### LINKED ISSUE #783: Round number calculation

osrib:
## Checklist:

### DIFF
--- a/src/main/java/com/limechain/beefy/state/BeefyState.java
+++ b/src/main/java/com/limechain/beefy/state/BeefyState.java
@@ -150,12 +150,13 @@ public class BeefyState extends AbstractState implements ServiceConsensusState {
     }
 
     public void vote() {
+
         // Get the first session (round)
         Map.Entry<BigInteger, BeefySession> sessionStart = sessions.firstEntry();
 
         // If no session is found, exit the method
         if (sessionStart == null) {
-            System.out.println("BEEFY: No voting round started");
+            log.info("Vote BEEFY: No voting round started");
             return;
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/limechain/beefy/state/BeefyState.java
```diff
@@ -144,6 +149,51 @@ public BigInteger getThreshold() {
         return numOfValidators.subtract(faulty);
     }
 
+    public void vote() {
+        // Get the first session (round)
```
Zurcusa, line 154:
```suggestion

        // Get the first session (round)
```

src/main/java/com/limechain/beefy/state/BeefyState.java
```diff
@@ -144,6 +149,51 @@ public BigInteger getThreshold() {
...
+
+        // If the mandatory block (sessionStart) does not have a beefy justification yet, vote on it
+        if (beefyFinalized.compareTo(sessionStartBlock) < 0) {
+            log.info(String.format("BEEFY: vote target - mandatory block: #%s%n", sessionStartBlock));
+            targetVoteBlockNumber = sessionStartBlock;
+        } else {
+            BigInteger diff = grandpaFinalized.subtract(beefyFinalized).max(BigInteger.ZERO).add(BigInteger.ONE);
+            int diffInt = diff.min(BigInteger.valueOf(Integer.MAX_VALUE)).intValue();
+            int nextPowerOfTwo = Integer.highestOneBit(diffInt) << 1;
+            int adjustedDiff = Math.max(MIN_BLOCK_DELTA, nextPowerOfTwo);
```
Zurcusa: I don't understand this part. How is next power of two calculated by shifting a bit?

osrib: So to basically understand if N=5 is power of two we can check if bit representation of 5 have exactly one bit (meaning only one 1 ) 5 is 0101 so two ones in representation but 4 is 0100 exactly one 1 so it is power of two. To find the next power of two if n=13 (0000 1101) we take highest bit which is first 1 -> so (0000 1000) 8 then we shift left by one so we receive (0001 0000) which is 16. So 16 is next power of two for 13.

Zurcusa: What would happen if we have a value of 2 for `diffInt`?

Zurcusa: Ah I saw you updated the code to cover that edge case. Nevermind.

src/main/java/com/limechain/beefy/state/BeefyState.java
Zurcusa, line 1: I suggest we use the name of the method at the start of each log so that we have a reference to where it is from in case of repeating logs.

Zurcusa (approved):

src/main/java/com/limechain/beefy/state/BeefyState.java
osrib: I would add

PR #813, opened 2025-03-05
feat: Implemented fetching and storing of disabled authority for set.

### LINKED ISSUE #774: Store disable authorities

hMitov:
# Description

Implemented fetching and storing of disabled authority for set.

Fixes #774

### DIFF
--- a/src/main/java/com/limechain/beefy/state/BeefyState.java
+++ b/src/main/java/com/limechain/beefy/state/BeefyState.java
@@ -127,7 +127,6 @@ public class BeefyState extends AbstractState implements ServiceConsensusState {
     }
 
     /**
-     * Method is being called after processing all the accumulated justifications.
      * It checks for mandatory blocks by detecting set changes in all blocks
      * between the last BEEFY finalized and GRANDPA finalized blocks.
      */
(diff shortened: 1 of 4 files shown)


src/main/java/com/limechain/beefy/state/BeefyState.java
```diff
@@ -90,6 +91,22 @@ public void persistState() {
         persistRoundNumber(roundNumber);
     }
 
+    /**
+     * The threshold is determined as the numOfValidators - (numOfValidators - 1) / 3
+     *
+     * @return minimum required validators for finality.
+     */
+    public BigInteger getThreshold() {
```
Zurcusa: Imo each set would hold info on validators inside of it. This can either be moved inside of the init of a new round or inside of a `BeefyService` so that we avoid polluting the state class.

src/main/java/com/limechain/beefy/state/BeefyState.java
```diff
@@ -129,19 +146,17 @@ private Pair<BigInteger, ValidatorSet> detectValidatorSetChange(BigInteger maxBl
     }
 
     /**
-     * The threshold is determined as the numOfValidators - (numOfValidators - 1) / 3
-     *
-     * @return minimum required validators for finality.
+     * Method is being called after processing all the accumulated justifications.
```
Zurcusa: Is this comment related to the method below?

Zurcusa (approved):

PR #820, opened 2025-03-07
feat: Initiate grandpa.

### LINKED ISSUE #400: EPIC: Initiate GRANDPA
Join all the other functions into this grand one.

- [x] #654 
- [x] #395

Zurcusa:
# Description

The PR implements the following logic related to GRANDPA:
- Initializing a round at the end of the full sync process.
- Appropriately start rounds on catch up responses.
- Starting next round from within a round when a certain criteria is met.

Currently if we are not a part of the voter set we do not have have the logic to apply incoming justifications. This means that in a set that we are not voting in we would be stagnant in the first round. That is in no way ideal and will be improved in the future.

Fixes #400.

### DIFF
--- a/src/main/java/com/limechain/grandpa/GrandpaService.java
+++ b/src/main/java/com/limechain/grandpa/GrandpaService.java
@@ -1,11 +1,186 @@
 package com.limechain.grandpa;
 
+import com.limechain.exception.grandpa.GrandpaGenericException;
+import com.limechain.grandpa.round.GrandpaRound;
+import com.limechain.grandpa.state.AuthoritySet;
+import com.limechain.grandpa.state.GrandpaSetState;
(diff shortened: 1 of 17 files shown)


src/main/java/com/limechain/grandpa/GrandpaService.java
```diff
@@ -1,11 +1,186 @@
...
 public class GrandpaService {
+
+    private final StateManager stateManager;
+
+    public void start() {
+
+        try {
+            tryStartFromLastFinalizedBlock();
+
+            log.info(String.format("Grandpa service started with round #%d",
```
Grigorov-Georgi, line 34: Add the name of the method to the start of the log message

Grigorov-Georgi (approved):

PR #821, opened 2025-03-07
Handle beefy session key from keystore

### LINKED ISSUE #815: Retrieve and handle beefy session key from keystore.

osrib:
Handle beefy session key from keystore

## Checklist:


src/main/java/com/limechain/beefy/state/BeefyState.java
```diff
@@ -118,6 +120,15 @@ private void handleSessionTransitions(BigInteger grandpaFinalized) {
                 ValidatorSet validatorSet = validatorsSet.getRight();
 
                 BeefySession beefySession = new BeefySession(validatorSet);
+                Optional<org.javatuples.Pair<byte[], byte[]>> keyPair = validatorSet.getValidators().stream()
```
Zurcusa: I think we can extract this logic as a method inside of KeyStore. It would take an `AuthoritySet` object and a `KeyType` and have the same logic as this piece of code. In this way we can reuse it for all consensus algorithms.

src/main/java/com/limechain/beefy/state/BeefySession.java
```diff
@@ -1,16 +1,22 @@
...
+import lombok.Data;
+import org.javatuples.Pair;
 
 import java.math.BigInteger;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
-@Value
+@Data
```
Zurcusa: What's the reason you moved it to Data? We can easily just create a `BeefySession` object and not changes it's values afterwards.

src/main/java/com/limechain/beefy/state/BeefyState.java
Zurcusa: Also It would be good if we start using `AuthoritySet` object so that we have a uniform object for authorities across consensus algorithms.

src/main/java/com/limechain/beefy/state/BeefySession.java
osrib: I wanted to set key pair if we found it otherwise log that we are not chosen

src/main/java/com/limechain/beefy/state/BeefyState.java
```diff
@@ -108,16 +110,22 @@ private void initializeNextDigest() {
...
-                BeefySession beefySession = new BeefySession(validatorSet);
+                AuthoritySet authoritySet = validatorsSet.getRight();
+
+                BeefySession beefySession = new BeefySession(authoritySet);
+                keyStore.findKeyPair(authoritySet.getAuthorities(), KeyType.BEEFY).ifPresentOrElse(
+                        beefySession::setBeefyKeyPair,
+                        () -> log.info(
+                                String.format("BEEFY: We are not chosen to vote in current session, block number: %s",
+                                        sessionBlock))
+                );
```
Zurcusa:
```suggestion
                AuthoritySet authoritySet = validatorsSet.getRight();

                BeefySession beefySession;
                var keyPairOpt = keyStore.findKeyPair(authoritySet.getAuthorities(), KeyType.BEEFY);
                if ( keyPairOpt.isPresent()) {
                    beefySession = new BeefySession(authoritySet, keyPairOpt.get());
                } else {
                    beefySession = new BeefySession(authoritySet, null);
                    log.info(String.format("BEEFY: We are not chosen to vote in current session, block number: %s", sessionBlock));
                );
```

src/main/java/com/limechain/beefy/state/BeefySession.java
Zurcusa: Check my other comment on how we can keep the `@Value`

src/main/java/com/limechain/beefy/state/BeefySession.java
```diff
@@ -1,16 +1,22 @@
...
 public class BeefySession {
-    ValidatorSet validatorSet;
+    AuthoritySet authoritySet;
     Set<BigInteger> nonMandatoryBlockNumbers = Collections.synchronizedSet(new HashSet<>());
+    Pair<byte[], byte[]> beefyKeyPair = null;
```
Zurcusa:
```suggestion
    @Nullable
    Pair<byte[], byte[]> beefyKeyPair = null;
```

Zurcusa (approved):

PR #827, opened 2025-03-11
feat: Implement handling of consensus messages.

### LINKED ISSUE #773: Handle authorities changes in Beefy (research & implement)

hMitov:
# Description

- Implemented handling of consensus messages **(BEEFY_CHANGED_AUTHORITIES, BEEFY_ON_DISABLED)**.
- Introduces method, named **retrieveMandatoryBlockAndCreateBeefySession()** - it examines BEEFY authority consensus messages, within a specified block range to detect authority set changes or disabled authorities. This method is being used in two places in Beefy:
1) in **handleBeefySessionTransitions()** - it is applied to all the blocks between beefyDigest and grandpa finalized. Each found mandatory block with its corresponding beefy session is appended. This method is being used when we have to basically go through blocks that are between **lastBeefyFinalized**/**beefyGenesis** and **grandpaFinalized**

2) in **processJustification()** - it is applied to each of the incoming **justifications**(signed commitments) from our peers. This method will be used to **trigger verifying of justification, checking/storing justifications from/to DB, update beefy state etc.**
I provided partial implementation of this method - only parts that are related to **creating** and **erasing** **sessions** and **blocks**. Rest of the method is left with //TODOs for future implementation.
     

Fixes #773


src/main/java/com/limechain/consensus/beefy/BeefyService.java
```diff
@@ -54,14 +61,249 @@ private BigInteger getThreshold() {
...
+     * Examines BEEFY authority consensus messages, within a specified block range to detect authority set changes or
+     * disabled authorities.
+     * <p>
+     * Upon encountering BEEFY_CHANGED_AUTHORITIES message, it extracts the authority set and returns it within a
+     * BeefySession mapped to the corresponding block number. If no BEEFY_CHANGED_AUTHORITIES is found,
+     * empty map is returned.
+     * <p>
+     * If a BEEFY_ON_DISABLED message is found, it updates the beefyState with the disabled authority information.
+     */
+    private Map<BigInteger, BeefySession> retrieveMandatoryBlockAndCreateBeefySession(BigInteger higherBlock,
```
Zurcusa: We can keep this method, but rename it to `processConsensusMessages(List<BlockHeader> headers`.
The flow would be as follows:
- Check if the first element's number equals `grandpaFinalized` from `BeefyState` + 1. In this way we ensure that no unexpected jumps were made in the finalization notifications.
- For each block retrieve beefy consensus messages if any.
- For each consensus message handle it via `handleBeefyConsensusMessage` inside of the `BeefyState`.

src/main/java/com/limechain/consensus/beefy/BeefyService.java
```diff
@@ -54,14 +61,249 @@ private BigInteger getThreshold() {
...
+                return result;
+            }
+
+            currentBlockNumber = currentBlockNumber.subtract(BigInteger.ONE);
+        }
+
+        return Collections.emptyMap();
+    }
+
+    private Map<BigInteger, BeefySession> processAuthorityConsensusMessages(BlockHeader blockHeader,
```
Zurcusa: Move this method to `BeefyState` and rename it to `handleBeefyConsensusMessage`. For `BEEFY_CHANGED_AUTHORITIES` create a method inside of the `BeefyState` which does the following:
- For the authorities inside if the consensus message retrieves our session key from the `KeyStore` via `findKeyPair`.
- Creates a new session for the provided block and adds it to the queue.

src/main/java/com/limechain/consensus/beefy/dto/BeefySession.java
Zurcusa, line 1: Rework this class so that it includes the following fields:
- A Map<Commitment, BeefyRound> that stored the rounds inside of the session.
- A `BigInteger` that represents the first block of the session.
- The `BeefyAuthoritySet`.
- Boolean if the mandatory block of the session has been finalized.
- A nullable BigInteger representing the highest finalized beefy block for the session.

src/main/java/com/limechain/consensus/beefy/BeefyState.java
```diff
@@ -112,13 +110,15 @@ private void handleSessionTransitions(BigInteger grandpaFinalized) {
             //Todo: In kagome fetching header.
```
Zurcusa: This method will no longer be needed as it is. You can rename it to something like `handleAuthoritySetChange`. And implement the logic that I explained for the method used in the switch of `handleBeefyConsensusMessage`.

Zurcusa:
Please copy the branch into a separate one so that we have a duplicate of the code that will not make it into dev in case we need some of it in the future. Implement all the changes mentioned in the comments and republish this PR as ready for review.

src/main/java/com/limechain/consensus/beefy/dto/BeefySession.java
Zurcusa, line 1: Having `@Value` here wouldn't work. We have fields that needs to be updated, isMandatoryBlockFinalized and highestFinalizedForSession. I suggest we remove the annotation and instead have `@Data` with `private final` on everything, but the 2 fields mentioned. For those we would have setters. For now it's not necessary for custom logic.

src/main/java/com/limechain/consensus/beefy/dto/BeefyRound.java
```diff
@@ -0,0 +1,13 @@
...
+
+import com.limechain.network.protocol.beefy.messages.vote.VoteMessage;
+import lombok.Value;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+@Value
+public class BeefyRound {
+    //mapper key is authority public key
```
Zurcusa: We should add a TODO so that we find an appropriate key for the 33 byte array here.

src/main/java/com/limechain/consensus/beefy/BeefyState.java
```diff
@@ -74,11 +70,8 @@ public class BeefyState extends AbstractState implements ServiceConsensusState {
     @Nullable
     private VoteMessage lastVote;
 
-    //mapper key is mandatory block number where authority set change appeared
-    private LinkedHashMap<BigInteger, BeefySession> sessions = new LinkedHashMap<>();
+    private List<BeefySession> sessions = Collections.synchronizedList(new ArrayList<BeefySession>());
```
Zurcusa: I would change this to a `Dequeue`. We would always work with the front of the collection when processing votes/justifications and voting. However it would be good to have the option to peek at the last element just in case we decide to add some logic related to that. I also don't think this needs to be synchronized as no multi threading would happen when we create new sessions, not when we prune. Both of the processes would happen in the main beefy logic loop in various parts (eg we conclude a round, we finalize a justification)

Zurcusa (approved):

PR #828, opened 2025-03-11
feat: Implement ForkTree

### LINKED ISSUE #814: Fix the handling of Forced/Scheduled Authority Set Changes

Grigorov-Georgi:
# Description

- Implement only the necessary methods in the ForkTree for forced/scheduled authority set changes. 
- Add tests to ensure that the tree functions correctly.

Partly Fixes #814


src/main/java/com/limechain/storage/forktree/ForkTree.java
```diff
@@ -0,0 +1,427 @@
...
+            Collections.reverse(childrenCopy);
+            stack.addAll(childrenCopy);
+        }
+
+        return nodes;
+    }
+
+    @Getter
+    @Setter
+    @NoArgsConstructor
```
Zurcusa: Can't we use the `@Value` annotation? It seems that we don't set the values of a TreeNode anywhere in code after creation. Also do we need the no arg ctor for the tests somewhere?

Zurcusa (approved):

PR #847, opened 2025-03-18
Implement accept block interval.

### LINKED ISSUE #843: Logic for accepted vote/justification interval.
Incoming justifications can be handled in one of 3 ways:
- Process
- Enqueue
- Drop

Votes are similar, but only have 2 options:
- Process
- Drop

The action is decided based on the current accepted interval of BEEFY round. We need logic that can retrieve the accepted interval based on a few factors:
- If the latest session in the queue has it's mandatory block finalized.
In this case we we return an accepted interval in the range or `(max(last beefy finalized, session start block), last grandpa finalized)`
- If the latest in the queue doesn't have its mandatory block finalized.
In this case we return `(session start block number, session start block number)`

hMitov:
# Description

Implement accept block interval. Exceptions are thrown, so they should be handled in the parent method with try catch.

Fixes #843

### DIFF
--- a/src/main/java/com/limechain/consensus/beefy/BeefyService.java
+++ b/src/main/java/com/limechain/consensus/beefy/BeefyService.java
@@ -179,21 +179,21 @@ public class BeefyService {
     private Pair<BigInteger, BigInteger> findAcceptedBlocksInterval() {
         BeefyState beefyState = stateManager.getBeefyState();
 
-        BeefySession lastSession = beefyState.getSessions().peekLast();
-        if (Objects.isNull(lastSession)) {
+        BeefySession currentSession = beefyState.getSessions().peekFirst();
+        if (currentSession == null) {
             throw new BeefyGenericException("No beefy session exists.");
         }
 
         BigInteger beefyFinalized = beefyState.getBeefyFinalized();
-        if (Objects.isNull(beefyFinalized)) {
+        if (beefyFinalized == null) {
             throw new BeefyGenericException("Beefy finalized is not initialized yet.");
         }
 
-        if (lastSession.isMandatoryBlockFinalized()) {
-            BigInteger lowerBlock = beefyState.getBeefyFinalized().max(lastSession.getMandatoryBlock());
+        if (currentSession.isMandatoryBlockFinalized()) {
+            BigInteger lowerBlock = beefyFinalized.max(currentSession.getMandatoryBlock());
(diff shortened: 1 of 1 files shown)


src/main/java/com/limechain/consensus/beefy/BeefyService.java
```diff
@@ -174,4 +175,25 @@ private Optional<byte[]> extractMmrRootHash(BlockHeader blockHeader) {
                 .stream().map(BeefyConsensusMessage::getMmrRootHash)
                 .findFirst();
     }
+
+    private Pair<BigInteger, BigInteger> findAcceptedBlocksInterval() {
+        BeefyState beefyState = stateManager.getBeefyState();
+
+        BeefySession lastSession = beefyState.getSessions().peekLast();
```
Zurcusa:
```suggestion
        BeefySession lastSession = beefyState.getSessions().peekFirst();
```

Zurcusa: We want to check the head of the queue because that is always the current session that we are working with.

src/main/java/com/limechain/consensus/beefy/BeefyService.java
```diff
@@ -174,4 +175,25 @@ private Optional<byte[]> extractMmrRootHash(BlockHeader blockHeader) {
...
+        if (Objects.isNull(lastSession)) {
```
Zurcusa: I'd prefer if we stick to the good old `== null` syntax.

src/main/java/com/limechain/consensus/beefy/BeefyService.java
```diff
@@ -174,4 +175,25 @@ private Optional<byte[]> extractMmrRootHash(BlockHeader blockHeader) {
...
+            throw new BeefyGenericException("No beefy session exists.");
+        }
+
+        BigInteger beefyFinalized = beefyState.getBeefyFinalized();
+        if (Objects.isNull(beefyFinalized)) {
+            throw new BeefyGenericException("Beefy finalized is not initialized yet.");
+        }
+
+        if (lastSession.isMandatoryBlockFinalized()) {
+            BigInteger lowerBlock = beefyState.getBeefyFinalized().max(lastSession.getMandatoryBlock());
```
Zurcusa: Why don't you reuse the var that you created a few lines above?

Zurcusa (approved):

PR #855, opened 2025-03-24
feat: finalize beefy justification

### LINKED ISSUE #844: Finalize BEEFY justification.
If a justification covers these 2 conditions: 

- Is valid.
- Is in accepted interval (#843).

Then it can be finalized. This means that the round that the justification is for gets concluded and the beefy state gets updated. An example flow of this could look like this:
- Conclude the round from the justification.
- Prune stale/finished older sessions.
- Update BEEFY finalized block.
- Persist BEEFY state.

hMitov:
# Description

Implemented logic to finalize a round based on a signed commitment(justification).
It includes:

- Removing finalized rounds, that are <= justified round, from the current session.

- Removing finalized sessions, which mandatory block is finalized.

- Updating the BEEFY state within the session.

- Updating the BEEFY finalized block.

Fixes #844


src/main/java/com/limechain/consensus/beefy/BeefyService.java
```diff
@@ -171,11 +177,14 @@ private Optional<byte[]> extractMmrRootHash(BlockHeader blockHeader) {
     private void triageIncomingJustification(SignedCommitment signedCommitment) {
         BigInteger blockNumber = signedCommitment.getCommitment().getBlockNumber();
         RoundAction roundAction = determineRoundAction(blockNumber);
+        if (roundAction == RoundAction.INVALID) {
```
Grigorov-Georgi: You can put that if check in the following switch statement.

src/main/java/com/limechain/consensus/beefy/BeefyService.java
```diff
@@ -222,9 +237,55 @@ private Pair<BigInteger, BigInteger> findAcceptedRoundsInterval() {
...
+        BigInteger highestFinalized = currentSession.getHighestFinalizedForSession();
+        highestFinalized = highestFinalized == null ? blockNumber : highestFinalized.max(blockNumber);
+        currentSession.setHighestFinalizedForSession(highestFinalized);
+
+        if (blockNumber.equals(currentSession.getMandatoryBlock())) {
+            currentSession.setMandatoryBlockFinalized(true);
+            log.fine(String.format("finalizeJustification: Finalize mandatory round: %d.", blockNumber));
+        } else {
+            log.fine(String.format("finalizeJustification: Finalize non-mandatory round: %d.", blockNumber));
+        }
```
Zurcusa: I feel like we should move this logic inside the `BeefySession` itself.

Zurcusa (approved):

Grigorov-Georgi (approved):

PR #859, opened 2025-03-25
feat: added handling incoming votes

### LINKED ISSUE #848: Handle incoming votes.
After a vote has been validated we need to check if it can be added to the current ongoing session. The responsibility of this is spit into two:

`BeefySession` - The session itself takes care of the responsibility to decide what do to with the incoming vote. There are a few possible options:
- The vote's block number is < `mandatoryBlock` or <= `highestFinalizedForSession`. We can consider this vote stale and just drop it.
- The session has already registered a vote by this authority, but for a different commitment payload. In this case we have an equivocation. (as of now we don't have an optimal way to check if an authority has voted. Maybe we need a to preserve vote messages that we've received from authorities and check against them)
- If the vote passes the previous 2 checks it is valid. At this point we add it to the session's map based on its commitment. Every time we add a valid vote we can check if the round has >= 2/3 + 1 of the voters. If so the round can be considered as finished.

`BeefyService` - Retrieve the head of the session queue and try to add the vote to it. Based on the response from the functionality above there are a few ways to go from here:
- If the addition results in a finished round we create a signed commitment, finalize it (#844) and then it has to be gossiped (#776).
- If the vote is added, but the round is not finished we simply persist the beefy state.
- If it's an equivocation vote we report it (#782)

osrib:
## Checklist:


src/main/java/com/limechain/consensus/beefy/dto/BeefySession.java
```diff
@@ -30,6 +41,83 @@ public class BeefySession {
...
+    }
+
+    /**
+     * The threshold is determined as the numOfValidators - (numOfValidators - 1) / 3
+     *
+     * @return minimum required validators for finality.
+     */
+    private BigInteger getThreshold() {
+
+        if (Objects.isNull(authoritySet)) {
```
Zurcusa: Let's stick to the good old `== null` syntax.

src/main/java/com/limechain/consensus/beefy/dto/BeefySession.java
```diff
@@ -30,6 +41,83 @@ public class BeefySession {
...
+            log.warning("getThreshold: No authoritySet in BeefySession.");
+            return BigInteger.ZERO;
```
Zurcusa: I think it would be more appropriate to throw and exception here. A session shouldn't exist without authorities.

src/main/java/com/limechain/consensus/beefy/dto/DoubleVotingProof.java
```diff
@@ -0,0 +1,12 @@
...
+
+import com.limechain.network.protocol.beefy.messages.vote.VoteMessage;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class DoubleVotingProof {
+    private VoteMessage first;
+    private VoteMessage second;
```
Zurcusa:
```suggestion
@Value
public class DoubleVotingProof {
    VoteMessage first;
    VoteMessage second;
```

src/main/java/com/limechain/consensus/beefy/dto/BeefySession.java
```diff
@@ -30,6 +41,83 @@ public class BeefySession {
...
+    public SignedCommitment createSignedCommitment(BeefyRound round, Commitment commitment) {
+        Map<Hash264, VoteMessage> signedVotes = round.getSignedVotes();
+
+        List<Optional<byte[]>> signatures = authoritySet.getPublicKeys().stream()
+                .map(key -> Optional.ofNullable(signedVotes.get(new Hash264(key)))
+                        .map(VoteMessage::getSignature))
+                .collect(Collectors.toList());
+
+        return new SignedCommitment(commitment, signatures);
+    }
```
Zurcusa:
```suggestion
    public SignedCommitment createSignedCommitment(BeefyRound round, Commitment commitment) {
    
        Map<Hash264, VoteMessage> signedVotes = round.getSignedVotes();

        List<Optional<byte[]>> signatures = authoritySet.getPublicKeys().stream()
                .map(key -> Optional.ofNullable(signedVotes.get(new Hash264(key)))
                        .map(VoteMessage::getSignature))
                .collect(Collectors.toList());

        return new SignedCommitment(commitment, signatures);
    }
```

src/main/java/com/limechain/consensus/beefy/dto/BeefySession.java
```diff
@@ -30,6 +41,83 @@ public class BeefySession {
...
+                blockNumber.compareTo(highestFinalized) <= 0) {
+            log.info(String.format("Beefy: received vote for old stale round {%s}, ignoring", blockNumber));
+            return new VoteImportResult.Invalid();
+        } else if (!Objects.equals(commitment.getAuthoritySetId(), authoritySet.getSetId())) {
+            log.info(String.format("Beefy: expected set_id {%s}, ignoring vote {%s}", authoritySet.getSetId(), voteMessage));
+            return new VoteImportResult.Invalid();
+        } else if (!authoritySet.getPublicKeys().contains(authorityId)) {
+            log.info(String.format("Beefy: received vote {%s} from validator that is not in the validator set, ignoring", voteMessage));
+            return new VoteImportResult.Invalid();
+        }
```
Zurcusa:
```suggestion
        if (blockNumber.compareTo(mandatoryBlock) < 0 ||
                blockNumber.compareTo(highestFinalized) <= 0) {
            log.fine(String.format("Beefy: received vote for old stale round {%s}, ignoring", blockNumber));
            return new VoteImportResult.Invalid();
        } else if (!Objects.equals(commitment.getAuthoritySetId(), authoritySet.getSetId())) {
            log.fine(String.format("Beefy: expected set_id {%s}, ignoring vote {%s}", authoritySet.getSetId(), voteMessage));
            return new VoteImportResult.Invalid();
        } else if (!authoritySet.getPublicKeys().contains(authorityId)) {
            log.fine(String.format("Beefy: received vote {%s} from validator that is not in the validator set, ignoring", voteMessage));
            return new VoteImportResult.Invalid();
        }
```

src/main/java/com/limechain/consensus/beefy/dto/BeefySession.java
```diff
@@ -30,6 +41,83 @@ public class BeefySession {
     @Nullable
     private final Pair<byte[], byte[]> beefyKeyPair;
 
+
+    public VoteImportResult addVote(VoteMessage voteMessage) {
+        Commitment commitment = voteMessage.getCommitment();
```
Zurcusa, line 47:
```suggestion
    public VoteImportResult addVote(VoteMessage voteMessage) {
    
        Commitment commitment = voteMessage.getCommitment();
```

src/main/java/com/limechain/consensus/beefy/dto/BeefySession.java
```diff
@@ -30,6 +41,83 @@ public class BeefySession {
...
+        }
+
+        BeefyRound round = rounds.computeIfAbsent(commitment, _ -> new BeefyRound());
+        if (round.addVote(authorityIdHash, voteMessage) &&
+                round.isDone(getThreshold())) {
+            rounds.remove(commitment);
+            log.info(String.format("Beefy: Round # {%s} concluded, finality_proof: ", blockNumber));
+            return new VoteImportResult.RoundConcluded(createSignedCommitment(round, commitment));
+        }
+        return new VoteImportResult.Ok();
```
Zurcusa:
```suggestion
        }
        
        return new VoteImportResult.Ok();
```

src/main/java/com/limechain/consensus/beefy/dto/BeefySession.java
Zurcusa, line 1: Could you use the name of the method at the start of the logs instead of beefy? At some point we'll add a log filter that adds the name automatically.

src/main/java/com/limechain/consensus/beefy/BeefyService.java
```diff
@@ -174,6 +179,30 @@ private Optional<byte[]> extractMmrRootHash(BlockHeader blockHeader) {
...
+        switch (roundAction) {
+            case RoundAction.PROCESS -> {
+                log.fine(String.format("triageIncomingVotes: Process vote %s  for round: %d.", voteMessage, blockNumber));
+                Optional<SignedCommitment> finalityProof = handleVote(voteMessage);
+                if (finalityProof.isPresent()) {
+                    //TODO: gossip vote message
+                }
+            }
+            case RoundAction.ENQUEUE -> {
+                log.warning(String.format("triageIncomingVotes: Unexpected vote: %s", voteMessage));
```
Zurcusa:
```suggestion
                log.fine(String.format("triageIncomingVotes: Unexpected vote: %s", voteMessage));
```

src/main/java/com/limechain/consensus/beefy/BeefyService.java
```diff
@@ -174,6 +179,30 @@ private Optional<byte[]> extractMmrRootHash(BlockHeader blockHeader) {
                 .findFirst();
     }
 
+    private void triageIncomingVote(VoteMessage voteMessage) {
+        BigInteger blockNumber = voteMessage.getCommitment().getBlockNumber();
+        RoundAction roundAction = determineRoundAction(blockNumber);
```
Zurcusa:
```suggestion
    private void triageIncomingVote(VoteMessage voteMessage) {
    
        BigInteger blockNumber = voteMessage.getCommitment().getBlockNumber();
        RoundAction roundAction = determineRoundAction(blockNumber);
```

src/main/java/com/limechain/consensus/beefy/BeefyService.java
```diff
@@ -27,12 +28,12 @@
 import java.util.Objects;
 import java.util.Optional;
 
+
```
Zurcusa:
```suggestion
```

Zurcusa (approved):

PR #860, opened 2025-03-25
feat: persist beefy state

### LINKED ISSUE #778: Persist the beefy state

Grigorov-Georgi:
# Description

- Add Beefy- and GrandpaSetRepository which should be responsible for fetching and persisting data to the KVRepository
- Add methods for fetching & persisting needed fields from the BeefyState
- We persist the state from BeefyService when:
1) FinalizedBlockChangeEvent is received and the grandpaFinalized is updated
2) We vote and update the lastVoted field of the state
3) We handle a vote and a new vote is added to the session
4) We finalize a beefy justification and a session is removed and beefyFinalized is updated

Fixes https://github.com/LimeChain/Fruzhin/issues/778


src/main/java/com/limechain/consensus/beefy/BeefyState.java
```diff
@@ -41,13 +38,13 @@
...
+    private final KeyStore keyStore;
 
     private BigInteger disabledAuthority;
 
-    private final BlockState blockState;
-    private final KeyStore keyStore;
-    private final KVRepository<String, Object> repository;
+    private BeefyAuthoritySet authoritySet;
 
     private BigInteger roundNumber;
```
Zurcusa, line 49: I have a feeling we won't need this, as well.

src/main/java/com/limechain/consensus/beefy/BeefyState.java
```diff
@@ -141,79 +146,15 @@ private void reportDoubleVoting(VoteMessage voteMessage) {
...
-    }
-
-    private void persistDisabledAuthority() {
-        repository.save(
-                StateUtil.generateBeefyDisabledAuthorityKey(
-                        DBConstants.BEEFY_DISABLED_AUTHORITY, authoritySet.getSetId()
-                ),
-                disabledAuthority
-        );
-    }
```
Zurcusa, line 188: I suggest we keep this, just in case.

Zurcusa (approved):

src/main/java/com/limechain/consensus/beefy/BeefyState.java
Grigorov-Georgi: I forgot we talked about not persisting that field, I'll remove it.

PR #876, opened 2025-04-02
feat: Implement beefy justification request response protocol.

### LINKED ISSUE #835: Implement requesting beefy justifications on demand.
On different occasions we will need to request specific beefy justifications for blocks. That is done via the `/beefy/justifications/1` sub-stream in a similar approach to how we would request blocks from the block protocol.

Zurcusa:
# Description

This PR implements the logic for requesting BEEFY justification on demand via the request response protocol `/beefy/justifications/1`. It separates the two BEEFY protocols (notification and request response).

Fixes #835 

## Checklist:


src/main/java/com/limechain/network/protocol/warp/dto/BlockHeader.java
```diff
@@ -47,6 +47,7 @@ public byte[] getBlake2bHash(boolean sealed) {
         return HashUtils.hashWithBlake2b(scaleEncoded);
     }
 
+    // TODO We cannot do this as the blake2b hashing is irreversible.
```
Grigorov-Georgi: I am wondering if we can change every usage of BlockHeader.fromHash(hash) to stateManager.getBlockState.getHeader(hash), or it won't be so easy?

src/main/java/com/limechain/network/PeerMessageCoordinator.java
```diff
@@ -136,9 +136,10 @@ public void sendVoteMessageToPeers(VoteMessage voteMessage) {
 
     public void sendBeefyVoteMessageToPeers(BeefyVoteMessage voteMessage) {
         byte[] scaleMessage = ScaleUtils.Encode.encode(BeefyVoteMessageScaleWriter.getInstance(), voteMessage);
-        sendMessageToActivePeers(peerId -> asyncExecutor.executeAndForget(() -> network.getBeefyService().sendVoteMessage(
-                network.getHost(), peerId, scaleMessage
-        )));
+        sendMessageToActivePeers(peerId -> asyncExecutor.executeAndForget(() ->
+                network.getBeefyNotificationService().sendVoteMessage(
```
Grigorov-Georgi: If https://github.com/LimeChain/Fruzhin/pull/877 is merged before this branch, you should use .sendMessage() instead of .sendVoteMessage()

Grigorov-Georgi (approved):

src/main/java/com/limechain/consensus/beefy/BeefyService.java
```diff
@@ -317,7 +317,7 @@ private RoundAction determineRoundAction(BigInteger roundNumber) {
         BeefyState beefyState = stateManager.getBeefyState();
 
         if (beefyState.getSessions().isEmpty()) {
-            log.warning("determineRoundAction: No beefy session exists.");
+            log.fine("determineRoundAction: No beefy session exists.");
```
hMitov, line 320: Maybe it is good idea to define a consistent rule for when something is truly unexpected and needs warning-level attention. We have similar checks in code where we use log.warning.

hMitov (approved):

src/main/java/com/limechain/network/protocol/beefy/requestresponse/BeefyJustificationMessages.java
```diff
@@ -0,0 +1,38 @@
...
+        super(protocolId, protocol);
+    }
+
+    public SignedCommitment remoteJustificationRequest(Host us, PeerId peer, BigInteger from) {
+        try {
+            BeefyJustificationController controller = dialPeer(us, peer, us.getAddressBook());
+
+            return controller
+                    .sendJustificationRequest(from)
+                    .get(3, TimeUnit.SECONDS);
```
osrib, line 29: Maybe it is better to use completeOnTimeout()

osrib (approved):

src/main/java/com/limechain/network/protocol/warp/dto/BlockHeader.java
Zurcusa: It looks like we can indeed.

src/main/java/com/limechain/consensus/beefy/BeefyService.java
Zurcusa: Indeed. We need stricter logging rules. Not on the radar as of now, but this one was spamming quite a bit when I started the node so I changed the logging level.

src/main/java/com/limechain/network/protocol/beefy/requestresponse/BeefyJustificationMessages.java
Zurcusa: If we use `completeOnTimeout()`, we'd have to return either null or an empty optional on timeout. Throwing an exception seems more suitable in the current setup.

PR #878, opened 2025-04-02
feat: implement beefy main loop

### LINKED ISSUE #852: Implement main BEEFY loop.

osrib:
# Description

Implemented Beefy main loop that consist of applying pending justification and voting

Fixes #852 

## Checklist:


src/main/java/com/limechain/consensus/beefy/BeefyService.java
```diff
@@ -472,4 +475,14 @@ public boolean isBeefyMessageAcceptable(Commitment commitment) {
 
         return true;
     }
+
+    public void run() {
+
+        log.info("runBeefy: Started Beefy Service main loop");
```
Zurcusa, line 481: What happens if an exception is thrown from within the thread? Does the thread just die? How do we restart the loop?

Zurcusa (approved):

hMitov (approved):

Grigorov-Georgi (approved):

PR #879, opened 2025-04-03
feat: Improve neighbor message sending logic

### LINKED ISSUE #708: Improve neighbor message sending logic.
As per [spec](https://spec.polkadot.network/chap-networking#sect-grandpa-neighbor-msg):

> Neighbor messages are sent to all connected peers but they are not re-propagated on reception. A message should be sent whenever the values of the message change and at least every 5 minutes. The sender should take the recipient's state into account and avoid sending messages to peers that are using different voter sets or are in a different round. Messages received from a future voter set or round can be dropped and ignored.

Our current flow is as follows:

- We send neighbor messages every 5 minutes (5 minute init delay) to all connected peers. - This complies with the spec regarding the 5 minutes rule, but we do not check which of our peers are actually in the same set and round as us. Also the 5 minute initial delay can be skipped. A more suitable approach would be to check if we are the head of the chain and then actually send scheduled messages.
- Every time we receive a neighbor message we send one back to the sending peer - I am not sure why this was done as it does not comply to the spec in any way.

Desirable flow:
We send neighbor messages in the following scenarios:

- When we finalize a block. 
  - For light client, full node and authoring node not a part of the voter set this will happen when commit message.
  - For authoring node that is a part of the voter set this will happen when we finalize a block via the `Atemp-to-Finalize`
- When we switch a setId.
  - For light client, full node and authoring node not a part of the voter set this will happen when we handle authority changes from incoming blocks.
  - For authoring node that is a part of babe producers this will happen when we handle authority changes from produced and incoming blocks.
- When we switch a round.
  - For light client, full node and authoring node not a part of the voter set this will happen on commit message when it has a different roundId from ours.
  - For authoring node that is a part of the voter set this will happen when we play a new grandpa round.

hMitov:
# Description

### Workflow
Neighbor messages are broadcast:
- On block finalization;
- When the authority set changes;
- When switching to a new round;
- And periodically, every 5 minutes.

**Finalization of block**
We can finalize a block from a commit message and when we `Attempt-to-Finalize`.

**Switch a setID**
When we handle authority changes from incoming blocks and from produced blocks (when node is part of babe producers).

**Switch a round**
When we receive a commit message, that implies that the next round has started, we should broadcast is(and maybe if we are active voters we should start a new round, with number from the commit message). The other scenario is when we play a new grandpa round(when node is active authority).

### Changes in PR
- Renamed `sendMessagesToPeers()` method to `sendNeighborMessageToPeers()` - it is more descriptive this way;
- Trigger neighbor message broadcast when a commit is received and during `attemptToFinalize()`;
- Trigger neighbor message on starting of new authority set;
- Added setid, roundnum and lastfinalized to peerInfo with corresponding update method, used on receiving of neighbor message;
- Added checks for peer set id and round num on receiving neighbor messages with //TODOs for future rating trackings;
- Added check when broadcasting neighbor message - if peer set id is same as ours and peer is Light client, do not broadcast.

Fixes #708 >

## Checklist:


src/main/java/com/limechain/network/ConnectionManager.java
```diff
@@ -285,6 +308,14 @@ public void removePeer(PeerId peerId) {
         log.info(String.format("removePeer: removed %s", peerId));
     }
 
+    public boolean checkIfPeerIsAuthorNode(PeerId peerId) {
+        return NodeRole.AUTHORING.getValue().equals(getPeerInfo(peerId).getNodeRole());
```
Grigorov-Georgi: This may result in a null pointer exception.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
```diff
@@ -111,16 +112,25 @@ public void writeHandshakeToStream(Stream stream, PeerId peerId) {
...
      *
      * @param stream <b>responder</b> stream to write the message to
      * @param peerId peer to send to
      */
     public void writeNeighbourMessage(Stream stream, PeerId peerId) {
+        PeerInfo peerInfo = connectionManager.getPeerInfo(peerId);
+        NeighbourMessage neighbourMessage = ProtocolMessageBuilder.buildNeighbourMessage();
+
+        boolean sameSetId = peerInfo.getSetId() != null
+                && neighbourMessage.getSetId().compareTo(peerInfo.getSetId()) == 0;
```
Grigorov-Georgi: You can use BigInteger.equals(BigInteger) instead of null check + BigInteger.compareTo(BigInteger) == 0. It's shorter and internally makes the null check and then compares the values of BigIntegers.

Grigorov-Georgi (approved):

osrib (approved):

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
```diff
@@ -111,16 +112,25 @@ public void writeHandshakeToStream(Stream stream, PeerId peerId) {
...
 
     /**
-     * Send our GRANDPA neighbour message from {@link WarpSyncState} on a given <b>responder</b> stream.
+     * Send our GRANDPA neighbour message on a given <b>responder</b> stream.
      *
      * @param stream <b>responder</b> stream to write the message to
      * @param peerId peer to send to
      */
     public void writeNeighbourMessage(Stream stream, PeerId peerId) {
+        PeerInfo peerInfo = connectionManager.getPeerInfo(peerId);
```
Zurcusa, line 121: It is possible for `getPeerInfo` to return null. Maybe we should have a null check for it here and log in case we end up at this edge case.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
```diff
@@ -224,17 +234,31 @@ private GrandpaMessageType getGrandpaMessageType(byte[] message) {
 
     private void handleNeighbourMessage(byte[] message, Stream stream) {
 
+        PeerId peerId = stream.remotePeerId();
+        PeerInfo peerInfo = connectionManager.getPeerInfo(peerId);
```
Zurcusa, line 242: Same comment for null check applies here.

src/main/java/com/limechain/network/protocol/grandpa/GrandpaEngine.java
```diff
@@ -224,17 +234,31 @@ private GrandpaMessageType getGrandpaMessageType(byte[] message) {
...
+
         NeighbourMessage neighbourMessage = ScaleUtils.Decode.decode(
                 message,
                 NeighbourMessageScaleReader.getInstance()
         );
 
-        log.log(Level.FINE, "Received neighbour message from Peer " + stream.remotePeerId() + "\n" + neighbourMessage);
-        // TODO: We need to actually update our peer's infos on each message.
-        writeNeighbourMessage(stream, stream.remotePeerId());
+        log.log(Level.FINE, "Received neighbour message from Peer " + peerId + "\n" + neighbourMessage);
```
Zurcusa:
```suggestion
        log.fine(String.format("Received neighbour message from Peer %s \n %s", peerId, neighbourMessage);
```

Zurcusa (approved):

PR #882, opened 2025-04-07
feat: stop services and save states appropriately when node stops

### LINKED ISSUE #791: Stop services and save states appropriately when node stops

osrib:
# Description

Fixes #791: stop services and save states appropriately when node stops

## Checklist:
- [ ] All new and existing tests passed.

### DIFF
--- a/src/main/java/com/limechain/Main.java
+++ b/src/main/java/com/limechain/Main.java
@@ -62,8 +62,7 @@ public class Main {
             prometheusServer.stop();
             rpcApp.stop(); // NOTE: rpcApp is responsible for stopping everything that could've been started
 
-            // TODO: Maybe think of another place to hold the logic below
-            log.log(Level.INFO, "\uD83D\uDED1Stopped {0} client!", nodeRole);
+            System.out.printf("\uD83D\uDED1Stopped %s client!%n", nodeRole);
             System.exit(0);
         });
     }
--- a/src/main/java/com/limechain/Main.java
+++ b/src/main/java/com/limechain/Main.java
@@ -62,7 +62,7 @@ public class Main {
             prometheusServer.stop();
             rpcApp.stop(); // NOTE: rpcApp is responsible for stopping everything that could've been started
 
-            System.out.printf("\uD83D\uDED1Stopped %s client!%n", nodeRole);
+            System.out.printf("Stopped %s client!%n", nodeRole);
             System.exit(0);
         });
     }
--- a/src/main/java/com/limechain/consensus/beefy/BeefyRepository.java
+++ b/src/main/java/com/limechain/consensus/beefy/BeefyRepository.java
@@ -14,6 +14,7 @@ import java.util.ArrayDeque;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.List;
+import java.util.function.Consumer;
 
 @Component
 @RequiredArgsConstructor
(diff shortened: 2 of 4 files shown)


src/main/java/com/limechain/Main.java
```diff
@@ -62,8 +62,7 @@ public static void main(String[] args) throws IOException {
             prometheusServer.stop();
             rpcApp.stop(); // NOTE: rpcApp is responsible for stopping everything that could've been started
 
-            // TODO: Maybe think of another place to hold the logic below
-            log.log(Level.INFO, "\uD83D\uDED1Stopped {0} client!", nodeRole);
```
Zurcusa, line 66: Do we have an idea why the logger no longer works? Not a crucial part of the ticket, but it would clear out some questions regarding the spring lifecycle.

src/main/java/com/limechain/consensus/beefy/BeefyState.java
```diff
@@ -169,4 +176,10 @@ private void loadPersistedState() {
         this.lastVoted = repository.fetchLastVoted();
         this.sessions = repository.fetchSessions();
     }
+
+    private <T> void saveIfNotNull(Consumer<T> saveMethod, T value) {
```
Zurcusa: Why not put this inside of the `BeefyRepository` ?

Zurcusa:
Please add a `Fixes #791` to the PR description.

Zurcusa:
A ticket that goes over the various states and defines proper cases on what to persist and what not would be a good idea. As of now I have a feeling that the BEEFY state is properly planned for persistence. We should also take into account thread safety so that we don't persist in the middle of an update operation.

src/main/java/com/limechain/Main.java
osrib: Because it is also bean from Log4j which dies earlier

src/main/java/com/limechain/consensus/beefy/BeefyState.java
hMitov: I think that we might put it in KVRepository instead, as we have some other customized methods for interaction with DB there, too.

osrib:
Currently, we cannot ensure thread safety in the @PreDestroy method, so we avoid persisting data during this phase to prevent saving partial or inconsistent state updates, as the application is continuously updating and saving the latest state.

osrib:
> s Currently, we cannot ensure thread safety in the @PreDestroy method, so we avoid persisting data during this phase to prevent saving partial or inconsistent state updates, as the application is continuously updating and saving the latest state.

osrib:
> A ticket that goes over the various states and defines proper cases on what to persist and what not would be a good idea. As of now I have a feeling that the BEEFY state is properly planned for persistence. We should also take into account thread safety so that we don't persist in the middle of an update operation.

Currently, we cannot ensure thread safety in the @PreDestroy method, so we avoid persisting data during this phase to prevent saving partial or inconsistent state updates, as the application is continuously updating and saving the latest state.

src/main/java/com/limechain/consensus/beefy/BeefyRepository.java
```diff
@@ -109,9 +110,17 @@ public SignedCommitment fetchJustification(BigInteger blockNumber) {
     }
 
     public void saveJustification(BigInteger blockNumber, SignedCommitment justification) {
-        repository.save(
-                StateUtil.generateBeefyJustificationKey(DBConstants.BEEFY_JUSTIFICATION, blockNumber),
-                justification
-        );
+        if (justification != null) {
+            repository.save(
```
Grigorov-Georgi: Is there a problem with using saveIfNotNull() instead of this null check?

Grigorov-Georgi (approved):

hMitov (approved):

PR #903, opened 2025-04-30
fix: babe validators flow

### LINKED ISSUE #422: Consensus integration tests

Grigorov-Georgi:
# Description

This PR refactors the BABE validator logic to enable starting a local network from genesis, ensuring proper block production and sealing from the very first block. The changes are backward-compatible and apply seamlessly when acting as a validator on other networks.

Partially fixes:  https://github.com/LimeChain/Fruzhin/issues/422

## Checklist:
- [X] I have read the [contributing guidelines](https://github.com/LimeChain/Fruzhin/blob/dev/CONTRIBUTING.md).
- [X] My PR title matches the [Conventional Commits spec](https://www.conventionalcommits.org/).
- [X] All new and existing tests passed.

### DIFF
--- a/src/main/java/com/limechain/storage/block/state/BlockState.java
+++ b/src/main/java/com/limechain/storage/block/state/BlockState.java
@@ -14,8 +14,10 @@ import com.limechain.network.protocol.warp.dto.BlockHeader;
 import com.limechain.network.protocol.warp.dto.Justification;
 import com.limechain.network.protocol.warp.scale.reader.BlockBodyReader;
 import com.limechain.network.protocol.warp.scale.writer.BlockBodyWriter;
+import com.limechain.rpc.server.AppBean;
 import com.limechain.rpc.subscriptions.chainsub.ChainSub;
 import com.limechain.runtime.Runtime;
+import com.limechain.runtime.RuntimeBuilder;
 import com.limechain.state.AbstractState;
 import com.limechain.storage.DBConstants;
 import com.limechain.storage.KVRepository;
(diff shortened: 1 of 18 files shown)


src/main/java/com/limechain/storage/block/state/BlockState.java
```diff
@@ -53,15 +55,17 @@ public class BlockState extends AbstractState {
...
+    public BlockState(KVRepository<String, Object> db,
+                      GenesisBlockHash genesisBlockHash) {
+
         this.db = db;
-        unfinalizedBlocks = new HashMap<>();
-        justifications = new LinkedHashMap<>();
-        genesisBlockHeader = genesisBlockHash.getGenesisBlockHeader();
+        this.unfinalizedBlocks = new HashMap<>();
+        this.justifications = new LinkedHashMap<>();
+        this.genesisBlockHash = genesisBlockHash;
```
Zurcusa, line 68: Doesn't `this` make sense only if the parameter has the same name as the field? I've always seen it as a way to distinguish class fields from parameters. Is there another reason to use it?

Zurcusa (approved):

hMitov (approved):

src/main/java/com/limechain/storage/block/state/BlockState.java
Grigorov-Georgi: I like to follow the rule that if one variable requires ```this```., then all of them should use it for consistency and better visual clarity.

PR #908, opened 2025-05-09
feat: Implemented syncing of state through rpc.

### LINKED ISSUE #642: Fix polkadot network TrieStorage fetching problem

hMitov:
# Description

### Initial Problem
 After completing warp sync, Fruzhin's build trie structure produced a trie which calculated merkle root (state root) did not match the expected state root of the last finalized block received from the network.

### Research revealed
- Key-value pairs retrieved over protocol were corrupted — missing keys, incorrect values or wrong ordering.
- After manually providing correct key-value pairs (dumped from other node implementations) to Fruzhin's trie construction logic, the reconstructed trie successfully produced the correct state root hash, confirming that the issue was not with the trie algorithm itself, but rather with the data retrieval process.

### Solution - fetch the state via RPC not protocol.
**Initialization**
- A pool of 100 WebSocket connections is created and maintained.
- Virtual threads, using Executors.newVirtualThreadPerTaskExecutor for high concurrency with low overhead. This enables fetching many values from keys in parallel, which results into speeding up the sync process.

**Key Fetching**
- It starts from a null as starting key and requests keys in batches (up to 1000 keys) using state_getKeysPaged RPC. On each follow request starting key is the last retrieved key from last batch.
- The method keeps paging through keys until there are no more keys left.
- If an empty list is returned, program retries up to 100 times to tolerate temporary node issues before aborting.

**Value Fetching:**
- For every key retrieved, the corresponding value is fetched concurrently using the state_getStorage RPC.
- If fetching a value fails or returns null, program retries up to 100 times per key.
- If after 100 attempts the value is still missing, the program aborts - this guarantees that no corrupted or incomplete state continues being used.

## Checklist:

### DIFF
--- a/src/main/java/com/limechain/config/HostConfig.java
+++ b/src/main/java/com/limechain/config/HostConfig.java
@@ -52,6 +52,13 @@ public class HostConfig {
     @Value("${genesis.path.local}")
     private String localGenesisPath;
 
+    @Value("${sync.state.path.polkadot}")
+    private String polkadotNodePath;
+    @Value("${sync.state.path.kusama}")
+    private String kusamaNodePath;
+    @Value("${sync.state.path.westend}")
+    private String westendNodePath;
+
     public HostConfig(CliArguments cliArguments) {
         this.rocksDbPath = cliArguments.dbPath();
         this.dbRecreate = cliArguments.dbRecreate();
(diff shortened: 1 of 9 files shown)


src/main/java/com/limechain/rpc/client/AbstractRpcClient.java
```diff
@@ -13,6 +13,7 @@
  */
 @Log
 public abstract class AbstractRpcClient extends WebSocketClient {
+    public static final String REG_EXPRESSION_NUMBER = "-?\\d+(\\.\\d+)?";
```
Grigorov-Georgi: Regular expressions are considered bad practice because very few people can read them seamlessly. I am okay using them if there isn't another option, but in this case, if we want to check if a String is a number, we can simply use ```org.apache.commons.lang3.math.NumberUtils.isDigits(String)``` or ```org.apache.commons.lang3.StringUtils.isNumeric(String)```

src/main/java/com/limechain/config/HostConfig.java
```diff
@@ -91,4 +98,18 @@ public String getGenesisPath() {
...
+
+    /**
+     * Gets the path to peer node based on the chain the node is configured
+     * Intended to be used when syncing state after warp sync
+     *
+     * @return genesis(chain spec) file path
+     */
+    public String getNodeSynckPath() {
+        return switch (chain) {
+            case POLKADOT, LOCAL -> polkadotNodePath;
```
Grigorov-Georgi: IMO We should probably combine westend with local networt together, instead of polkadot + local

Grigorov-Georgi (approved):

Zurcusa (approved):

PR #911, opened 2025-05-13
fix: grandpa validator's workflow

### LINKED ISSUE #422: Consensus integration tests

Grigorov-Georgi:
# Description

This PR refactors the Grandpa validator logic to enable starting a local network from the genesis, ensuring proper processing of grandpa rounds. The changes are backward-compatible and apply seamlessly when acting as a validator on other networks.

Partially fixes: https://github.com/LimeChain/Fruzhin/issues/422

## Checklist:

### DIFF
--- a/src/main/java/com/limechain/consensus/grandpa/GrandpaService.java
+++ b/src/main/java/com/limechain/consensus/grandpa/GrandpaService.java
@@ -79,8 +79,7 @@ public class GrandpaService {
 
         GrandpaSetState grandpaSetState = stateManager.getGrandpaSetState();
 
-        // We can add directly the authorities from the genesis in order to have them as the first authority set
-        // TODO: This may make the authority set changes to fail -> check it
+        // On the genesis we have empty set changes collection, so we should take the authorities from the chain spec
         if (grandpaSetState.getSetChanges().isEmpty()) {
             return Optional.of(grandpaSetState.getAuthoritySet());
         }
--- a/src/main/java/com/limechain/consensus/grandpa/GrandpaService.java
+++ b/src/main/java/com/limechain/consensus/grandpa/GrandpaService.java
@@ -79,9 +79,12 @@ public class GrandpaService {
 
         GrandpaSetState grandpaSetState = stateManager.getGrandpaSetState();
 
-        // On the genesis we have empty set changes collection, so we should take the authorities from the chain spec
-        if (grandpaSetState.getSetChanges().isEmpty()) {
-            return Optional.of(grandpaSetState.getAuthoritySet());
+        // In the genesis set we should take the authorities from the chain spec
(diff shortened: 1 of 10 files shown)


hMitov (approved):

src/main/java/com/limechain/consensus/grandpa/GrandpaService.java
```diff
@@ -64,16 +67,23 @@ public GrandpaRound createInitialRound(RoundState roundState) {
...
+        );
     }
 
     public Optional<GrandpaAuthoritySet> getAuthoritiesForBlock(BigInteger blockNumber) {
         GrandpaAuthoritySet authorities = null;
 
         GrandpaSetState grandpaSetState = stateManager.getGrandpaSetState();
 
+        // On the genesis we have empty set changes collection, so we should take the authorities from the chain spec
+        if (grandpaSetState.getSetChanges().isEmpty()) {
```
Zurcusa: We should check if we're in set 0 (genesis set). If we don't this could hide potential issues if we don't populate the `setChanges` correctly.

src/main/java/com/limechain/consensus/grandpa/round/GrandpaRound.java
```diff
@@ -232,6 +230,9 @@ public synchronized void update(boolean isPrevRoundChanged,
 
         if (previous != null) {
             shouldStartNextRound = previous.finalizedBlock != null;
+        } else {
+            // at the genesis -> round number is equal to one
+            shouldStartNextRound = roundNumber.equals(BigInteger.ONE);
```
Zurcusa: We have rounds with number 1 in every set. This logic would start R + 1 at every first round of a set. We should check if we're in genesis set 0 here as well. Maybe have a utility method that checks if the round's setId is 0.

src/main/java/com/limechain/consensus/grandpa/round/GrandpaRound.java
```diff
@@ -442,8 +476,12 @@ void attemptToFinalize() {
                 blockHeaders.removeFirst();
             }
 
-            blockState.finalizeBlock(finalizedBlock, createJustification(), authoritySet.getSetId());
-            syncState.finalizeBlock(finalizedBlock);
+            try {
+                blockState.finalizeBlock(finalizedBlock, createJustification(), authoritySet.getSetId());
+                syncState.finalizeBlock(finalizedBlock);
+            } catch (BlockStorageGenericException e) {
+                log.warning("Block cannot be finalized: " + e.getMessage());
```
Zurcusa:
```suggestion
                log.warning(String.format("Block cannot be finalized: %s", e.getMessage());
```

Zurcusa (approved):

PR #912, opened 2025-05-15
fix: Handled case when recovering public key from signature removed leading 0.

### LINKED ISSUE #899: Error: invalid compressed length for ECDSA keys
### Is there an existing issue?

- [x] I have searched the existing issues

### Description of bug

We get some errors on some ECDSA keys for invalid compressed length.

### Steps to reproduce

_No response_

hMitov:
# Description

This PR improves the handling of recovered public keys by correctly interpreting the behavior of BigInteger serialization with toByteArray(), based on the state of the second byte after construction.

First byte is not part of the public key, so it is always trimmed.

1. If the second byte after BigInteger creation is positive:
  - No extra leading 0 byte is needed during serialization.
  - Resulting byte array length = 64.
2. If the second byte after BigInteger creation is negative:
  - A leading 0 byte is appended during serialization to enforce positive two's complement encoding.
  - Resulting byte array length = 65.
3. If the first non-zero byte after leading 0 byte is positive:
  - All leading 0x00 bytes are stripped after toByteArray().
  - The resulting byte array length will be shorter than 64 bytes, depending on how many leading zeros were removed.
  
  
###   P.S.

An additional issue was identified in the EcdsaUtils.generateKeyPair(...) method.
The root cause is the same as described above. 
- The returned private key must be exactly 32 bytes
- The returned public key must be exactly 33 bytes (compressed format).

However, due to missing normalization, there were cases where the keys were either too short (with trimmed 0x00 bytes) or too long (with added 0x00 byte). 
  
Fixes #899 >

## Checklist:

### DIFF
--- a/src/main/java/com/limechain/runtime/hostapi/CryptoHostFunctions.java
+++ b/src/main/java/com/limechain/runtime/hostapi/CryptoHostFunctions.java
@@ -41,6 +41,8 @@ import java.util.Set;
 import java.util.logging.Level;
 
 import static com.limechain.runtime.hostapi.PartialHostApi.newImportObjectPair;
+import static com.limechain.utils.EcdsaUtils.PRIVATE_KEY_LEN;
+import static com.limechain.utils.EcdsaUtils.PUBLIC_KEY_COMPRESSED_LEN;
 
 /**
  * Implementations of the Crypto HostAPI functions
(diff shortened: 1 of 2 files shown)


src/main/java/com/limechain/runtime/hostapi/CryptoHostFunctions.java
```diff
@@ -515,7 +517,13 @@ public int ecdsaGenerateV1(int keyTypeId, RuntimePointerSize seed) {
             keyPair = EcdsaUtils.generateKeyPair();
         }
 
-        keyStore.put(pair.getKeyType(), keyPair.getSecond().raw(), keyPair.getFirst().raw());
+        byte[] publicKey = keyPair.getSecond().raw();
```
Zurcusa, line 520: It would be better if we normalize directly inside of the `generateKeyPair(String s)`. The problem here only happens when we invoke `BigInteger.toByteArray()` or in the case of the generate method
`PrivKey privKey = Secp256k1Kt.unmarshalSecp256k1PrivateKey(keyPair.getPrivateKey().toByteArray());`
If we don't normalize before this code we might end up with unexpected `PrivKey`.

Zurcusa (approved):

src/main/java/com/limechain/runtime/hostapi/CryptoHostFunctions.java
hMitov: The normalization logic was placed outside of generateKeyPair(...) because the raw() method in Secp256k1PrivateKey  class ends with a call to toByteArray(), which may lead to private keys with inconsistent or invalid lengths.
To ensure correctness and consistency, the private key should also be normalized within the generateKeyPair(...) method itself.

Grigorov-Georgi (approved):

PR #919, opened 2025-05-19
feat: Zombienet tests for consensuses 

### LINKED ISSUE #422: Consensus integration tests

Grigorov-Georgi:
# Description

- Add support for injecting session keys (Babe, Grandpa, Beefy) via environment variables
- Enable adding ECDSA keys via the author_insertKey RPC
- Add two new chainspecs for local dev and Zombienet testing (one with Alice only, one with Alice & Bob as authorities)
- Add Zombienet tests for Babe, Grandpa, and Beefy consensus mechanisms
- Enable setting a custom Prometheus port via cli argument; also add two new Prometheus metrics: `substrate_best_block_number` and `substrate_last_finalized_block_number`

Fixes https://github.com/LimeChain/Fruzhin/issues/422

## Checklist:


hMitov (approved):

src/main/java/com/limechain/client/LightClient.java
Zurcusa, line 1: Did you test if this works? I remember we had issues with light client when BlockState is initialized.

zombienet/0002-peer-discovery.zndsl
```diff
@@ -5,4 +5,4 @@ Creds: config
 # metrics
 alice: log line matches "Successfully connected to ([1-9]\d*) boot nodes" within 30 seconds
 alice: log line matches "Node successfully connected to a peer! Sync can start" within 30 seconds
-alice: log line matches "Connected peers: ([1-9]\d*)" within 60 seconds
+alice: log line matches "Bootstrap connected to ([1-9]\d*) nodes close to us" within 60 seconds
```
Zurcusa, line 8: I'd keep the old log. This one is printed by the libp2p before we open up a block announce connection. That means that we might have a p2p connection open, but it's still not considered a peer in Fruzhin terms.

Zurcusa (approved):

src/main/java/com/limechain/client/LightClient.java
Grigorov-Georgi: I fixed the issue by adding BlockState to the LightClient. And yes, the test works now.

